PR #6, opened 2019-04-10
Update sidecar injector to create/monitor/update webhook configuratio…

knrc:
…n: MAISTRA-205


install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml.tpl
```diff
@@ -1,3 +1,4 @@
+{{ define "mutatingwebhookconfiguration.yaml.tpl" }}
```
rcernich, line 1: If we've got a custom version of this file in Maistra/istio, can we integrate all of the updates here (i.e. the changes being made by the sed statements in update-yaml and patch-charts.sh)?

install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/clusterrole.yaml
```diff
@@ -14,4 +14,16 @@ rules:
...
 - apiGroups: ["admissionregistration.k8s.io"]
   resources: ["mutatingwebhookconfigurations"]
-  verbs: ["get", "list", "watch", "patch"]
+  verbs: ["*"]
+- apiGroups: ["extensions","apps"]
+  resources: ["deployments"]
+  resourceNames: ["istio-sidecar-injector"]
+  verbs: ["get"]
+- apiGroups: ["extensions"]
+  resources: ["deployments/finalizers"]
```
rcernich, line 23: Where is the finalizer set?  Is this something that was added upstream?

install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml.tpl
knrc: I'd prefer not, since most of the changes you refer to are OpenShift specific ones or related to upstream changes we're not involved with.  I intend this to go upstream soon so we are pulling in a change we expect from a future upstream release, we've done this a couple of times in the past but not very often.

install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/clusterrole.yaml
knrc: I'm not sure what you mean by 'added upstream', this hasn't been pushed up there yet.

This is a kubernetes requirement for any account which sets blockOwnerDeletion.

install/kubernetes/helm/istio/charts/sidecarInjectorWebhook/templates/mutatingwebhookconfiguration.yaml.tpl
rcernich: The one thing that I would have recommended was removing the namespace field from this, since these resources are non-namespaced.

knrc: The namespace is coming from upstream, while superfluous it doesn't have any bearing on the configuration and doesn't prevent the code from working so it's not necessary for us to remove it.

PR #7, opened 2019-05-01
Initial support for multiple namespaces in Pilot

luksa:
To make it watch multiple namespaces, run it with `--appNamespace ns1,ns2,ns3`

Based on my testing, it requires the following privileges:
- pods,services,endpoints in the watched namespaces
- secrets in the control plane's namespace
- nodes (it doesn't start without this)

To make it run without requiring privileges for reading nodes, run it with `--podLocalitySource pod`. An additional controller with cluster-level privileges (not implemented yet) must be used to copy the node's labels to the pod.


luksa:
This & more is now in https://github.com/Maistra/istio/tree/multitenancy

PR #9, opened 2019-05-13
Fix webhookName/webhookConfigName problems in sidecar injector

luksa:
Before this commit, the webhookConfigName flag was totally ignored,
while the webhookName was used as the name of the
mutatingwebhookconfiguration resource, which didn't match the flag's
description.

The webhookConfigName is the name of the mutatingwebhookconfiguration
resource, while the webhookName is the name of the webhook entry in
that resource. Currently, the webhookName is not used, since the resource
is created from a YAML file instead of simply adding an additional
webhook entry to the existing resource.


knrc:
Do not merge this, the original change in the multitenancy branch deviates from the upstream codebase so we need to review

luksa:
Replaced by #25

PR #10, opened 2019-05-17
Multi tenancy member roll

knrc:


security/cmd/istio_ca/main.go
```diff
@@ -228,6 +235,12 @@ func init() {
 	flags.BoolVar(&opts.dualUse, "experimental-dual-use",
 		false, "Enable dual-use mode. Generates certificates with a CommonName identical to the SAN.")
 
+	// OpenShift ServiceMesh options
+	flags.StringVar(&opts.memberRollName, "member-role-name", "",
+		"The name of the ServiceMeshMemberRoll resource.  If specified the server will monitor this resource to discover the application namespaces.")
+	flags.DurationVar(&opts.memberRollResyncPeriod, "member-role-resync-period",
```
rcernich: Maybe rename this to `memberRollName` to match the other component exe's?

knrc: I've fixed the typo however I've kept the format with dashes since that's used with the rest of the arguments.  Unfortunately this is inconsistent with the other services.

pkg/servicemesh/apis/servicemesh/v1alpha3/doc.go
```diff
@@ -0,0 +1,5 @@
+// +k8s:deepcopy-gen=package
```
brian-avery: What purpose does this file serve?

brian-avery (approved):
LGTM

pkg/servicemesh/apis/servicemesh/v1alpha3/doc.go
knrc: It's part of the generation directives for the package

PR #12, opened 2019-05-22
Include namespace indexing for services: MAISTRA-421

knrc:


knrc:
@brian-avery I'm merging this into the pilot codebase, will try to rebuild before tomorrow

PR #13, opened 2019-05-22
Return meaningful resource versions from multiListWatcher: MAISTRA-422

knrc:


knrc:
@brian-avery I'm also going to merge this, we will need to rebuild citadel, galley and pilot after this one

PR #25, opened 2019-06-19
[MAISTRA-503] Fix the references to webHookName -> webHookConfigName

jwendell:
WebhooksConfigs were being created with the wrong name (the one in the
webHookName field), which is static. We should be referencing the
WebHookConfigName, which has a different name for each control plane.

This error was preventing the sidecar being attached to pods.

### DIFF
--- a/pilot/cmd/sidecar-injector/main.go
+++ b/pilot/cmd/sidecar-injector/main.go
@@ -83,7 +83,7 @@ var (
 				Port:                flags.port,
 				WebhookConfigFile:   flags.webhookConfigFile,
 				Namespace:           flags.namespace,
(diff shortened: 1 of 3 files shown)


luksa:
Hmm, I fixed something like this a while ago. Let me find the PR.

luksa:
Ah, my PR was never was never merged: https://github.com/Maistra/istio/pull/9

jwendell:
@luksa :( It seems it got lost. No Jira for it, duplicated work :( :(

jwendell:
@knrc any thoughts here?

knrc (approved):
lgtm

PR #26, opened 2019-06-20
Enable injecting additional annotations into pods (MAISTRA-518)

luksa:
When using Istio CNI through Multus CNI, the pod needs to contain an
annotation that tells Multus to use the Istio CNI plugin on the pod.
The annotation needs to be injected at pod creation time so that it 
already exists when the kubelet starts setting up the pod's network. 
Adding the annotation afterwards has no effect, since the pod's network 
has already been set up.  

For this reason, the sidecar-injector now allows you to specify any 
number of additional annotations to inject (in addition to the
(init)containers, volumes, etc.)

### DIFF
--- a/pilot/pkg/kube/inject/webhook_test.go
+++ b/pilot/pkg/kube/inject/webhook_test.go
@@ -605,12 +605,14 @@ func TestWebhookInject(t *testing.T) {
 			wantFile:     "TestWebhookInject_http_probe_nosidecar_rewrite.patch",
 			templateFile: "TestWebhookInject_http_probe_nosidecar_rewrite_template.yaml",
 		},
+		{
+			inputFile:    "TestWebhookInject_annotations.yaml",
+			wantFile:     "TestWebhookInject_annotations.patch",
+			templateFile: "TestWebhookInject_annotations_template.yaml",
+		},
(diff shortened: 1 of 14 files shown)


pilot/pkg/kube/inject/webhook_test.go
```diff
@@ -605,12 +605,14 @@ func TestWebhookInject(t *testing.T) {
...
+			inputFile:    "TestWebhookInject_annotations.yaml",
+			wantFile:     "TestWebhookInject_annotations.patch",
+			templateFile: "TestWebhookInject_annotations_template.yaml",
+		},
 	}
 
 	for i, c := range cases {
-		if c.inputFile != "TestWebhookInject_http_probe_nosidecar_rewrite.yaml" {
-			continue
-		}
```
jwendell, line 613: Is this an upstream typo/bug/leftover that should be reported/fixed?

jwendell (approved):

pilot/pkg/kube/inject/webhook_test.go
luksa: Yes. It was a bug, but it was fixed already.

PR #27, opened 2019-06-27
MAISTRA-551 Run sidecar with a dynamically-determined runAsUser ID

luksa:


pilot/pkg/kube/inject/webhook.go
```diff
@@ -687,6 +712,19 @@ func (wh *Webhook) inject(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionRespons
 	return &reviewResponse
 }
 
+func getProxyUID(pod corev1.Pod) (*uint64, error) {
```
jwendell, line 715: What about return -1 in cases of errors? That would eliminate the need of dealing with pointers above

pilot/pkg/kube/inject/inject.go
```diff
@@ -715,7 +717,8 @@ func intoObject(sidecarTemplate string, meshconfig *meshconfig.MeshConfig, in ru
 		podSpec,
 		metadata,
 		meshconfig.DefaultConfig,
-		meshconfig)
+		meshconfig,
+		DefaultSidecarProxyUID)
 	if err != nil {
```
jwendell, line 722: Is it safe to use the default value here?

jwendell (approved):
LGTM, just a few comments/questions

pilot/pkg/kube/inject/webhook.go
luksa: Yeah, that was my first instinct, too... but IMHO a pointer to represent optionality is much better than using magic values. Not sure what the golang position on this is.

pilot/pkg/kube/inject/inject.go
luksa: This preserves the old behavior. The thing is, we can't put the UID into the configmap, because the uid that should be used differs from one namespace to the next.

PR #28, opened 2019-06-27
fix(sample): Correct file permissions to run under any uuid

aslakknutsen:


samples/bookinfo/src/productpage/uid_entrypoint
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+if ! whoami &> /dev/null; then
```
luksa, line 2: Looks like the `&> /dev/null` doesn't do what you'd expect, because the script is running under `sh` not `bash`.

samples/bookinfo/src/productpage/Dockerfile
```diff
@@ -14,18 +14,30 @@
...
+RUN chgrp -R 0 ${APP_ROOT} && \
+    chmod -R og=rwx ${APP_ROOT} && \
+    chmod -R g=rw /etc/passwd
 
 EXPOSE 9080
-WORKDIR /opt/microservices
+USER 10001
+WORKDIR ${APP_ROOT}
+
 RUN python -m unittest discover
```
luksa, line 41: This line creates the `microservice.log` file owned by `10001`, which isn't writable by the actual user ID when you run the container.

luksa (changes requested):
Found two issues.

samples/bookinfo/src/productpage/Dockerfile
aslakknutsen: Hmm, it all ran fine on my cluster when I was testing it. A barebone 3.11 install tho.

samples/bookinfo/src/productpage/uid_entrypoint
aslakknutsen: This is from the Openshift docs :)

samples/bookinfo/src/productpage/Dockerfile
aslakknutsen: Maybe it's missing the `chmod -R g=u /etc/passwd` bit...

luksa:
```
$ docker run -it --user 1000130000 luksa/examples-bookinfo-productpage-v1:maistra-0.12 -- ls -la
total 64
drwxrwxrwx. 1 root  root  4096 Jun 28 13:58 .
drwxr-xr-x. 1 root  root  4096 Jun 27 15:45 ..
-rw-r--r--. 1 10001 root     0 Jun 28 13:58 microservice.log
-rwxrwxrwx. 1 root  root 11649 Jun 27 15:43 productpage.py
-rw-r--r--. 1 10001 root 10655 Jun 28 13:58 productpage.pyc
-rw-rwxrwx. 1 root  root   380 Jun 28 12:16 requirements.txt
drwxrwxrwx. 1 root  root  4096 Jun 28 12:44 templates
-rw-rwxrwx. 1 root  root  3062 Jan  9 12:12 test_productpage.py
-rw-r--r--. 1 10001 root  2114 Jun 28 13:58 test_productpage.pyc
-rwxrwxrwx. 1 root  root   200 Jun 28 12:44 uid_entrypoint
```

luksa: BTW, I simply removed this line to make it work. But I guess the whole point of this line is to run the unit test during the build and to pre-compile the .pyc files. Should probably chown the created files after running the unit tests.

aslakknutsen: the directory has this tho: `chmod -R og=rwx ${APP_ROOT}` which should allow anyone to write to that directory ?

aslakknutsen: Ah, in Docker.. hmm, not sure I actually ever tried to run that, e.g. the unit test part of it.

luksa: yeah, the directory is writable, but since the microservice.log file is created during build time (and as you can see, its permissions are rw-r-r), it isn't writable to anyone else than user id 10001.

aslakknutsen: Reproducible locally. checking...

aslakknutsen: When running it in Openshift I get this tho:

```
$ ls -la
total 72
drwxrwxrwx. 1 root    root  4096 Jun 28 14:50 .
drwxr-xr-x. 1 root    root  4096 Jun 28 14:34 ..
-rw-r--r--. 1 default root   326 Jun 28 14:51 microservice.log
-rwxrwxrwx. 1 root    root 11609 Jun 27 15:48 productpage.py
-rw-r--r--. 1 default root 10573 Jun 28 14:34 productpage.pyc
-rw-rwxrwx. 1 root    root   384 Jun 27 15:48 requirements.txt
-rw-r--r--. 1 default root     0 Jun 28 14:50 stderr.log
-rw-r--r--. 1 default root     0 Jun 28 14:50 stdout.log
drwxrwxrwx. 1 root    root  4096 Jun 28 14:34 templates
-rw-rwxrwx. 1 root    root  3062 Jun 27 15:45 test_productpage.py
-rw-r--r--. 1 default root  2114 Jun 28 14:34 test_productpage.pyc
-rwxrwxrwx. 1 root    root   198 Jun 27 15:49 uid_entrypoint
$ cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
bin:x:2:2:bin:/bin:/usr/sbin/nologin
sys:x:3:3:sys:/dev:/usr/sbin/nologin
sync:x:4:65534:sync:/bin:/bin/sync
games:x:5:60:games:/usr/games:/usr/sbin/nologin
man:x:6:12:man:/var/cache/man:/usr/sbin/nologin
lp:x:7:7:lp:/var/spool/lpd:/usr/sbin/nologin
mail:x:8:8:mail:/var/mail:/usr/sbin/nologin
news:x:9:9:news:/var/spool/news:/usr/sbin/nologin
uucp:x:10:10:uucp:/var/spool/uucp:/usr/sbin/nologin
proxy:x:13:13:proxy:/bin:/usr/sbin/nologin
www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
backup:x:34:34:backup:/var/backups:/usr/sbin/nologin
list:x:38:38:Mailing List Manager:/var/list:/usr/sbin/nologin
irc:x:39:39:ircd:/var/run/ircd:/usr/sbin/nologin
gnats:x:41:41:Gnats Bug-Reporting System (admin):/var/lib/gnats:/usr/sbin/nologin
nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
_apt:x:100:65534::/nonexistent:/bin/false
default:x:10001:0:default user:/:/sbin/nologin
```

And it seems to start up just fine. Writing to the microservices.log file just fine. 

I do get the print line about missing user tho.

luksa: You're running it with `runAsUser: 10001`, right? Are you sure you're not using the `anyuid` SCC?

aslakknutsen: docker run -it --user 10001 ... -- ls -la


work in docker

luksa: Yes, that works since the build-time and run-time user ids are the same. When running in openshift under the `restricted` SCC, the run-time user id will be something like 1000130000.

PR #33, opened 2019-07-18
MAISTRA-451 Do not log error when we cannot delete secret

dgn:
When istio-operator reconciles the SMMR and a namespace has been
removed, it will revoke Citadel's access to that namespace by
removing the RoleBinding. Once Citadel picks up the change in the
SMMR, it tries to delete the workload secrets in that namespace,
but fails to due missing permissions.

I guess this is only a temporary solution until we find a way to clean up the secrets properly.

### DIFF
--- a/security/pkg/k8s/controller/workloadsecret.go
+++ b/security/pkg/k8s/controller/workloadsecret.go
@@ -23,7 +23,7 @@ import (
 
 	"istio.io/istio/pkg/servicemesh/controller"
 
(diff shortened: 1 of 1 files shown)


rcernich (approved):
lgtm

knrc (approved):
lgtm

knrc (approved):

knrc (approved):

PR #36, opened 2019-07-29
MAISTRA-587 Add cmdline flag to webhooks to disable management of webhook configurations

dgn:


dgn:
Okay, I just realised that upstream doesn't do this, it was added by @knrc in [this pull request](https://github.com/Maistra/istio/pull/6) from April. Should I revert the changes made back then in order to stay closer to upstream? It doesn't make sense IMO to first add this functionality and then disable it with a flag.

knrc:
@dgn There is an issue upstream to add this capability, it was assigned to me within the last week or so to implement.

galley/pkg/crd/validation/webhook.go
```diff
@@ -120,6 +120,12 @@ type WebhookParameters struct {
 
 	// Enable galley validation mode
 	EnableValidation bool
+
+	// ManageWebhookConfig determines whether the ValidatingWebhookConfiguration
+	// should be watched and updated by the webhook itself. This can be disabled,
+	// in which case we simply assume the configuration to be correct. This is
+	// helpful if you want to remove ClusterRole permissions from galley.
+	ManageWebhookConfig bool
```
jwendell, line 130: I'd add: `... remove ClusterRole permissions from galley and it is safe to do that e.g. if someone else is taking care of that`.
Wording may be bad, hopefully you get the idea

galley/pkg/crd/validation/webhook.go
```diff
@@ -319,14 +334,17 @@ func (wh *Webhook) Run(stopCh <-chan struct{}) {
 		return
 	}
 
-	// Try to create the initial webhook configuration (if it doesn't
-	// already exist). Setup a persistent monitor to reconcile the
-	// configuration if the observed configuration doesn't match
-	// the desired configuration.
-	if err := wh.rebuildWebhookConfig(); err == nil {
-		wh.createOrUpdateWebhookConfig()
+	webhookChangedCh := make(chan struct{}, 1000)
```
jwendell: `1000` is not necessary here. This variable won't be used anyway if the flag is false; and if it's true, it will be overridden in the block below.

I wish there could be a way to remove it completely from the `select` if the flag is false...

pilot/pkg/kube/inject/webhook.go
```diff
@@ -156,6 +157,12 @@ type WebhookParameters struct {
 	// that is periodically updated.
 	HealthCheckFile string
 
+	// ManageWebhookConfig determines whether the MutatingWebhookConfiguration
+	// should be watched and updated by the webhook itself. This can be disabled,
+	// in which case we simply assume the configuration to be correct. This is
+	// helpful if you want to remove ClusterRole permissions from the injector.
+	ManageWebhookConfig bool
```
jwendell, line 166: Same comment as above

pilot/pkg/kube/inject/webhook.go
```diff
@@ -258,14 +270,17 @@ func (wh *Webhook) Run(stop <-chan struct{}) {
 	}()
 	defer wh.stop()
 
-	// Try to create the initial webhook configuration (if it doesn't
-	// already exist). Setup a persistent monitor to reconcile the
-	// configuration if the observed configuration doesn't match
-	// the desired configuration.
-	if err := wh.rebuildWebhookConfig(); err == nil {
-		wh.createOrUpdateWebhookConfig()
+	webhookChangedCh := make(chan struct{}, 1000)
```
jwendell: Same comment as above

jwendell (approved):

brian-avery (approved):

galley/pkg/crd/validation/webhook.go
```diff
@@ -120,6 +120,14 @@ type WebhookParameters struct {
 
 	// Enable galley validation mode
 	EnableValidation bool
+
+	// ManageWebhookConfig determines whether the ValidatingWebhookConfiguration
+	// should be watched and updated by the webhook itself. This can be disabled,
+	// in which case we simply assume the configuration to be correct. This is
+	// helpful if you want to remove ClusterRole permissions from galley.
+	// NOTE: settings this to false requires the WebhookConfiguration to be created
```
dgn: oh, typo.

pilot/pkg/kube/inject/webhook.go
```diff
@@ -156,6 +157,14 @@ type WebhookParameters struct {
 	// that is periodically updated.
 	HealthCheckFile string
 
+	// ManageWebhookConfig determines whether the MutatingWebhookConfiguration
+	// should be watched and updated by the webhook itself. This can be disabled,
+	// in which case we simply assume the configuration to be correct. This is
+	// helpful if you want to remove ClusterRole permissions from the injector.
+	// NOTE: settings this to false requires the WebhookConfiguration to be created
```
dgn: again, typo.

galley/cmd/galley/cmd/server.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
knrc: Minor nit but remove this change since it doesn't serve any purpose.

knrc (changes requested):

knrc (approved):

PR #37, opened 2019-07-29
MAISTRA-668 Fix expected output of webhook injection

luksa:


dgn (approved):
unit tests pass as expected.

there are still some failing, but those are related to https://github.com/Maistra/istio/pull/6

jwendell (approved):

PR #39, opened 2019-08-05
Add support for ServiceMeshPolicy and ServiceMeshRbacConfig: MAISTRA-630

knrc:


galley/pkg/metadata/types.go
```diff
@@ -40,8 +40,8 @@ var Types *resource.Schema
 
 var (
 
-	// istio/authentication/v1alpha1/meshpolicies metadata
-	IstioAuthenticationV1alpha1Meshpolicies resource.Info
+	// maistra/authentication/v1/servicemeshpolicies metadata
+	MaistraAuthenticationV1Servicemeshpolicies resource.Info
```
dgn: did you generate these changes? when I run `$GOPATH/src/istio.io/istio/galley/tools/gen-meta/gen-meta.sh runtime pkg/metadata/types.go` I get a file that is semantically identical but entries are in a different order.

pilot/pkg/config/coredatamodel/controller_test.go
```diff
@@ -375,7 +375,7 @@ func TestApplyClusterScopedAuthPolicy(t *testing.T) {
 	controller := coredatamodel.NewController(testControllerOptions)
 
 	message0 := convertToResource(g, model.AuthenticationPolicy.MessageName, []proto.Message{authnPolicy0})
-	message1 := convertToResource(g, model.AuthenticationMeshPolicy.MessageName, []proto.Message{authnPolicy1})
+	message1 := convertToResource(g, model.AuthenticationServiceMeshPolicy.MessageName, []proto.Message{authnPolicy1})
```
dgn, line 378: nit, but should we rename tests to `TestApplyMeshScopedAuthPolicy`?

install/kubernetes/helm/istio/charts/security/templates/create-custom-resources-job.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
     heritage: {{ .Release.Service }}	
     release: {{ .Release.Name }}	
 rules:	
-- apiGroups: ["authentication.istio.io"] # needed to create default authn policy	
+- apiGroups: ["authentication.maistra.io"] # needed to create default authn policy
```
dgn, line 23: I guess we could also move this into a normal `Role`, limiting it to being able to create these resources in the `{{ .Release.Namespace }}` namespace

pilot/pkg/config/kube/crd/conversion.go
```diff
@@ -111,6 +111,9 @@ func ResourceName(s string) string {
 
 // ResourceGroup generates the k8s API group for each schema.
 func ResourceGroup(schema *model.ProtoSchema) string {
+	if schema.GroupDomain != "" {
+		return schema.Group + schema.GroupDomain
```
dgn: This is missing a `.` in between values -- `model.IstioAPIGroupDomain` is set to `.istio.io`, whereas we set our domain to `maistra.io`. I think we should inject the `.` here instead of having to put it in the `ProtoSchema`

install/kubernetes/helm/istio-init/files/crd-10.yaml
```diff
@@ -184,16 +184,15 @@ metadata:
...
-    kind: ClusterRbacConfig
-    plural: clusterrbacconfigs
-    singular: clusterrbacconfig
+    kind: ServiceMeshRbacConfig
+    plural: servicemeshrbacconfigs
+    singular: servicemeshrbacconfig
     categories:
     - istio-io
     - rbac-istio-io
-  scope: Cluster
```
dgn, line 195: this will have to be `scope: Namespaced`

galley/pkg/metadata/types.go
knrc: I thought I did but I can redo it

pilot/pkg/config/coredatamodel/controller_test.go
knrc: Sounds good

install/kubernetes/helm/istio/charts/security/templates/create-custom-resources-job.yaml
knrc: I believe this is handled by Rob's scripts

pilot/pkg/config/kube/crd/conversion.go
knrc: Good catch

install/kubernetes/helm/istio-init/files/crd-10.yaml
knrc: Will add that

tools/convert_RbacConfig_to_ClusterRbacConfig.sh
```diff
@@ -2,10 +2,10 @@
...
 
-# This script is provided to help converting RbacConfig to ClusterRbacConfig automatically. The RbacConfig
-# will be deleted after the corresponding ClusterRbacConfig is successfully applied.
-# The RbacConfig is deprecated by ClusterRbacConfig due to an implementation bug that could cause the
-# RbacConfig to be namespace scoped in some cases. The ClusterRbacConfig has exactly same specification
+# This script is provided to help converting RbacConfig to ServiceMeshRbacConfig automatically. The RbacConfig
+# will be deleted after the corresponding ServiceMeshRbacConfig is successfully applied.
+# The RbacConfig is deprecated by ServiceMeshRbacConfig due to an implementation bug that could cause the
+# RbacConfig to be namespace scoped in some cases. The ServiceMeshRbacConfig has exactly same specification
 # as the RbacConfig, but with correctly implemented cluster scope.
```
dgn: I'm not entirely sure about this script. I don't think it makes sense to change it like this, because it doesn't produce the expected outcome: the ServiceMeshRbacConfig will have to live in the CP namespace; this script will just create it wherever the RbacConfig used to be. Also, this last sentence in the comment is now wrong (ServiceMeshRbacConfig is not cluster-scoped).

knrc: I can either back out these changes or fix the comment to say mesh scoped.  The transition from RbacConfig to ClusterRbacConfig was some time ago so I do not believe we will need to use this script.

install/kubernetes/helm/istio-init/files/crd-10.yaml
luksa: Defaults to `Namespaced`, so not really necessary to add it, apart from making it explicit.

lucasponce:
Note this PR is going to have a significant impact in Kiali.
MeshPolicies and ClusterRbacConfig are resources managed by Kiali.
MeshPolicies are also used to validate mTLS configuration.

With these changes, we need to add support for these two objects (if I understand correctly) and change the logic of calculation when we are in Maistra vs Istio upstream.

Also the UI should be changed to only show these objects.

pilot/pkg/model/config.go
```diff
@@ -36,6 +36,12 @@ const (
 	DefaultAuthenticationPolicyName = "default"
 )
 
+var (
+	// defaultAuthenticationServiceMeshPolicyNamespace is the namespace containing the mesh-scoped authentication policy.
+	// only policy in this namespace will be considered.
+	defaultAuthenticationServiceMeshPolicyNamespace = "istio-system"
```
luksa: Remove "default" from the variable name (I'd only expect constants to be named "defaultSomething")?

lucasponce:
@knrc can we discuss/study if there is a possibility to have this PR's logic, but maintaining Istio upstream naming for compatibility ?

knrc:
> @knrc can we discuss/study if there is a possibility to have this PR's logic, but maintaining Istio upstream naming for compatibility ?

To retain the upstream name it would need to remain cluster scoped.  We actually tried that last week but there were security issues since information leaked between the meshes.

Since we need the scope to change we need to change the name of the CRD.

knrc:
> Note this PR is going to have a significant impact in Kiali.
> MeshPolicies and ClusterRbacConfig are resources managed by Kiali.
> MeshPolicies are also used to validate mTLS configuration.

@lucasponce This is surprising since we have been discussing this change for some time during our weekly multi-tenant calls.  We were lead to believe the change would be relatively straight forward since the contents remain the same and it's only the name of the resource (and being namespaced) which is changing.

> With these changes, we need to add support for these two objects (if I understand correctly) and change the logic of calculation when we are in Maistra vs Istio upstream.

+1

> Also the UI should be changed to only show these objects.

How much effort do you think this would require?

pilot/pkg/model/config.go
knrc: Will do

lucasponce:
> @lucasponce This is surprising since we have been discussing this change for some time during our weekly multi-tenant calls. We were lead to believe the change would be relatively straight forward since the contents remain the same and it's only the name of the resource (and being namespaced) which is changing.

Yes, but we discussed several scenarios, IIRC, one maintaining the API, but adding them namespace scoped, and other renaming them. The last would take more effort as basically Kiali will need to support Maistra specific objects, meanwhile, maintaining the same naming, the code base will remain intact.

lucasponce:
From https://issues.jboss.org/browse/MAISTRA-630 description:

> There are two alternative solutions

> - Keep the upstream names but make their CRDs namespaced
> - Replace these resources with alternatives such as ServiceMeshPolicy and ServiceMeshRbacConfig
> The first would likely be simpler to implement but lead to clashes with upstream installations whereas the second would be a more invasive change for ourselves and likely also the Kiali team

Which is the case, in the first scenario, Kiali codebase would be compatible, but the second will require some days-effort to adapt it and prepare it.

knrc:
> From https://issues.jboss.org/browse/MAISTRA-630 description:
> 
> > There are two alternative solutions
> 
> > * Keep the upstream names but make their CRDs namespaced

We tried to do this last week.  We couldn't make them namespaced as this would clash with the upstream istio definition however we tried to rename them so the resource would share the name of the service mesh control plane namespace.  The filtering didn't work and was leaking information to other meshes.

> > * Replace these resources with alternatives such as ServiceMeshPolicy and ServiceMeshRbacConfig
> >   The first would likely be simpler to implement but lead to clashes with upstream installations whereas the second would be a more invasive change for ourselves and likely also the Kiali team
> 
> Which is the case, in the first scenario, Kiali codebase would be compatible, but the second will require some days-effort to adapt it and prepare it.

The first solution would likely also require changes, either to namespaced resources (breaking upstream) or resource names (still cluster wide but filtered).  We tried but rejected those solutions.

install/kubernetes/helm/istio/charts/security/templates/create-custom-resources-job.yaml
rcernich: Yes, patch-charts.sh removes all these jobs.

tools/convert_RbacConfig_to_ClusterRbacConfig.sh
rcernich: Do we want to do this here, or in the operator (like we do for most of the changes to the helm charts)?

knrc: I'll rollback the script changes in the next PR.  As far as the helm chart updates are concerned I think it's better to keep the changes with the code so we have them in a single commit.  We've done this for other changes in the past where code changes were needed but we try to minimise this.

install/kubernetes/helm/istio-init/files/crd-10.yaml
knrc: +1 but I believe the rest declare their scope so I've added it to make the scope explicit/obvious.

pilot/pkg/model/config.go
knrc: Have updated both to remove default

knrc:
I've squashed the commits into one

install/kubernetes/helm/istio-init/files/crd-10.yaml
dgn: It has a default value, but it's also a required field (not sure how those two fit together - I guess if you set it to an illegal value the default is used). I saw an error trying to create a CRD without this field

lucasponce:
Please @knrc , how I can deploy this PR in a local 3.11 version that I can use to propagate these changes into Kiali ?

Could you describe us some steps ?

Thanks.

pilot/pkg/model/validation.go
```diff
@@ -129,15 +129,15 @@ func (descriptor ConfigDescriptor) Validate() error {
 		}
 		descriptorTypes[v.Type] = true
 		if v.ClusterScoped {
-			if _, exists := clusterMessages[v.MessageName]; exists {
-				errs = multierror.Append(errs, fmt.Errorf("duplicate message type: %q", v.MessageName))
+			if _, exists := clusterMessages[v.Collection]; exists {
```
dgn, line 132: I don't 100% agree with this change; I think we should either
- change ClusterScoped to MeshScoped, but that would be a breaking change to ProtoSchema, so rather not
- add MeshScoped field to ProtoSchema with similar behaviour so that a messageType can be defined as one of
  - only namespace-scoped
  - namespace-scoped & cluster-scoped
  - namespace-scoped & mesh-scoped

However I'm happy to keep things as they are for now, as this is really only ever used in validation.

dgn (approved):

pilot/pkg/model/validation.go
knrc: I don't get your comment, this is essentially no longer used

knrc: The resources are either namespaces or clustered, there is no mesh scoped.  This behaviour has not changes with the PR

pilot/pkg/config/kube/crd/conversion.go
```diff
@@ -111,6 +111,9 @@ func ResourceName(s string) string {
 
 // ResourceGroup generates the k8s API group for each schema.
 func ResourceGroup(schema *model.ProtoSchema) string {
+	if schema.GroupDomain != "" {
+		return schema.Group + "." + schema.GroupDomain
```
luksa, line 115: Why not add the dot to the actual value of `schema.GroupDomain` to keep that consistent with `model.IstioAPIGroupDomain`, which contains a dot?

brian-avery (approved):

pilot/pkg/config/kube/crd/conversion.go
knrc: Those come from parameters elsewhere in the code, it would be easier if those areas only had to configure the domain aspect and not have to include the '.'

knrc: For example
`		Group:         "authentication",`
`		GroupDomain:   "maistra.io",`
`		Version:       "v1",`

luksa: OK. I found this odd because of the comment here: https://github.com/Maistra/istio/blob/5c495dfab939aa419dc0863c57bcb14f59dcbdfe/pilot/pkg/model/config.go#L218

Strictly reading the comment: if it truly defaults to IstioAPIGroupDomain, then its value contains a dot. But if users set it, they should not add the dot, which is a minor inconsistency.

luksa: I mean. A user may go look for the value of `IstioAPIGroupDomain` and think they need to add the dot. And the code here doesn't skip adding the dot if the dot is there.

knrc: I'll tidy that up

knrc: I've removed the `.` from the default and am explicitly adding it.

luksa (approved):

knrc:
The latest PR squashes the commits in preparation for merging.

PR #41, opened 2019-08-09
Maistra-711 Fix webhook unit tests + MAISTRA-745 Fix nil pointer dereference in pilot

dgn:
This fixes two issues and makes unit tests pass again

### DIFF
--- a/pilot/pkg/kube/inject/webhook_test.go
+++ b/pilot/pkg/kube/inject/webhook_test.go
@@ -1218,7 +1218,8 @@ func createWebhook(t testing.TB, sidecarTemplate string) (*Webhook, func()) {
 	}
 
 	wh, err := NewWebhook(WebhookParameters{
-		ConfigFile: configFile, MeshFile: meshFile, CertFile: certFile, KeyFile: keyFile, Port: port})
+		ConfigFile: configFile, MeshFile: meshFile, CertFile: certFile, KeyFile: keyFile, Port: port,
+		ManageWebhookConfig: false})
 	if err != nil {
 		cleanup()
 		t.Fatalf("NewWebhook() failed: %v", err)
@@ -1255,7 +1256,10 @@ func TestRunAndServe(t *testing.T) {
          "name":"istio-proxy",
          "resources":{
 
-         }
+		 },
+		 "securityContext":{
+			 "runAsUser":1337
+		 }
       }
    },
    {
@@ -1370,7 +1374,7 @@ func TestRunAndServe(t *testing.T) {
 			}
 
 			if !bytes.Equal(gotPatch.Bytes(), wantPatch.Bytes()) {
-				t.Fatalf("got bad patch: \n got %v \n want %v", gotPatch, wantPatch)
+				t.Fatalf("got bad patch: \n got %v \n want %v", gotPatch.String(), wantPatch.String())
 			}
 		})
 	}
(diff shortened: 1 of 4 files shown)


pilot/pkg/kube/inject/webhook_test.go
```diff
@@ -1370,7 +1374,7 @@ func TestRunAndServe(t *testing.T) {
 			}
 
 			if !bytes.Equal(gotPatch.Bytes(), wantPatch.Bytes()) {
-				t.Fatalf("got bad patch: \n got %v \n want %v", gotPatch, wantPatch)
+				t.Fatalf("got bad patch: \n got %v \n want %v", gotPatch.String(), wantPatch.String())
```
luksa, line 1377: Is this necessary? I thought it calls `.String()` automatically?

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -66,14 +66,18 @@ func (configgen *ConfigGeneratorImpl) buildGatewayListeners(env *model.Environme
 				break
 			}
 		}
+		listenerPort := int(portNumber)
+		if si != nil {
+			listenerPort = si.Endpoint.Port
+		}
```
luksa, line 72: Since we're not using the ServiceInstance anywhere, why not declare listenerPort just above the for loop that finds the ServiceInstance and setting `listenerPort` when it finds it (instead of setting the `si` variable).

luksa (commented):
Two minor comments. But shouldn't this be two separate PRs?

pilot/pkg/networking/core/v1alpha3/gateway.go
luksa: Ah, nevermind. Didn't see that we're using `si` further down.

luksa: Oh, but my comment is still valid. You could set `listenerPort` just after we set `si`.

pilot/pkg/kube/inject/webhook_test.go
dgn: Oh, no this is not strictly necessary. Used it for debugging; it was spitting out byte arrays like
```
[25 05 155 ...]
```

and I wanted to look at the string representation. I can remove it again.

pilot/pkg/networking/core/v1alpha3/gateway.go
dgn: True. Will set it in the loop.

pilot/pkg/kube/inject/webhook_test.go
luksa: No, if it helps, then leave it in. I just thought Fatalf already calls .String(), but apparently not?

dgn:
> Two minor comments. But shouldn't this be two separate PRs?

I don't like merging things to a broken state ;) You need both to have all unit tests pass. But you're right, I will split up.

pilot/pkg/kube/inject/webhook_test.go
luksa: Yeah, definitely leave `.String()` in.

luksa:
This wouldn't be squashed anyway, so having a single PR wouldn't help with things like `git bisect`.

PR #42, opened 2019-08-09
MAISTRA-745 Don't panic if ServiceInstance not found

dgn:


luksa (approved):

brian-avery (approved):

jwendell (approved):

rcernich:
Should we even try to create a listener if there's no ServiceInstance?  Can we have a gateway without a ServiceInstance?  I think not configuring a listener and emitting an error would be more appropriate.

dgn:
> Should we even try to create a listener if there's no ServiceInstance? Can we have a gateway without a ServiceInstance? I think not configuring a listener and emitting an error would be more appropriate.

Yes, I thought about that, too. Upstream just passes on `nil`, too, which is then picked up by the plugins later; I thought we should not change the behavior if not necessary.

rcernich:
I guess that's the safer option.

rcernich (approved):

knrc (approved):

PR #47, opened 2019-08-22
MAISTRA-809: Force a push if we discovery a pod with an existing conn…

knrc:
…ection

### DIFF
--- a/pilot/pkg/proxy/envoy/v2/eds.go
+++ b/pilot/pkg/proxy/envoy/v2/eds.go
@@ -501,7 +501,7 @@ func (s *DiscoveryServer) edsIncremental(version string, push *model.PushContext
 	}
 	adsLog.Infof("Cluster init time %v %s", time.Since(t0), version)
 
-	s.startPush(version, push, false, edsUpdates)
+	s.startPush(version, push, false, edsUpdates, "")
 }
 
 // WorkloadUpdate is called when workload labels/annotations are updated.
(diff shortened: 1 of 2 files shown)


pilot/pkg/proxy/envoy/v2/eds.go
```diff
@@ -521,6 +521,23 @@ func (s *DiscoveryServer) WorkloadUpdate(id string, labels map[string]string, _
...
+			// then do a full xDS push for this proxy;
+			// otherwise:
+			//   case 1: the workload has no sidecar proxy, no need xDS push at all.
+			//   case 2: the workload xDS connection has not been established,
+			//           also no need to trigger a full push here.
+			if connection.modelNode.IPAddresses[0] == id {
+				// There is a possibility that the pod comes up later than endpoint.
+				// So no endpoints add/update events after this, we should request
+				// full push immediately to speed up sidecar startup.
+				go s.startPush(versionInfo(), s.globalPushContext(), false, nil, id)
```
dgn, line 538: The upstream PR seems to be triggering a full push for the connection in this case, shouldn't we do that, too?

dgn:
This looks promising, I'm currently trying to verify that this resolves the issue

pilot/pkg/proxy/envoy/v2/eds.go
dgn: Going through the logs, it seems sufficient to do an incremental push. Missing clusters and listeners are pushed out immediately when a late pod event comes in, before a full push is triggered.

dgn:
I finally managed to get this to reproduce reliably by building with `golang==1.11`. Interestingly, when building using `golang>=1.12`, these errors happen much more rarely - but they do still happen. This was very confusing, because adding a debug statement and then building with my local golang installation (`1.12`) would cause the error rate to drop considerably. Average time to reproduction on `golang==1.11` is ~4min for me, whereas with `golang==1.12` I could only ever reproduce once after 90mins

Update: Nevermind, in attempt 3 I now have flaky reproduction again

pilot/pkg/proxy/envoy/v2/eds.go
dgn: My only request here is: if we stick with incremental push, we should adjust the comment above this line that still mentions a full push.

rcernich (approved):

dgn (approved):

pilot/pkg/proxy/envoy/v2/eds.go
dgn: Apparently this is a misunderstanding on my part about incremental vs full push

brian-avery (approved):

knrc:
Updated comments to refer to the related upstream issue and squashed

PR #48, opened 2019-08-30
MAISTRA-862 Stop leaking goroutines in MultiNamespaceListWatcher

dgn:


pkg/listwatch/listwatch.go
```diff
@@ -219,15 +219,23 @@ func (mlw *multiListerWatcher) newMultiWatch(resourceVersions map[string]string,
...
-	wg.Add(len(mlw.lwMap)+1)
+	wg.Add(1)
+
+	// result chan must be closed,
+	// once all event sender goroutines exited.
+	go func() {
+		wg.Wait()
+		close(combinedResult)
+		close(result)
+	}()
```
luksa, line 230: Why did you move this up here? If I see correctly, moving this up here makes no difference, since you're just changing the order of running two go routines. There's going to be just a small difference in when both codes run, but that shouldn't have any impact. And therefore, it's better not to move the code at all, so it's easier to see what the actual fix is.

luksa: Ah, wait. I didn't notice that the `for n, lws := range mlw.lwMap` loop was outside the goroutine...

dgn: Yeah, I had to move this up, otherwise the `Stop()` wouldn't propagate completely, because nobody's waiting for that WaitGroup

dgn:
Marking this WIP for now, I want to implement a unit test next week and then push a PR upstream as well.

dgn:
Closing as this is not the simplest solution. Will push another PR in a bit

PR #49, opened 2019-09-02
MAISTRA-862 Prevent deadlock in case of Watch() errors in subsequent List()

dgn:
If a `Watch()` would fail, the client-go Reflector would attempt to `List()` again - but as we just set the `mlw.state` to `watching`, we would enter this if-block and hit a deadlock, because `Stop()` will try to acquire the same Mutex that we're already holding. I included a test that demonstrates the problem.

This will fix the issue without too many code changes. My other PR would also work, because it explicitly triggered `Stop()` when there was an error, which would then go and change the `mlw.state`, so we would never hit this deadlock. The cleaner, more minimal solution is this one though. I will push the other changes separately.

Upstream is *not* affected by this bug, as they don't have this if-clause and the error reporting logic it is part of.


brian-avery (approved):

pkg/listwatch/listwatch.go
```diff
@@ -134,7 +134,9 @@ func (mlw *multiListerWatcher) List(options metav1.ListOptions) (runtime.Object,
 	defer mlw.lock.Unlock()
 
 	if mlw.state == watching || mlw.state == stopping {
+		mlw.lock.Unlock()
 		mlw.Stop()
+		mlw.lock.Lock()
```
luksa: Hmm, why not split `Stop()` into two functions - a `StopInternal()` which assumes the lock is already held, and `Stop()`, which acquires the lock, calls `StopInternal()` and then releases the lock.  This would allow you to call `StopInternal()` here and not have to release the lock and re-acquire it (twice). 

What  happens if something else acquires the lock in between - is there a potential bug hiding in there?

luksa (changes requested):

pkg/listwatch/listwatch.go
dgn: Good point; that way no other goroutine can slip through the cracks

dgn: Done. Thank you!

luksa (approved):

dgn:
@knrc ping

knrc (approved):

PR #50, opened 2019-09-03
WIP: Stop leaking goroutines in newMultiWatch

dgn:
This fixes a goroutine leak in cases where Watch() is called again without a new List() inbetween Watch() calls. This happens when one of the underlying watches returns a ConnectionRefused error. See https://github.com/kubernetes/client-go/blob/3fe2abece89efc7b3a2da0ff549525f20f435263/tools/cache/reflector.go#L286

Note that I stumbled upon this leak while fixing another bug; this probably has other implications, so I'll mark it WIP for now.

### DIFF
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -263,6 +263,7 @@ func (mlw *multiListerWatcher) newMultiWatch(resourceVersions map[string]string,
 		o.ResourceVersion = resourceVersions[n]
 		w, err := lws.lw.Watch(*o)
 		if err != nil {
+			mlw.stopInternal()
 			return err
 		}
 
--- a/pkg/listwatch/listwatch_test.go
+++ b/pkg/listwatch/listwatch_test.go
@@ -17,6 +17,7 @@ package listwatch
 import (
 	"errors"
 	"k8s.io/api/core/v1"
+	goruntime "runtime"
 	"sync"
 	"testing"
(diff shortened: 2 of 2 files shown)


luksa:
> This happens when one of the underlying watches returns a ConnectionRefused error. 

Note that watches are also closed by the API Server every few minutes. That's another case where Watch() is called again without a new List().

pkg/listwatch/listwatch.go
```diff
@@ -262,6 +262,9 @@ func (mlw *multiListerWatcher) newMultiWatch(resourceVersions map[string]string,
 		o.ResourceVersion = resourceVersions[n]
 		w, err := lws.lw.Watch(*o)
 		if err != nil {
+			mlw.lock.Unlock()
+			mlw.Stop()
+			mlw.lock.Lock()
```
luksa: Same comment as in #49 (don't unlock; call StopInternal() instead)

dgn: Applied the same change. Thanks!

dgn:
> > This happens when one of the underlying watches returns a ConnectionRefused error.
> 
> Note that watches are also closed by the API Server every few minutes. That's another case where Watch() is called again without a new List().

That's correct, but the leak is only triggered when Watch() fails and is not followed by either a call to Stop() or List(). The reflector will in fact call Stop() before calling Watch() again in that case.

PR #51, opened 2019-09-05
MAISTRA-1012 Refactor MemberRollController to be more efficient and less verbose

dgn:
Where previously every component that registered with the controller would add its own handler to the memberRollInformer, we now have a single handler that then broadcasts changes (if there are any) to all registered MemberRollListeners. This reduces some overhead because code paths in the
controller are only executed once instead of n times for every registered component. It also now only logs the namespace update once instead of n times.

### DIFF
--- a/pkg/servicemesh/controller/controller.go
+++ b/pkg/servicemesh/controller/controller.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"fmt"
-	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/servicemesh/controller/controller.go
```diff
@@ -70,78 +75,43 @@ func (smmrc *serviceMeshMemberRollController) Start(stop chan struct{}) {
...
-			smmrl.listener.UpdateNamespaces(smmrl.currentNamespaces)
+		namespaces := sets.NewString(members...)
+		namespaces.Insert(smmrc.namespace)
+		if !smmrc.currentNamespaces.Equal(namespaces) {
+			smmrc.currentNamespaces = namespaces
+			log.Infof("ServiceMeshMemberRoll %v %s, namespaces now %q", memberRollName, operation, smmrc.currentNamespaces.List())
+			for _, listener := range smmrc.listeners {
+				// give everyone their own copy
+				go listener.UpdateNamespaces(smmrc.currentNamespaces.UnsortedList())
+			}
```
rcernich, line 100: Would it be preferable to use a sorted list?  This might provide some consistency when testing.

rcernich (approved):

pkg/servicemesh/controller/controller.go
dgn: I'm not sure that's a good idea, as it would sort the list every time this is called. If we really want that, we should probably cache the sorted list every time the set is updated

dgn:
Marked this as WIP again; will reopen once `maistra-1.2` branch is there

PR #59, opened 2019-10-21
MAISTRA-1005 Do not expect namespaces to have a stable ordering

dgn:
The ServiceMeshMemberRoll controller was previously checking the equality of the existing list of namespaces vs incoming using a simple for-loop that would report inequality in case of different ordering. This now became a problem when a bug in the operator caused it to repeatedly reconcile the MemberRoll, changing the order of namespaces in the process.

Note that this increases complexity of the equality check by quite a margin, but the changes in https://github.com/Maistra/istio/pull/51 could counteract this.

### DIFF
--- a/pkg/servicemesh/controller/controller.go
+++ b/pkg/servicemesh/controller/controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -123,6 +124,7 @@ func (smmrl *serviceMeshMemberRollListener) updateNamespaces(operation string, m
 		log.Infof("ServiceMeshMemberRoll using incorrect name %v, ignoring", memberRollName)
 	} else {
 		namespaces := smmrl.smmrc.getNamespaces(members)
+		sort.Strings(namespaces)
 		if !smmrl.checkEquality(smmrl.currentNamespaces, namespaces) {
 			smmrl.currentNamespaces = namespaces
 			log.Infof("ServiceMeshMemberRoll %v %s, namespaces now %q", memberRollName, operation, smmrl.currentNamespaces)


luksa (approved):
LGTM, but you could also shorten this a lot by using `sets.NewString()` and `HasAll()` or `Equal()` (see vendor/k8s.io/apimachinery/pkg/util/sets/string.go)

dgn:
> LGTM, but you could also shorten this a lot by using `sets.NewString()` and `HasAll()` or `Equal()` (see vendor/k8s.io/apimachinery/pkg/util/sets/string.go)

Good call. I'm using `sets` now. I tried to make the change as non-invasive as possible, I will include a refactored version into my WIP PR that doesn't rebuild the sets on every comparison but stores a set instead of `[]string` in its struct.

dgn:
Was able to confirm that this change will at least keep the control plane sane when the operator goes into the endless reconciliation loop. Still looking at how to fix the operator side

pkg/servicemesh/controller/controller.go
```diff
@@ -104,18 +105,9 @@ type serviceMeshMemberRollListener struct {
...
-	}
-	for n, val := range lhs {
-		if val != rhs[n] {
-			return false
-		}
-	}
-	return true
+	lset := sets.NewString(lhs...)
+	rset := sets.NewString(rhs...)
+	return lset.Equal(rset)
```
knrc: It would be better to handle this in updateNamespaces, then you only need to sort the new list

knrc (changes requested):

pkg/servicemesh/controller/controller.go
dgn: okay, I now applied the bigger change from MAISTRA-1012 here as well; we're storing a `sets.String` instead of `[]string` now, for quicker comparison. I'm sorting the array before logging, otherwise the log output is confusing - we log once per component and they would then come out ordered differently:
```
2019-10-22T14:23:56.612773Z	info	ServiceMeshMemberRoll default added, namespaces now ["test7" "test20" "test2" "test23" "test9" "test10" "test19" "test22" "test18" "test6" "test11" "test21" "test4" "test24" "test" "cp" "test8" "test5"]
2019-10-22T14:23:56.612813Z	info	ServiceMeshMemberRoll default added, namespaces now ["test2" "test23" "test7" "test19" "test20" "test4" "test9" "test18" "cp" "test11" "test22" "test10" "test" "test5" "test6" "test21" "test24" "test8"]
2019-10-22T14:23:56.612837Z	info	ServiceMeshMemberRoll default added, namespaces now ["test10" "test24" "test5" "test7" "test8" "test19" "test11" "test21" "test23" "test9" "test6" "cp" "test20" "test2" "test4" "test22" "test18" "test"]
```

knrc: This seems to be much larger than is necessary, the only required change would be to sort the incoming array before doing the comparison.  I think you can do this with one or two lines.

dgn: Ah. Sorry, was a misunderstanding. That was basically the first iteration of this PR. I can do it that way

knrc: No worries, I think we can do this with just a call to sort.Strings within the updateNamespaces method.  We should only need to do this with the new list as the older one would already be sorted.

dgn: That's how I did it now. Quickly confirmed it's all good on my repro cluster as well

knrc (approved):

PR #67, opened 2019-12-19
[maistra-1.1-future] Pilot

dgn:
This covers all Maistra-specific changes that went into Pilot. I started out trying to re-arrange commits (e.g. separating out the PodLocality work), but because then the following commits wouldn't apply anymore, I just left the order as-is. The only changes I made are about what parts were touched: this PR really only contains Pilot changes.

**This does not yet contain the changes made to the Istio CRDs from https://github.com/Maistra/istio/pull/39 -- that commit changes both Galley and Pilot at the same time, so my plan is to push that one out last, when all other changes to Galley and Pilot have been applied**

### DIFF
--- a/install/kubernetes/helm/istio/charts/pilot/templates/deployment.yaml
+++ b/install/kubernetes/helm/istio/charts/pilot/templates/deployment.yaml
@@ -59,9 +59,9 @@ spec:
 {{- end}}
           - --domain
           - {{ .Values.global.proxy.clusterDomain }}
-{{- if .Values.global.oneNamespace }}
-          - "-a"
-          - {{ .Release.Namespace }}
+{{- if .Values.appNamespace }}
+          - --appNamespace
+          - {{ .Values.appNamespace }}
 {{- end }}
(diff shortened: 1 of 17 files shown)


brian-avery (approved):

rcernich (approved):

rcernich (changes requested):
The code changes look fine, but I think you need to update [clusterrole.yaml](https://github.com/Maistra/istio/blob/maistra-1.1-future/install/kubernetes/helm/istio/charts/pilot/templates/clusterrole.yaml) file to include permissions to get/list/watch SMMR, authentication.maistra.io, and rbac.maistra.io resources.

dgn:
Hey Rob, the permissions for `authentication.maistra.io` and `rbac.maistra.io` will be part of the upcoming PR that re-introduces those CRs. The SMMR permissions are being patched in by the operator scripts: https://github.com/Maistra/istio-operator/blob/6fee5c2ad170d5532976805ffc57bbc73ca9640a/tmp/build/patch-charts.sh#L232

rcernich (approved):

PR #68, opened 2019-12-19
[maistra-1.1-future] Galley

dgn:
This PR has all the changes made to Galley except the ones mentioned below. What's missing is the implementation of multi-tenancy for the new pipelines code, I have not worked on that- if we want to use what's there, we would currently have to disable the new pipeline processing code using a feature flag.

**This does not yet contain the changes made to the Istio CRDs from https://github.com/Maistra/istio/pull/39 -- that commit changes both Galley and Pilot at the same time, so my plan is to push that one out last, when all other changes to Galley and Pilot have been applied**

### DIFF
--- a/galley/pkg/source/kube/builtin/source_test.go
+++ b/galley/pkg/source/kube/builtin/source_test.go
@@ -33,7 +33,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
 )
@@ -67,7 +66,6 @@ func TestNewWithUnknownSpecShouldError(t *testing.T) {
 	defer restoreLogLevel(prevLevel)
 
 	client := fake.NewSimpleClientset()
-	informerFactory := informers.NewSharedInformerFactory(client, 0)
 
 	spec := schema.ResourceSpec{
 		Kind:      "Unknown",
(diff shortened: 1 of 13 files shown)


galley/pkg/source/kube/builtin/source_test.go
```diff
@@ -265,7 +263,10 @@ func TestPods(t *testing.T) {
 		if pod, err = client.CoreV1().Pods(namespace).Create(pod); err != nil {
 			t.Fatalf("failed creating pod: %v", err)
 		}
-		expected := toEvent(resource.Added, spec, pod, pod)
+		// multi namespace lister watcher prepends the namespace used on its creation to the original resource version
+		expectedPod := pod.DeepCopy()
```
brian-avery, line 267: does it make sense to move all of these into a function?

brian-avery (approved):

rcernich (changes requested):
The code changes look fine, but I think you need to update [clusterrole.yaml](https://github.com/Maistra/istio/blob/maistra-1.1-future/install/kubernetes/helm/istio/charts/galley/templates/clusterrole.yaml) file to include permissions to get/list/watch SMMR, authentication.maistra.io, and rbac.maistra.io resources.

dgn:
Hey Rob, the permissions for authentication.maistra.io and rbac.maistra.io will be part of the upcoming PR that re-introduces those CRs. The SMMR permissions are being patched in by the operator scripts: https://github.com/Maistra/istio-operator/blob/6fee5c2ad170d5532976805ffc57bbc73ca9640a/tmp/build/patch-charts.sh#L232

rcernich (approved):

galley/pkg/source/kube/builtin/source_test.go
dgn: Yeah, it might. I just ported what we had though, did not look at refactoring. I'm happy to tackle as part of a follow-up PR

dgn:
@Mergifyio refresh

dgn:
@Mergifyio refresh

dgn:
Not sure why mergify doesn't work. Merging manually

PR #69, opened 2019-12-19
[maistra-1.1-future] Mixer

dgn:


pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+type combinedEvent struct {
+	namespace string
+	event     watch.Event
+}
+
+func newMultiListerWatcher(namespaces []string, f func(string) cache.ListerWatcher) *multiListerWatcher {
+	return &multiListerWatcher{
+		namespaces: append(namespaces[:0:0], namespaces...),
```
brian-avery: what is `namespaces[:0:0]` doing?

dgn: It is basically giving you a fresh slice of the same type as `namespaces`; it not just has length `0` but also capacity `0`, the second part making sure you're not pointing to the same memory. Another way would be defining a new `[]string` and `append`ing to that.

dgn: Note that this comment has since disappeared because the change was coming in from the already-merged MultiListerWatcher PR

brian-avery: Thanks. Odd syntax for that.

brian-avery (approved):

rcernich (changes requested):
The code changes look fine, but I think you need to update [clusterrole.yaml](https://github.com/Maistra/istio/blob/maistra-1.1-future/install/kubernetes/helm/istio/charts/mixer/templates/clusterrole.yaml) file to include permissions to get/list/watch SMMR resources.

dgn:
Hey Rob, the permissions for authentication.maistra.io and rbac.maistra.io will be part of the upcoming PR that re-introduces those CRs. The SMMR permissions are being patched in by the operator scripts: https://github.com/Maistra/istio-operator/blob/6fee5c2ad170d5532976805ffc57bbc73ca9640a/tmp/build/patch-charts.sh#L232

rcernich (approved):

dgn:
Merging by hand because mergify won't

PR #72, opened 2019-12-20
[maistra-1.1-future] sidecar-injector

dgn:
I left out the commit adding annotation functionality, as that has already been implemented upstream - **we will have to look out for this in the operator repository, as the way to specify annotations is slightly different**: `injectedAnnotations` instead of `annotations`. For the UID injection I had to change a few more of the unit tests to make them pass.


brian-avery (approved):

rcernich (approved):
lgtm

PR #74, opened 2020-01-06
MAISTRA-1059: fix failing tests

knrc:


dgn (approved):
tests certainly work now, apparently there's two filterChains (it also works with index 1) of which the latter is the passthrough, but I'm not sure why this broke?

knrc:
+1, I have used -1 as the index as the filter chain being tested should always be the last regardless of how many more are added

knrc:
Merging by hand for the release

PR #77, opened 2020-02-07
MAISTRA-1159 Add missing label "release" to CRDs

luksa:
Some of CRDs are missing the "release" label. Our `patch-charts.sh` script only adds the `maistra-release` label to all objects that have the `release` label. The end result is that these CRDs don't have the `maistra-release` label, which is crucial for multi-version CRD handling.


jwendell (approved):
Does 1.1 branch (Istio 1.4) suffer from this as well?

luksa:
> Does 1.1 branch (Istio 1.4) suffer from this as well?

No.

dgn (approved):

PR #81, opened 2020-02-21
MAISTRA-1079: Replace deprecated API's (part 2)

jwendell:
Add selectors to deployments. They were not necessary previously because
they were added by operator patch charts script.

They are needed here now because we run E2E tests that grab files directly
from this repo - and not from operator charts.


knrc (approved):

rcernich (commented):
You should verify that this is actually required because I believe we do a lot of this in istio-operator patch-scripts.sh.  I'll approve this after verification.

rcernich (approved):
This is related to maistra/istio-operator#349, which modifies patch-charts.sh

PR #100, opened 2020-04-17
Ensure bookinfo example works on Maistra

luksa:

### DIFF
--- a/samples/bookinfo/add_inject_annotation.sh
+++ b/samples/bookinfo/add_inject_annotation.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -o errexit
+
+#Add sidecar.istio.io/inject annotation
+find . -name "*bookinfo*.yaml" -exec sed -i.bak 's/      \(labels:.*\)/      annotations:\
+        sidecar.istio.io\/inject: "true"\
+      \1/g' {} +
+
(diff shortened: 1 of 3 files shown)


luksa:
Please merge https://github.com/Maistra/istio/pull/101 first.

samples/bookinfo/add_inject_annotation.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
jwendell, line 1: Why not change the files directly? Those files don't change that much, and, since we are a fork, it makes sense to change the files directly and get rid of those sed black magics...

jwendell (changes requested):

samples/bookinfo/add_inject_annotation.sh
luksa: I added this script so we can then simply pull in the files from upstream and run these two scripts to make the necessary changes without having to think about merging.

jwendell: Those are trivial files so I'm confident we don't have to think about merges. git will do its job pretty easily :)

luksa: Honestly, I feel it's too easy to miss a detail like this annotation. It happened to me today and left me wondering why injection stopped working all of a sudden. That's why having some kind of centralized "here's how we need to modify the upstream files to get a working version" file. 

I'd actually also add a {{Makefile}} to automate the whole process.

luksa:
I'll recreate this PR from a proper branch.

PR #104, opened 2020-04-21
MAISTRA-1218 Rebase bookinfo changes

luksa:


jwendell (approved):

dgn (approved):

dgn:
Do we have/need a process that actually triggers the build and pushes these images?

luksa:
@dgn We don't. It would be nice to have if we want to provide images for each patch version. If we're ok with one set of images per minor version, we can just do it manually.

PR #116, opened 2020-06-02
MAISTRA-1374 Whitelist the k8s.v1.cni.cncf.io/networks annotation for…

dgn:
…templating


jwendell (approved):
please submit it upstream as well

dgn:
Pulling back as @knrc already has a fix in the pipeline

PR #117, opened 2020-06-05
MAISTRA-1374: support multiple multus CNI interfaces

knrc:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/install/kubernetes/helm/istio/files/injection-template.yaml
+++ b/install/kubernetes/helm/istio/files/injection-template.yaml
@@ -453,6 +453,13 @@ dnsConfig:
     {{- end }}
 {{- end }}
 podRedirectAnnot:
+{{- if and (.Values.istio_cni.enabled) (ne .Values.istio_cni.istio_cni_network "") }}
+{{ if isset .ObjectMeta.Annotations `k8s.v1.cni.cncf.io/networks` }}
+   k8s.v1.cni.cncf.io/networks: "{{ index .ObjectMeta.Annotations `k8s.v1.cni.cncf.io/networks`}}, {{.Values.istio_cni.istio_cni_network}}"
+{{- else }}
+   k8s.v1.cni.cncf.io/networks: "{{ .Values.istio_cni.istio_cni_network }}"
+{{- end }}
+{{- end }}
    sidecar.istio.io/interceptionMode: "{{ annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode }}"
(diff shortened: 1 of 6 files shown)


install/kubernetes/helm/istio/values.yaml
```diff
@@ -96,6 +96,7 @@ certmanager:
 #
 istio_cni:
   enabled: false
+  istio_cni_network: "istio-cni"
```
jwendell, line 99: If this is a new field (i.e., we don't need to keep api), why not just call it `network`? It's already within a `istio_cni` structure.

pkg/kube/inject/inject_test.go
```diff
@@ -121,6 +122,19 @@ func TestIntoResourceFile(t *testing.T) {
 			readinessFailureThreshold:    DefaultReadinessFailureThreshold,
 			enableCni:                    true,
 		},
+		// verify cni network
```
jwendell, line 125: I'm curious, in the case above, shouldn't the file `hello.yaml.cni.injected` contain the annotation with the default value?

install/kubernetes/helm/istio/values.yaml
knrc: It's not a new field, it's already in use within the operator

pkg/kube/inject/inject_test.go
knrc: No, the default values are not consumed in these tests

jwendell (approved):

install/kubernetes/helm/istio/files/injection-template.yaml
```diff
@@ -453,6 +453,13 @@ dnsConfig:
     {{- end }}
 {{- end }}
 podRedirectAnnot:
+{{- if and (.Values.istio_cni.enabled) (ne .Values.istio_cni.istio_cni_network "") }}
+{{ if isset .ObjectMeta.Annotations `k8s.v1.cni.cncf.io/networks` }}
+   k8s.v1.cni.cncf.io/networks: "{{ index .ObjectMeta.Annotations `k8s.v1.cni.cncf.io/networks`}}, {{.Values.istio_cni.istio_cni_network}}"
```
dgn, line 458: just out of curiosity: is the order important here? in my fix I had this the other way around

dgn (approved):

luksa (approved):

install/kubernetes/helm/istio/files/injection-template.yaml
knrc: I believe it is applied in order so the Istio CNI should be last to run

PR #123, opened 2020-06-17
MAISTRA-1301 Make sure integration tests run with latest toolchain

dgn:


prow/lib.sh
```diff
@@ -160,13 +160,13 @@ function setup_kind_cluster() {
...
   # Create KinD cluster
-  if ! (kind create cluster --name="${NAME}" --config "${CONFIG}" --loglevel debug --retain --image "${IMAGE}" --wait=60s); then
+  if ! (kind create cluster --name="${NAME}" --config "${CONFIG}" --retain --image "${IMAGE}" --wait=60s); then
     echo "Could not setup KinD environment. Something wrong with KinD setup. Exporting logs."
     exit 1
   fi
 
-  KUBECONFIG="$(kind get kubeconfig-path --name="${NAME}")"
-  export KUBECONFIG
+  # KUBECONFIG="$(kind get kubeconfig-path --name="${NAME}")"
```
jwendell: why?

prow/lib.sh
```diff
@@ -64,7 +64,7 @@ function kind_load_images() {
   for i in {1..3}; do
     # Archived local images and load it into KinD's docker daemon
     # Kubernetes in KinD can only access local images from its docker daemon.
-    docker images "${HUB}/*:${TAG}" --format '{{.Repository}}:{{.Tag}}' | xargs -n1 kind --loglevel debug --name "${NAME}" load docker-image && break
+    docker images "${HUB}/*:${TAG}" --format '{{.Repository}}:{{.Tag}}' | xargs -n1 kind --name "${NAME}" load docker-image && break
```
jwendell: don't like the verbosity? does it hurt?

prow/lib.sh
dgn: This call is not required anymore in current kind versions - in fact it will crash the script. kind automatically sets $KUBECONFIG after successfully creating a cluster. I changed this to actually remove the code instead of commenting

prow/lib.sh
dgn: removing it is actually not the same, that's true. I replaced it with `-v9` which is the equivalent on current kind versions

tests/integration/tests.mk
```diff
@@ -91,7 +91,7 @@ test.integration.%.local: | $(JUNIT_REPORT)
 	2>&1 | tee >($(JUNIT_REPORT) > $(JUNIT_UNIT_TEST_XML))
 
 # TODO: Exclude examples and qualification since they are very flaky.
-TEST_PACKAGES = $(shell go list ./tests/integration/... | grep -v /qualification | grep -v /examples)
+TEST_PACKAGES = $(shell go list ./tests/integration/... | grep -v /qualification | grep -v /examples | grep -v /istioio)
```
dgn, line 94: this is added so that `istioio` tests are skipped (they are flaky)

jwendell (approved):

knrc (approved):

tvieira (approved):
lgtm

PR #124, opened 2020-06-23
Run sidecars with dynamically-determined user IDs

bison:
Cherry-pick for 1.6.0 rebase including the following:

### Run sidecars with dynamically-determined user IDs


dgn (approved):
CNI tests seem to pass now

PR #135, opened 2020-07-07
MAISTRA-1400: Add IOR to Pilot

jwendell:


bison:
I haven't had time to read through this yet, but is there a good reason to put this into Pilot? In general, it would be great if we could keep any mostly self-contained components that are Maistra / OpenShift specific separate. I think we should be trying to shrink the diff between upstream and Maistra, otherwise rebases are going to continue to be a nightmare. Personally, I'd prefer a bit of extra complexity in our operator and deployment bits if it means a smaller diff from upstream in the core components.

jwendell:
@bison I've written the rationale of putting this in pilot in MAISTRA-1400.

Also, if there are no objections from our side, I'm planning to upstream this as well.

bison:
Ah, okay. If we're upstreaming this, I have no problem with it going into Pilot.

dgn:
In the `make gen` commit there are a lot of whitespace, newline and import order changes - are those accidental? I'd prefer changes like that to happen in a separate PR

jwendell:
I'll rebase/resubmit this PR. This PR is what motivated me to fix all lint and gen-chek failures in maistra/istio.
Now that everything is merged I can proceed with a rebase for this one. Thanks for reminding me.

operator/pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -538,6 +538,7 @@ message GlobalConfig {
 
   // Controls whether one central istiod is enabled.
   google.protobuf.BoolValue centralIstiod = 62;
+
```
brian-avery: why change the proto?

brian-avery (approved):

operator/pkg/apis/istio/v1alpha1/values_types.proto
jwendell: yeah, this is a leftover from previous interaction. I've noticed it but will only remove it after this one gets approved - trying to make eventual changes in batch.

knrc (approved):

PR #138, opened 2020-07-10
MAISTRA-1627, MAISTRA-1660: Fix lint, gen-check and integration test failures

jwendell:
This PR covers a bunch of fixes for failures:
- Lint failures: now `make lint` works (MAISTRA-1627)
- Gen failures: now `make gen-check` works (MAISTRA-1627)
- Integration test failures: now integration tests work (MAISTRA-1660)

For the last item we had to bring the fix from upstream release-1.6 branch, hence some commits not related to this PR itself. This is to avoid cherry-picking just one commit and have to resolve conflicts later.


knrc (approved):

pkg/listwatch/listwatch_test.go
```diff
@@ -456,7 +456,7 @@ func TestMultiWatchDoesNotDeadlock(t *testing.T) {
 			},
 			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
 				if namespace == "4" {
-					return nil, errors.New("Some error occurred")
+					return nil, errors.New("some error occurred")
```
brian-avery, line 459: is there a reason we don't return the error? Also, would it be better to say "an error occurred"?

pkg/servicemesh/apis/servicemesh/v1/doc.go
```diff
@@ -1,3 +1,17 @@
+// Copyright Istio Authors
```
brian-avery: @knrc who owns the copyright for code that we write in our fork of Istio? Files that we add? Should this be copyright Istio or copyright Red Hat? Or even copyright Maistra?

brian-avery (changes requested):
Leaving this open until questions are answered.

knrc (changes requested):
Holding until we have Brian's question answered

pkg/listwatch/listwatch_test.go
knrc: If this is not code we have authored then we should keep it matching upstream, the only possible link error would be the capitalisation.

brian-avery: oh. Just realized this is in test code anyway. But I think we added the listwatcher for the multitenancy work didn't we?

knrc: No, they had already copied a version of it from the prometheus codebase

dgn (approved):

pkg/servicemesh/apis/servicemesh/v1/doc.go
knrc: This does need to change, the answer is that we should not be using "Copyright Istio Authors" unless it's part of the upstream

jwendell:
Fixed the Copyright

jwendell:
All tests pass now. Turns out integration tests had an issue. Created MAISTRA-1660 for that.
Pulled in the fix for it from upstream 1.6 branch in this PR so that all tests pass and we have a single PR fixing all test/lint failures.

knrc:
@jwendell This PR seems to include more than just lint and gen-check failures, can these be separated out into individual PRs?

jwendell:
no. If we split it, tests will fail for both PR's. Unless we force-merge them.

knrc:
@jwendell  please create a second JIRA to cover the additional work and include it in the description of this PR, we should not be sneaking in additional changes just for expedience.

knrc (changes requested):
Requesting changes to cover additional commits included within this PR

jwendell:
@knrc done

knrc (approved):
@jwendell thanks very much

brian-avery (approved):

PR #156, opened 2020-08-06
fixes to get istiod working with constrained permissions

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
-	informer "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -76,7 +75,7 @@ func NewNamespaceController(data func() map[string]string, options Options, kube
 
 	watchedNamespaceList := strings.Split(options.WatchedNamespaces, ",")
 
-	mlw := listwatch.MultiNamespaceListerWatcher(watchedNamespaceList, func(namespace string) cache.ListerWatcher {
+	configmapmlw := listwatch.MultiNamespaceListerWatcher(watchedNamespaceList, func(namespace string) cache.ListerWatcher {
 		return &cache.ListWatch{
 			ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
 				opts.LabelSelector = fields.SelectorFromSet(configMapLabel).String()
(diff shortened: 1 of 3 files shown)


pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -132,7 +131,21 @@ func NewNamespaceController(data func() map[string]string, options Options, kube
 	})
 	c.configMapController = configmapInformer
 
-	namespaceInformer := informer.NewNamespaceInformer(kubeClient, options.ResyncPeriod, cache.Indexers{})
+	namespacemlw := listwatch.MultiNamespaceListerWatcher(watchedNamespaceList, func(namespace string) cache.ListerWatcher {
```
dgn, line 134: This won't work as namespaces are cluster-scoped. I think we'll need to remove the watching of namespaces entirely

dgn: Looks like this is used for watching namespaces and creating/updating a configMap with a certificate. We should probably move that into the operator. I filed https://issues.redhat.com/browse/MAISTRA-1724 to take care of it

dgn (commented):
I think we should split this into two PRs as it's covering two separate permissions; nodes and namespaces. Namespaces are a bit more complicated (see my inline comment). The part about the nodeInformer should be straightforward as we have that functionality in the operator already.

I created https://issues.redhat.com/browse/MAISTRA-1722 to track any permission issues we find. I think there's no rush for now; these can be fixed in later sprints if need be

bison:
I called out one of the Node informers after the rebase. More info on that in [MAISTRA-1622](https://issues.redhat.com/browse/MAISTRA-1622). I think that change is partly to support exposing gateways via NodePort. Maybe we don't support that, but we should probably document that.

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -76,7 +75,7 @@ func NewNamespaceController(data func() map[string]string, options Options, kube
 
 	watchedNamespaceList := strings.Split(options.WatchedNamespaces, ",")
 
-	mlw := listwatch.MultiNamespaceListerWatcher(watchedNamespaceList, func(namespace string) cache.ListerWatcher {
+	configmapmlw := listwatch.MultiNamespaceListerWatcher(watchedNamespaceList, func(namespace string) cache.ListerWatcher {
```
bison, line 78: This seems unrelated.

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
rcernich: Yeah, I figured this was pretty cheesy.  I think we had something like this in 1.1 though, although maybe mlw worked differently there (that's where I pulled it from).  I was considering adding a listener to the member roll controller, but that required a bit more work to hook it in. (Basically, have the listener add events to the queue for each namespace in the list.)

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
rcernich: Because I added another mlw, I simply made the name more descriptive.

rcernich:
Yeah, I'm not sure why it's adding the node IPs for the gateways into the service registry.  Maybe this is for multicluster?  If so, maybe we should be watching routes instead?  (Although, that probably won't work, as you can't access the service using an IP, without specifying SNI host too, and that means we need to get the host IP for the router anyway.)

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
bison: Sure. I was mostly thinking that it was an unnecessary diff to carry, but this has all vanished upstream anyway.

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
dgn: Yeah, that probably makes the most sense - use the NamespaceUpdated events from the MRC and pipe it into this ConfigMap-creating code

PR #157, opened 2020-08-07
MAISTRA-1739 Fix validation of AuthorizationPolicy fields

dgn:
It would previously detect the field `request.regex.headers` as invalid, even though it is supported.

I'm adding a hold as this will only go into 1.1.8

### DIFF
--- a/pkg/config/security/security.go
+++ b/pkg/config/security/security.go
@@ -33,23 +33,24 @@ type JwksInfo struct {
 }
 
 const (
-	attrRequestHeader    = "request.headers"        // header name is surrounded by brackets, e.g. "request.headers[User-Agent]".
-	attrSrcIP            = "source.ip"              // supports both single ip and cidr, e.g. "10.1.2.3" or "10.1.0.0/16".
-	attrSrcNamespace     = "source.namespace"       // e.g. "default".
-	attrSrcUser          = "source.user"            // source identity, e.g. "cluster.local/ns/default/sa/productpage".
-	attrSrcPrincipal     = "source.principal"       // source identity, e,g, "cluster.local/ns/default/sa/productpage".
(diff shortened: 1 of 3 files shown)


jwendell (commented):
Can we add a unit or integration test to validate this feature?

dgn:
Good point. I added a test

pkg/config/security/security.go
```diff
@@ -33,23 +33,24 @@ type JwksInfo struct {
...
-	attrDestIP           = "destination.ip"         // supports both single ip and cidr, e.g. "10.1.2.3" or "10.1.0.0/16".
-	attrDestPort         = "destination.port"       // must be in the range [0, 65535].
-	attrDestLabel        = "destination.labels"     // label name is surrounded by brackets, e.g. "destination.labels[version]".
-	attrDestName         = "destination.name"       // short service name, e.g. "productpage".
-	attrDestNamespace    = "destination.namespace"  // e.g. "default".
-	attrDestUser         = "destination.user"       // service account, e.g. "bookinfo-productpage".
-	attrConnSNI          = "connection.sni"         // server name indication, e.g. "www.example.com".
-	attrExperimental     = "experimental.envoy.filters."
+	attrRequestHeader      = "request.headers"        // header name is surrounded by brackets, e.g. "request.headers[User-Agent]".
+	attrRequestHeaderRegex = "request.regex.headers"  // header name is surrounded by brackets, e.g. "request.headers[User-Agent]".
```
luksa: Comment is the same as in the previous line - should be `request.regex.headers[User-Agent]`

luksa (approved):
LGTM. Just one minor fix needed in code comment.

pkg/config/security/security.go
dgn: Whoops. Thanks!

jwendell (approved):

PR #159, opened 2020-08-11
MAISTRA-1741 Store resourceVersions locally instead of returning them

dgn:
This will resolve the errors caused by changes in the cache reflector code. Returning resourceVersion "0" will make sure we're getting the latest object versions from cache when re-List()ing. When Watch()ing, we're injecting the stored resourceVersions we observed during List().

Note that this is merely a stop-gap solution that'll get things working again until we have a better implementation of MLW (or rather, an informer-based rewrite). Storing the resourceVersions is safe because we only do it while holding the mutex.


dgn:
Note that another solution would be to return always the highest `resourceVersion` in the List() function. This works because resourceVersion is a global counter; that would be against the [official advice](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency) not to assume a specific format of resourceVersion or any other implementation-specific information though, so I decided against doing that.

rcernich (approved):

knrc (approved):

brian-avery (approved):

PR #160, opened 2020-08-11
MAISTRA-1745 Don't watch namespaces if MRC is used

dgn:
This removes the need for namespace permissions by reacting to the `UpdateNamespaces()` calls from the MRC instead of watching namespaces directly on the API server.


dgn:
Oops, wrong ticket number

PR #161, opened 2020-08-11
MAISTRA-1724 Don't watch namespaces if MRC is used

dgn:
This will remove the need for namespace permissions if the MRC is used. Instead of creating a controller that watches namespaces on the API server, we only react to the `UpdateNamespaces()` calls from the MRC.


rcernich (approved):

jwendell (approved):
nit: Whenever possible use full names, not acronyms, it took me a while to figure out what MRC is :)

PR #162, opened 2020-08-11
MAISTRA-1751 Set securityContext.fsGroup to comply with restricted SCC

dgn:
`securityContext.fsGroup` needs to be set correctly to support `third-party-jwt` using a projected token volume


rcernich (approved):

dgn:
Note that I found some problems with just removing this, so I'm marking as WIP for now

dgn:
Problems resolved. This now works both for the upstream integration tests as well as on OCP with the restricted SCC

jwendell (approved):

PR #167, opened 2020-08-13
MAISTRA-1755 Always create istio-ca-root-cert in control plane namespace

dgn:


jwendell (approved):

rcernich:
@dgn, I think `UpdateNamespaces()` should be called as part of `MemberRollController.Register()` to ensure that registered listeners are initialized properly, otherwise, they won't be initialized with a namespace list until the SMMR has changed.  (Local control plane workspace is added by default, regardless of whether or not an SMMR is present, which takes care of the case where istiod needs to create root cert ca secret in its namespace so other components can become ready.)

https://github.com/maistra/istio/pull/169

dgn:
Closing in favor of https://github.com/maistra/istio/pull/169

PR #180, opened 2020-09-29
MAISTRA-1723: Allow disabling NodePort gateway support

bison:
This adds an option to istiod that can disable the support for
NodePort gateways.  With this disabled, istiod will not create an
informer for Node objects, which removes the need for cluster-wide
read permissions on these.

### DIFF
--- a/pilot/cmd/pilot-discovery/main.go
+++ b/pilot/cmd/pilot-discovery/main.go
@@ -149,6 +149,8 @@ func init() {
 	if err := discoveryCmd.PersistentFlags().MarkDeprecated("appNamespace", "please use ${APP_NAMESPACE} environment variable instead"); err != nil {
 		panic(err)
 	}
+	discoveryCmd.PersistentFlags().BoolVar(&serverArgs.Config.ControllerOptions.EnableNodePortGateways, "enableNodePortGateways",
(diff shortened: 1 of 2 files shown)


pilot/cmd/pilot-discovery/main.go
```diff
@@ -149,6 +149,8 @@ func init() {
 	if err := discoveryCmd.PersistentFlags().MarkDeprecated("appNamespace", "please use ${APP_NAMESPACE} environment variable instead"); err != nil {
 		panic(err)
 	}
+	discoveryCmd.PersistentFlags().BoolVarP(&serverArgs.Config.ControllerOptions.EnableNodePortGateways, "enableNodePortGateways", "",
+		false, "Whether to watch nodes and advertise their addresses as endpoints for gateways with NodePort services")
```
jwendell: You could have it `true` by default - so that it doesn't change current behavior - and we could set that value to false in our charts. Thus you can upstream this patch.

bison: Sure. I wasn't really sure upstream would be interested in it. It feels like whack-a-mole until there is any kind of policy around cluster-scope permissions upstream, but we can try.

bison: Updated.

jwendell (approved):

dgn (approved):

PR #75, opened 2020-01-09
[maistra-1.1] MAISTRA-630 Add support for ServiceMeshPolicy and ServiceMeshRbacConfig

dgn:
This ports the work done in #39

PR #186, opened 2020-10-05
MAISTRA-1856: Disable ClusterRbacConfig resources

bison:
I'm actually not sure what all needs to be done here. This doesn't remove every bit that references the deprecated ClusterRbacConfig resource, but I think it's enough to keep istiod from trying to watch it, which I think is good enough.


bison:
Previous PRs: #39 and #75

luksa (approved):

dgn:
There are a few integration tests that need to be disabled/removed, other than that this looks good

dgn (approved):

bison:
Yeah, missed those. All fixed.

PR #188, opened 2020-10-07
MAISTRA-1897: Allow disabling IngressClass support

bison:
This adds an option to istiod that allows disabling reading
IngressClass resources, which are cluster-scoped.  When this is
disabled, processing of Ingress resources that use the new (as of
Kubernetes v1.18) `ingressClassName` field in their spec is not
supported.  Only resources that use the deprecated annotation will be
processed.


luksa (approved):

dgn (approved):
It's a bit unfortunate that we can't support it, but we might be able to work something out for future releases. The annotations are still going to be around until the K8s 1.22 release

PR #191, opened 2020-10-09
MAISTRA-1924 Fix handling of resource deletions

dgn:


pilot/pkg/config/kube/crd/controller/controller.go
```diff
@@ -209,20 +209,26 @@ func (c *controller) checkReadyForEvents(curr interface{}) error {
 }
 
 func (c *controller) tryLedgerPut(obj interface{}, schema collection.Schema) {
-	iobj := obj.(crd.IstioObject)
+	iobj, err := castToIstioObject(obj)
+	if err != nil {
+		scope.Errorf("Failed to cast to crd.IstioObject in ledger put: %v", err)
+	}
```
rcernich, line 216: Shouldn't there be a return or something here?  the object failed to cast, so the var will be nil, which is probably what's causing the panic in the unit test.

pilot/pkg/config/kube/crd/controller/controller.go
```diff
@@ -209,20 +209,26 @@ func (c *controller) checkReadyForEvents(curr interface{}) error {
...
 		scope.Errorf("Failed to update %s in ledger, status will be out of date.", key)
 	}
 }
 
 func (c *controller) tryLedgerDelete(obj interface{}, schema collection.Schema) {
-	iobj := obj.(crd.IstioObject)
+	iobj, err := castToIstioObject(obj)
+	if err != nil {
+		scope.Errorf("Failed to cast to crd.IstioObject in ledger delete: %v", err)
+	}
```
rcernich, line 230: Ditto, return?

pilot/pkg/config/kube/crd/controller/controller.go
dgn: Oops. Yes, of course. Will push a fix. Thanks!

rcernich (approved):

bison (approved):

PR #205, opened 2020-11-06
Support for HTTP1.1 Only

twghu:

Issue raised by https://github.com/istio/istio/issues/14708

"Is it possible to disable http2 and always use http 1.1"

This change allows for istio to configure envoy to use "h2" only, "http/1.1" only or default to the "h2,http/1.1" and relates to problems encountered with the use of wildcard certificates for multiple backend services.  Including the problem with sticky sessions when using h2 and browsers for example with the OCP Console.  

[x] Configuration Infrastructure
Change to deployment template.
[x] Docs
      tls_context:
         common_tls_context:
           tls_certificates:
            certificate_chain: { filename: „/etc/example-com.crt“ }
             private_key: { filename: „/etc/example-com.key“ }
           alpn_protocols: [ "http/1.1" ]
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/pkg/features/alpn_protocols.go
+++ b/pkg/features/alpn_protocols.go
@@ -1,0 +1,70 @@
+package features
+
+import (
+	"istio.io/istio/pilot/pkg/networking/util"
+	"istio.io/pkg/env"
+	"istio.io/pkg/log"
+	"strings"
+
+	//"strings"
+	"sync"
+)
+
+var (
+	ALPNProtocols = RegisterALPNProtocolsVar(
+		"ALPN_PROTOCOLS",
+		"",
+		"The supported ALPN Protocols",
+	)
+	elock = &sync.Mutex{}
+)
+
+type ALPNProtocolsVar struct {
+	env.StringVar
+	protocols []string
+}
+
+func RegisterALPNProtocolsVar(name string, defaultValue string, description string) ALPNProtocolsVar {
+	v := env.RegisterStringVar(name, defaultValue, description)
+	return ALPNProtocolsVar{v, nil }
+}
+
+func (v *ALPNProtocolsVar) initAlpnProtocols() {
+	lock.Lock()
+	defer lock.Unlock()
+	if v.protocols == nil {
+		protocols := []string{}
+		protocolsParam, _ := v.Lookup()
+		if protocolsParam != "" {
+			protocolsParamSlice := strings.Split(protocolsParam, ",")
+			for _, protocolsParam := range protocolsParamSlice {
+				trimmed := strings.Trim(protocolsParam, " ")
+				// ensure only supported values are accepted
+				if trimmed == util.ALPNH11Only[0] || trimmed == util.ALPNH2Only[0] {
+					protocols = append(protocols, trimmed)
+				} else {
+					log.Warnf("ALPN Protocol %v is not supported, this entry will be ignored", trimmed)
+				}
+			}
+		}
+		if len(protocols) == 0 {
+			v.protocols = util.ALPNHttp
+		} else {
+			v.protocols = protocols
+		}
+		log.Infof("ALPN Protocol is %v", v.protocols)
+	}
+}
+
+func (v *ALPNProtocolsVar) Reset() {
+	elock.Lock()
+	defer elock.Unlock()
+	v.protocols = nil
+}
+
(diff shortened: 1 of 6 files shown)


twghu:
Should we include this in 2.0 as well. 

Istio 1.5  introduced this:  

https://docs.openshift.com/container-platform/4.5/networking/ingress-operator.html#nw-http2-haproxy_configuring-ingress

pkg/features/alpn_protocols.go
```diff
@@ -0,0 +1,70 @@
+package features
+
+import (
+	"istio.io/istio/pilot/pkg/networking/util"
+	"istio.io/pkg/env"
+	"istio.io/pkg/log"
+	"strings"
+
+	//"strings"
```
dgn: nit: commented out code

pkg/features/alpn_protocols.go
```diff
@@ -0,0 +1,70 @@
...
+			v.protocols = util.ALPNHttp
+		} else {
+			v.protocols = protocols
+		}
+		log.Infof("ALPN Protocol is %v", v.protocols)
+	}
+}
+
+func (v *ALPNProtocolsVar) Reset() {
+	elock.Lock()
```
dgn: shouldn't this mutex be included in the ALPNProtocolsVar struct?

pkg/features/alpn_protocols.go
```diff
@@ -0,0 +1,70 @@
...
+	protocols []string
+}
+
+func RegisterALPNProtocolsVar(name string, defaultValue string, description string) ALPNProtocolsVar {
+	v := env.RegisterStringVar(name, defaultValue, description)
+	return ALPNProtocolsVar{v, nil }
+}
+
+func (v *ALPNProtocolsVar) initAlpnProtocols() {
+	lock.Lock()
```
dgn: this seems to be using the wrong mutex, should probably be `elock`?

pkg/features/alpn_protocols.go
```diff
@@ -0,0 +1,70 @@
...
+	"sync"
+)
+
+var (
+	ALPNProtocols = RegisterALPNProtocolsVar(
+		"ALPN_PROTOCOLS",
+		"",
+		"The supported ALPN Protocols",
+	)
+	elock = &sync.Mutex{}
```
dgn: nit: confusing name, maybe alpnProtocolLock?

In fact, as `RegisterALPNProtocolsVar()` is only called once, in this file, the mutex should probably live in the struct that is returned by that func

dgn (commented):
we'll need a JIRA issue to track this

pamoedom:
Hi all, any progress here?

Thanks and regards.

pkg/features/alpn_protocols.go
knrc: Yes, and it's a mistake I made in the ECDH curve case which is where this has come from.  The method is only used during testing although that's not relevant, it should be fixed here and another jira created to address tls_ecdh_curves.go

pkg/features/alpn_protocols.go
knrc: A stylistic change but it wouldn't hurt, if we do this then we should open another issue to make the same change to the others.

pkg/features/alpn_protocols.go
knrc: A stylistic change but it wouldn't hurt, if we do this then we should open another issue to make the same change to the others.

brian-avery (approved):

pkg/features/alpn_protocols.go
brian-avery: @twghu can you please address this?

brian-avery:
Might also be better to target this at 2.0.

pamoedom:
Hi all, any update here?

pkg/features/alpn_protocols.go
twghu: updated.

pkg/features/alpn_protocols.go
twghu: changed

jwendell:
@twghu Is this safe to close? If we still need this feature, can it be done in 2.1 instead? Also, please create a Jira ticket for it.

PR #206, opened 2020-11-13
MAISTRA-1993 Add unit tests for ServiceMeshExtensions; MAISTRA-1948 Fix URL generation and parsing in wasm-cacher

dgn:
Had to put this into one PR to use the new test harness for the fix


mec/pkg/model/types.go
```diff
@@ -51,25 +53,52 @@ func (i *ImageRef) String() string {
 	return fmt.Sprintf("%s/%s:%s", i.Hub, i.Repository, i.Tag)
 }
 
-func StringToImageRef(ref string) *ImageRef {
-	colonSplit := strings.SplitN(ref, ":", 2)
-	tag := colonSplit[1]
-	slashSplit := strings.Split(colonSplit[0], "/")
-	repo := slashSplit[len(slashSplit)-1]
-	hub := strings.Join(slashSplit[:len(slashSplit)-1], "/")
+func StringToImageRef(ref string) (result *ImageRef) {
```
jwendell, line 56: perhaps return an `error` as well?

this looks like a common function for every project that deals with images... probably it's already implemented somewhere else? perhaps `github.com/containers/image`?

mec/cmd/mec/main.go
```diff
@@ -95,7 +94,10 @@ func createCommand(args []string) *cobra.Command {
 				log.Errorf("Failed to create OSSMPullStrategy: %v", err)
 			}
 			w := server.NewWorker(config, p, baseURL, serveDirectory)
-
+			if w == nil {
```
jwendell: probably not on the scope of this pr, but we could switch to `RunE` and return proper errors.

mec/pkg/model/types_test.go
```diff
@@ -0,0 +1,76 @@
...
+			input: "registry.io/corp/repo@sha256:41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3",
+			output: &ImageRef{
+				Repository: "repo",
+				Hub:        "registry.io/corp",
+				SHA256:     "41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3",
+			},
+		},
+		{
+			input:  "registry.io/corp/repo@sha256:af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3",
+			output: nil,
```
jwendell, line 46: if that function returned an error, you could expect an error here and in the others

jwendell (approved):
This looks great, made minor comments.

mec/cmd/mec/main.go
```diff
@@ -95,7 +94,10 @@ func createCommand(args []string) *cobra.Command {
...
+				log.Errorf("Failed to create worker: %v", err)
```
brian-avery, line 98: What's the error you're reporting here? `server.NewWorker` doesn't return an error. Looks like that would be the error for `NewOSSMPullStrategy` since that doesn't return on error?

mec/pkg/model/types.go
```diff
@@ -29,19 +29,21 @@ type ImageRef struct {
...
 
 type Manifest struct {
-	SchemaVersion ManifestSchemaVersion `json:"schemaVersion"`
-	Name          string                `json:"name"`
-	Description   string                `json:"description"`
-	Version       string                `json:"version"`
-	Phase         v1alpha1.FilterPhase  `json:"phase"`
-	Priority      int                   `json:"priority"`
-	Module        string                `json:"module"`
+	SchemaVersion ManifestSchemaVersion `yaml:"schemaVersion"`
```
brian-avery, line 32: not sure if it matters, but I think all of the common yaml libraries use the `json` and `yaml` tags equivalently.

mec/pkg/model/types.go
dgn: yeah I have that on my list of things to check, especially as this thing is growing :) I found https://github.com/containers/image/blob/master/docker/reference which looks to be what we need. I'd like to put that into the 2.1 bucket though, because it'll require changes all over the code base to use the new reference format

mec/pkg/model/types.go
dgn: not quite, apparently. the unmarshalling of `ManifestSchemaVersion` didn't work previously. I suspect the reason being that it's a number in string format, e.g. "2", so in the file it looks like:
```yaml
schemaVersion: 1
```

which works in yaml but not in json (it would detect it as int)

mec/cmd/mec/main.go
dgn: good catch. will fix

mec/pkg/model/types.go
dgn: I filed https://issues.redhat.com/browse/MAISTRA-2006 to track it

mec/pkg/model/types.go
brian-avery: interesting. Thanks.

mec/cmd/mec/main.go
dgn: I made it return an error for now

brian-avery (approved):

PR #210, opened 2020-11-24
MAISTRA-1980: Fix private destination rules in root namespace

bison:
This is cherry-picks or partially cherry-picks a few upstream commits to try to fix private destination rules in the root namespace for the maistra-1.1 branch.

This seems to fix the issue for me, but I'm actually not convinced we should merge this. It's a fairly big change, and it wasn't possible to fully bring over the changes to this stuff from 1.6 and above without making the diff huge. So, it's hard to say whether this has other side effects. There's also a workaround, which is to simply run the gateway in a namespace that is not the root config namespace. In any case, I wanted to put this PR up so we can decide one way or the other.

### DIFF
--- a/pilot/pkg/model/push_context.go
+++ b/pilot/pkg/model/push_context.go
@@ -70,15 +70,11 @@ type PushContext struct {
 	privateVirtualServicesByNamespace map[string][]Config
 	publicVirtualServices             []Config
 
-	// destination rules are of three types:
+	// destination rules are of two types:
 	//  namespaceLocalDestRules: all public/private dest rules pertaining to a service defined in a given namespace
 	//  namespaceExportedDestRules: all public dest rules pertaining to a service defined in a namespace
-	//  allExportedDestRules: all (public) dest rules across all namespaces
-	// We need the allExportedDestRules in addition to namespaceExportedDestRules because we select
-	// the dest rule based on the most specific host match, and not just any destination rule
 	namespaceLocalDestRules    map[string]*processedDestRules
 	namespaceExportedDestRules map[string]*processedDestRules
-	allExportedDestRules       *processedDestRules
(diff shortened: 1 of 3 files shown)


pilot/pkg/model/push_context.go
```diff
@@ -606,14 +600,6 @@ func (ps *PushContext) GetAllSidecarScopes() map[string][]*SidecarScope {
 
 // DestinationRule returns a destination rule for a service name in a given domain.
 func (ps *PushContext) DestinationRule(proxy *Proxy, service *Service) *Config {
-	// FIXME: this code should be removed once the EDS issue is fixed
-	if proxy == nil {
```
dgn, line 610: I think we can't remove this check in this older version; you can see some unit tests stack tracing. I suppose the mentioned "EDS issue" isn't fixed on this branch yet :-1:

dgn (commented):
I don't think this would be very problematic to merge, as its impact is pretty well-defined. it shouldn't impact anything but DestinationRule processing.

Then again though, I'm not sure in which support phase we are with 1.1 - it's likely that we're only fixing security and critical issues at this point, which I wouldn't think this is, as there's a quite simple workaround

pilot/pkg/model/push_context.go
bison: Ah, yeah you're right. I couldn't easily bring over the refactor of services that I think make this unnecessary. Hmm.. not actually sure at first glance what the right thing to do here is.

bison:
After reviewing this, and some offline discussions. I'm not sure we can fix this without pulling in more changes. Given there's a workaround and the 1.1 release is going into maintenance mode, I'm going to close this. We can always revisit if it's determined to be more critical.

PR #220, opened 2020-12-15
Add vendor/ directory

dgn:
Also skips over stackdriver tests, as these require a GCP environment


tests/integration/telemetry/stackdriver/stackdriver_filter_test.go
```diff
@@ -136,6 +136,7 @@ func TestStackdriverMonitoring(t *testing.T) {
 
 func TestMain(m *testing.M) {
 	framework.NewSuite(m).
+		Skip("https://github.com/istio/istio/issues/29268").
```
jwendell, line 139: Given this can be forgotten by us, we could create a Maistra Jira issue and link it with the upstream issue, and reference the maistra issue here.

tests/integration/telemetry/stackdriver/vm/main_test.go
```diff
@@ -126,6 +126,7 @@ spec:
 func TestMain(m *testing.M) {
 	framework.
 		NewSuite(m).
+		Skip("https://github.com/istio/istio/issues/29268").
```
jwendell, line 129: ditto

jwendell (approved):

tests/integration/telemetry/stackdriver/stackdriver_filter_test.go
dgn: the linter verifies that this specifies an istio/istio issue, that's why I put that there

jwendell: verifies in what sense? My concern is that this issue is fixed upstream and we never re-enable this test because it went unnoticed. We could at least create a Jira issue so that we can revisit it in the future.

dgn: it just checks that the skipReason contains an istio/istio issue. sure we can track it in a MAISTRA ticket

bison:
Looks good to me, but I agree with @jwendell. If we are creating issues for this can we link them here?

Also, more generally how are we tracking our patches this time around? I would love to be able to review what we brought over or dropped etc without having to go through git logs. Google Doc / JIRA / whatever is fine, but we should probably get it started now.

bison:
In case it's useful, this is my tracking issue from last time: https://issues.redhat.com/browse/MAISTRA-1480

dgn:
> Looks good to me, but I agree with @jwendell. If we are creating issues for this can we link them here?

I created https://issues.redhat.com/browse/MAISTRA-2050 and will update this PR to include both ours and the upstream issue in the skipReason.

> Also, more generally how are we tracking our patches this time around? I would love to be able to review what we brought over or dropped etc without having to go through git logs. Google Doc / JIRA / whatever is fine, but we should probably get it started now.

Good point. I created https://issues.redhat.com/browse/MAISTRA-2051 to track the rebase, I will also create a Google Sheet to give an overview of what parts we cherry-pick, reimplement or drop. I'll link it on the issue. I also linked above issue to this one.

bison (approved):
:100:

PR #221, opened 2020-12-16
MAISTRA-2051 xns-informers for istio types

dgn:


dgn:
This is ready for review now

bison:
This looks good, but should we be worried about it changing the version of `istio/api` and `istio/client-go`? The changes seem minor here, but in general I wonder what the best approach is. Maybe we eventually create a `maistra-2.1` branch of `xns-informers` and make sure the versions there are the same as we expect here?

dgn:
> This looks good, but should we be worried about it changing the version of `istio/api` and `istio/client-go`? The changes seem minor here, but in general I wonder what the best approach is. Maybe we eventually create a `maistra-2.1` branch of `xns-informers` and make sure the versions there are the same as we expect here?

Yeah, those changes are not problematic; it all comes down to istio/client-go going to v1.8.1, which is the version we're rebasing against anyway. I agree we should be moving to a `maistra-2.1` branch of xns-informer though

bison:
Cool. Agreed. Let's just merge this, and come back to that.

bison (approved):

dgn:
Nevermind, you were completely right. I had accidentally pulled in a commit from master of istio/api. The changes were only in comments, so would probably not have caused issues (which is why I had previously assumed them to be updates that happened on `release-1.8` branch where I just happened to pull in a later commit), but we should definitely not pull them in if not needed. Fixed now. Thank you!

PR #225, opened 2021-01-04
MAISTRA-2060: Point to Red Hat Jira for issues

jwendell:


README.md
```diff
@@ -1,6 +1,10 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/istio/istio)](https://goreportcard.com/report/github.com/istio/istio)
 [![GoDoc](https://godoc.org/istio.io/istio?status.svg)](https://godoc.org/istio.io/istio)
 
+# Issues for this repository are disabled
```
brian-avery, line 4: I wonder how many people actually read this and whether it would be better to use an issue template to tell people to go to Red Hat's Jira.

brian-avery (approved):

README.md
jwendell: Can an issue template prevent people from creating issues?

When we disable issues in GH settings, people just won't see the Issues tab, they might wonder what's going on and eventually will land on README.

dgn (approved):

PR #227, opened 2021-01-07
MAISTRA-2051: Add MemberRollController

bison:
Includes the following:

* 75742b0cb4 - Add Makefile for Maistra client generation

This adds a Makefile with tasks for generating the Maistra specific
clients. It is adapted from the one in the Istio client-go repository.

* 5f8562cde4 - Add MemberRollController

Squashed commit, consisting of:

  * MAISTRA-417: Add MemberRollController

  * MAISTRA-450: Switch across to ConfiguredMembers and check existing
    config for changes

  * MAISTRA-1005 Do not expect namespaces to have a stable ordering

The clients have also been regenerated.

* 1b7a89dac6 - Add missing licenses

* fe2f8cda8f - Fix Copyright holder

* 06ea2db574 - MAISTRA-1666: Add `go mod vendor` as part of `make gen`

This is to make sure all PR's will have an updated
vendor directory. The `gencheck` prow job will ensure that.


bison:
I don't know what's up with the unit tests, but I have more commits coming to this PR anyway. I'll look into it when everything's ready.

pilot/pkg/bootstrap/server.go
```diff
@@ -492,6 +493,18 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
...
+
+		memberRollName := args.RegistryOptions.KubeOptions.MemberRollName
+		if memberRollName != "" {
+			err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName)
+			if err != nil {
+				return fmt.Errorf("failed creating member roll: %v", err)
+			}
+		} else {
+			// No MemberRoll configured, set namespaces based on args.
+			namespaces := strings.Split(args.RegistryOptions.KubeOptions.WatchedNamespaces, ",")
```
dgn: this might be problematic, as it's going to end up as empty slice instead of a slice with an empty string (which would mean all namespaces)

bison: Oops. True.

dgn: except of course if we merge my pr: https://github.com/maistra/xns-informer/pull/4

bison: Fixed.

dgn (approved):
the memberroll controller is missing code that'll fire an update right when you register - this is to avoid permission issues. I think that was a later commit, though

PR #228, opened 2021-01-08
MAISTRA-2061 Fix meshnetworking test

dgn:
This is an upstream commit that fixes a potential deadlock, but it also changes the cache sync behaviour. Might solve our issues with the test


dgn:
Does not do the trick I'm afraid

PR #229, opened 2021-01-12
MAISTRA-2051: Add xns-informer for Service APIs

bison:
This enables multi-namespace aware informers for the optional Kubernetes Service APIs.

PR #230, opened 2021-01-14
MAISTRA-2051: Run sidecars with dynamically-determined user IDs

brian-avery:
A lot of this code was refactored in 1.8. This takes Marko's original PR and rewrites it to work with the refactor. 


luksa (approved):

bison:
Do you mind squashing the commits and adding a `MAISTRA-2051:` prefix? It will make going through the git logs during the next rebase a lot easier.

brian-avery:
Sure. I can do that. Was thinking about that in the meeting. Not sure how this will show up in commit history.

bison:
So, it looks like it ended up with my last commit to this branch in the commit message somehow, and still doesn't have the prefix in the summary line. Not sure what happened.

bison:
Yeah, this ended up pulling the the commits from #229. Probably needed a rebase.

PR #236, opened 2021-01-21
[MAISTRA-224][MAISTRA-2010]: Support for authz regex matching on request headers

brian-avery:
MAISTRA-224: Support for authz regex matching on request headers


dgn:
Can we include this commit in here as well? https://github.com/maistra/istio/commit/78d6a2c0f859f4dfc0b3a70bd1ab60f978c49597

The feature isn't complete without it and we missed it for the last release (fixed in 2.0.1) - let's squash these together so it'll never happen again :)

brian-avery:
@dgn done

bison (approved):

PR #242, opened 2021-01-25
MAISTRA-2105 Use -w flag when running iptables commands

luksa:
The -w flag makes iptables wait for the xtables lock. Without this flag,
the command fails if the lock is held by another process.


brian-avery (approved):

dgn (approved):

bison:
This needs the `okay to merge` label if it's ready to go in.

I assume this affects upstream as well? Is there an upstream issue we can reference here?

bison:
I guess it's this one: https://github.com/istio/istio/pull/30394

Looks like it is out of sync with this. We can fix it when that merges upstream though I suppose, or else catch it on the next rebase.

PR #245, opened 2021-01-27
[MAISTRA-1451][MAISTRA-1612] Add tests for pod redirect annotation

brian-avery:
Maistra-1451 added pod redirect annotation. Maistra-1612 removed reverted most of that code. In the mean time, that code greatly changed in Istio. This is a manual merge of the two.


bison:
I don't think this is necessary after the upstream changes in istio/istio PR #26596.

I'm going to close this. Re-open if you disagree.

PR #234, opened 2021-01-19
Do not download envoy-centos image

jwendell:
We do not use it.

Also, change proxy SHA to match maistra proxy.

PR #251, opened 2021-01-29
Make `make docker` work again

jwendell:
- Point the Base URL to ours instead of upstream
- Download envoy-centos binary (Reverts
  https://github.com/maistra/istio/pull/234/commits/bb004bd823640ea15e747e21bbb7c0d706a6f9e3)
- Bump proxy SHA

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -121,7 +121,7 @@ endif
 
 # Envoy binary variables Keep the default URLs up-to-date with the latest push from istio/proxy.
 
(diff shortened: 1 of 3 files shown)


jwendell:
@bison I tested `make docker` locally with this patch and it worked for me.

bison:
Awesome. Thank you! I will try it out in just a bit.

bison (approved):
Fixes `make docker` for me. Though, `make test` is still broken. Looks like our `envoy` isn't compatible with the libc in the build container. We can follow up on that though.

jwendell:
@bison Created https://issues.redhat.com/browse/MAISTRA-2110

PR #254, opened 2021-02-01
MAISTRA-2051: Use MultiNamespaceInformer in galley

bison:
This basically is a rewrite of #241, which updated Galley to use the central Kubernetes client. That turned out to not work because Galley manages stopping and starting individual informers on its own, which doesn't play nicely with shared / cached informers managed by a factory.

This creates individual multi-namepsace informers instead. The diff from upstream is much smaller and all the tests and mocks work as-is.

This also brings in the changes to integrate the MemberRoll controller with Galley.


bison:
The full diff here is kind of messy. The diff between the revert commit and the "Use MultiNamespaceInformer in galley" commit is more reflective of the actual changes to Galley.

dgn (approved):

PR #255, opened 2021-02-02
MAISTRA-2051: Allow istiod to run without Node access

bison:
This is a reimplementation of the following for the Maistra 2.1 /
Istio 1.8 rebase:

 - MAISTRA-1723: Allow disabling NodePort gateway support (#180)
 - Ensure Pilot can run without privileges for reading nodes (#125)

A new istiod flag is added, `disableNodeAccess`, that will disable any
feature relying on access to Node objects.

### DIFF
--- a/pilot/cmd/pilot-discovery/main.go
+++ b/pilot/cmd/pilot-discovery/main.go
@@ -149,6 +149,8 @@ func init() {
 		"The name of the MemberRoll resource")
 	discoveryCmd.PersistentFlags().BoolVar(&serverArgs.RegistryOptions.KubeOptions.EnableCRDScan, "enableCRDScan", true,
 		"Whether to scan CRDs at startup")
+	discoveryCmd.PersistentFlags().BoolVar(&serverArgs.RegistryOptions.KubeOptions.DisableNodeAccess, "disableNodeAccess", false,
+		"Whether to prevent istiod watching Node objects")
 
 	// using address, so it can be configured as localhost:.. (possibly UDS in future)
 	discoveryCmd.PersistentFlags().StringVar(&serverArgs.ServerOptions.HTTPAddr, "httpAddr", ":8080",
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -97,7 +97,8 @@ func (s *Server) initConfigController(args *PilotArgs) error {
 			leaderelection.
 				NewLeaderElection(args.Namespace, args.PodName, leaderelection.IngressController, s.kubeClient.Kube()).
 				AddRunFunction(func(leaderStop <-chan struct{}) {
-					ingressSyncer := ingress.NewStatusSyncer(meshConfig, s.kubeClient)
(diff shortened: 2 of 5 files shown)


bison:
cc: @rcernich @luksa

pilot/cmd/pilot-discovery/main.go
```diff
@@ -149,6 +149,8 @@ func init() {
 		"The name of the MemberRoll resource")
 	discoveryCmd.PersistentFlags().BoolVar(&serverArgs.RegistryOptions.KubeOptions.EnableCRDScan, "enableCRDScan", true,
 		"Whether to scan CRDs at startup")
+	discoveryCmd.PersistentFlags().BoolVar(&serverArgs.RegistryOptions.KubeOptions.DisableNodeAccess, "disableNodeAccess", false,
+		"Whether to prevent istiod watching Node objects")
```
rcernich, line 153: Previously, we've been using `podLocalitySource=pod` for this.  Is there a reason to change it?

bison: Well, this has all changed since the 1.6 release. In that particular case, we don't really need the pod locality thing anymore, as Istio already has an annotation that goes on Pods that indicates locality. So this checks that, and if it doesn't exist, it falls back looking for the Node labels that have been copied to the Pod -- if you've disabled Node access. Otherwise it goes on to look at the Nodes. Disabling the Node informer that would normally be used to check the Node labels for locality there also has the effect of disabling support for NodePort gateway because we can't read the addresses. And potentially other features in the future.

Basically if you disable the Node informer in the controller, you're affecting multiple features that might rely on it. So it seems better to just have a single flag for it.

rcernich: Okay.  We'll need to make changes to the istiod charts to accommodate the new flag.

bison: We may also want to have the operator start preferring the upstream locality label on pods:

https://github.com/istio/istio/blob/1.8.0/pilot/pkg/model/service.go#L139-L146

rcernich: I don't think we're modifying any charts related to those.

rcernich (approved):

pilot/cmd/pilot-discovery/main.go
bison: In case it wasn't clear, I mean right now our operator copies the topology information from Node objects to Pods, right? Istio now has a dedicated label for Pods to read that information. So if we change the operator to start writing that new label to pods, we can eventually drop almost all of this and be more inline with upstream.

rcernich: Ah, understood.  Could you create an issue for 2.1 please?

bison: https://issues.redhat.com/browse/MAISTRA-2115

rcernich: Thanks @bison!

PR #265, opened 2021-02-11
MAISTRA-1801: Support ServiceMeshExtension in istiod

bison:
This is the rebase of the ServiceMeshExtension work for the Maistra 2.1 / Istio 1.8 release.

This primarily comes from 8b1ed09411 - MAISTRA-1801 Support ServiceMeshExtension in istiod (#176), but includes later fixes as well. I've taken great care to split this up into commits that affect as few individual logical components as possible and to squash later fixes in the hope that the next rebase of this will be easier. Each commit message has detail on what it introduces and where the changes came from.

I've chosen to drop the changes that renamed the MemberRoll controller, as they don't materially affect this and complicated the rebase. If anyone feels strongly, I'm happy to re-add them as a separate commit.


brian-avery:
Skipping the rename is fine with me. I'll update the IOR PR once this merges.

brian-avery:
What's the reasoning for the don't squash here? I think that since they're all related, they should be squashed to make them easier to track.

bison:
> What's the reasoning for the don't squash here? I think that since they're all related, they should be squashed to make them easier to track.

I feel precisely the opposite. Tracking them is not the problem as long as the commit messages are clear. The diff-stat for this combined and excluding vendor is: `77 files changed, 4732 insertions(+), 38 deletions(-)`. I think the problem is trying to cherry-pick that in one go several versions of Istio later and having any idea if you're doing the right thing along the way.

This lets you easily cherry-pick just the addition of the clients, then the main changes, then `mec`, etc. and every step of the way should lint cleanly.

dgn (approved):

PR #267, opened 2021-02-11
MAISTRA-2128: Fix flaky TestClientNoCRDs

jwendell:
Introduced by MAISTRA-1895: https://github.com/maistra/istio/pull/260

That test did not exist when we introduced the original change in Istio
1.6. In Istio 1.8 it does, and the default behavior changed with our change.

Thus, we should keep the default behavior that the test expects, i.e.,
having access to all CRD's, aka `options.EnableCRDScan = true`.

Note that `true` is its default value, we explicitly pass the value
`false` in our operator.

### DIFF
--- a/pilot/pkg/config/kube/crdclient/client_test.go
+++ b/pilot/pkg/config/kube/crdclient/client_test.go
@@ -48,7 +48,7 @@ func makeClient(t *testing.T, schemas collection.Schemas) model.ConfigStoreCache
 		}, metav1.CreateOptions{})
 	}
 	stop := make(chan struct{})
-	config, err := New(fake, "", controller.Options{})
+	config, err := New(fake, "", controller.Options{EnableCRDScan: true})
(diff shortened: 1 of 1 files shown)


dgn:
If this caused a test failure, how come it's never come up on prow?

jwendell:
@dgn Look at the prow history for the unit job. It happened a lot since #260 was merged.

dgn:
> @dgn Look at the prow history for the unit job. It happened a lot since #260 was merged.

Oh I see now, this has caused flakes! Nevermind my question then. I had thought that this was something that always caused failures, so I was just surprised how it got merged in the first place :slightly_smiling_face: Thanks for the heads-up

dgn (approved):

bison (approved):
Oops. Good catch. Thanks!

I ninja edited the PR title to include the name of the test before this merged. Hope that's okay.

PR #271, opened 2021-02-16
MAISTRA-2132 Support deployment of multiple plugin versions in Istio CNI

luksa:


brian-avery (approved):

cni/pkg/install-cni/cmd/root.go
```diff
@@ -138,6 +138,7 @@ func constructConfig() (*config.Config, error) {
 
 		UpdateCNIBinaries: viper.GetBool(constants.UpdateCNIBinaries),
 		SkipCNIBinaries:   viper.GetStringSlice(constants.SkipCNIBinaries),
+		CNIBinariesPrefix: viper.GetString(constants.CNIBinariesPrefix),
```
jwendell, line 142: I think you need to register it in `init()`

jwendell (commented):
Worth to add unit tests? And submit upstream after it?

cni/pkg/install-cni/cmd/root.go
luksa: Thanks! Fixed.

jwendell (approved):
We can merge this as is and add unit tests in a separate PR: https://issues.redhat.com/browse/MAISTRA-2135

PR #273, opened 2021-02-17
MAISTRA-2137 Make network namespace setup executable name configurable

luksa:
To support the deployment of multiple CNI plugin versions, the name of the
executable that is invoked to set up the network namespace must be configurable.

### DIFF
--- a/cni/cmd/istio-cni/main.go
+++ b/cni/cmd/istio-cni/main.go
@@ -53,6 +53,7 @@ type Kubernetes struct {
 	NodeName             string   `json:"node_name"`
 	ExcludeNamespaces    []string `json:"exclude_namespaces"`
 	CNIBinDir            string   `json:"cni_bin_dir"`
+	NetnsSetupExecutable string   `json:"netns_setup_executable"`
 }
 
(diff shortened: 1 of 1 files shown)


cni/cmd/istio-cni/main.go
```diff
@@ -53,6 +53,7 @@ type Kubernetes struct {
 	NodeName             string   `json:"node_name"`
 	ExcludeNamespaces    []string `json:"exclude_namespaces"`
 	CNIBinDir            string   `json:"cni_bin_dir"`
+	NetnsSetupExecutable string   `json:"netns_setup_executable"`
```
jwendell, line 56: I wonder if concatenating the cni-bin-prefix in this [line](https://github.com/maistra/istio/blob/maistra-2.1/cni/cmd/istio-cni/iptables.go#L42) fixes the problem without  having to introduce another setting.

luksa: No, because there is no cni-bin-prefix in this context. The prefix is passed to `install-cni`, whereas this here is `istio-cni`, which gets called by the Multus. The only config that it gets is the YAML that comes from the configmap.

jwendell (approved):

PR #275, opened 2021-02-22
MAISTRA-2149: Make IOR robust in multiple replicas

jwendell:
In scenarios where multiple replicas of istiod are running,
only one IOR should be in charge of keeping routes in sync
with Istio Gateways. We achieve this by making sure IOR only
runs in the leader replica.

Also, because leader election is not 100% acurate, meaning
that for a small window of time there might be two instances
being the leader - which could lead to duplicated routes
being created if a new gateway is created in that time frame -
we also change the way the Route name is created: Instead of
having a generateName field, we now explicitly pass a name to
the Route object to be created. Being deterministic, it allows
the Route creation to fail when there's already a Route object
with the same name (created by the other leader in that time frame).

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -144,7 +144,7 @@ func (s *Server) initConfigController(args *PilotArgs) error {
 	if features.EnableIOR {
 		s.addStartFunc(func(stop <-chan struct{}) error {
 			go leaderelection.
-				NewLeaderElection(args.Namespace, args.PodName, leaderelection.StatusController, s.kubeClient).
+				NewLeaderElection(args.Namespace, args.PodName, leaderelection.IORController, s.kubeClient).
 				AddRunFunction(func(stop <-chan struct{}) {
 					ior.Register(s.kubeClient, s.configController, args.Namespace, s.mrc, stop)
 				}).Run(stop)
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -142,7 +142,14 @@ func (s *Server) initConfigController(args *PilotArgs) error {
 	s.environment.IstioConfigStore = model.MakeIstioStore(s.configController)
 
 	if features.EnableIOR {
-		ior.Register(s.kubeClient, s.configController, args.Namespace, s.mrc)
+		s.addStartFunc(func(stop <-chan struct{}) error {
+			go leaderelection.
+				NewLeaderElection(args.Namespace, args.PodName, leaderelection.StatusController, s.kubeClient).
+				AddRunFunction(func(stop <-chan struct{}) {
+					ior.Register(s.kubeClient, s.configController, args.Namespace, s.mrc, stop)
+				}).Run(stop)
+			return nil
+		})
 	}
 
 	// Defer starting the controller until after the service is created.
--- a/pilot/pkg/config/kube/ior/ior.go
+++ b/pilot/pkg/config/kube/ior/ior.go
@@ -29,7 +29,7 @@ import (
 var iorLog = log.RegisterScope("ior", "IOR logging", 0)
 
 // Register configures IOR component to respond to Gateway creations and removals
-func Register(client kubernetes.Interface, store model.ConfigStoreCache, pilotNamespace string, mrc memberroll.Controller) {
+func Register(client kubernetes.Interface, store model.ConfigStoreCache, pilotNamespace string, mrc memberroll.Controller, stop <-chan struct{}) {
 	iorLog.Info("Registering IOR component")
 
 	if !isRouteSupported(client) {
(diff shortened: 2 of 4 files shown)


pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -142,7 +142,14 @@ func (s *Server) initConfigController(args *PilotArgs) error {
 	s.environment.IstioConfigStore = model.MakeIstioStore(s.configController)
 
 	if features.EnableIOR {
-		ior.Register(s.kubeClient, s.configController, args.Namespace, s.mrc)
+		s.addStartFunc(func(stop <-chan struct{}) error {
+			go leaderelection.
+				NewLeaderElection(args.Namespace, args.PodName, leaderelection.StatusController, s.kubeClient).
```
brian-avery: If an Istiod instance is running IOR and another gets elected, does this need to stop?

jwendell: nope, I have to implement that. Thanks for the catch.

jwendell:
@rcernich @dgn @bison @luksa 

You that have more experience in controllers than me,

The problem arises when multiple replicas of istiod are in place.

IOR listens for Gateways changes and react to them. If there is more than one replica, each one will independently react to them and might end up, for instance, creating duplicate routes when a Gateway is created.

This PR "almost" fixes this issue by making use of the leader election. Only one IOR (the leader) is active. The problem is when it is demoted. The stop signal is not closed immediately when a new leader is elected, meaning that, for a small amount of time (a few seconds) both istiod/ior are the leaders. So, if in that time window a gateway is created, both IOR's will still react.

I talked to a few people, it seems that having exactly 1 leader at all times is not guaranteed by the election system. So, we should be more robust in our code to deal with that scenario.

Do you have any suggestions here? Perhaps a whole refactoring of the IOR code? How?

jwendell:
Perhaps a simpler solution would be to have a static name for the route, deterministic, instead of having a dynamic one. Currently we are making use of the `GenerateName` field passing only a prefix and letting OpenShift create the name, similar to pod names that derive from the deployment/replicaset name.

Thus, the second attempt to create the route will fail with a duplicate name.

rcernich:
I think this is an all around good idea.  Not having a consistent name makes the generated routes pretty useless, as you don't know what the name will be, so you can't have a static link for it.

jwendell:
@rcernich I meant the Route object's name, not the hostname.

bison:
> Perhaps a simpler solution would be to have a static name for the route, deterministic, instead of having a dynamic one. Currently we are making use of the GenerateName field passing only a prefix and letting OpenShift create the name, similar to pod names that derive from the deployment/replicaset name.

I think this is getting closer to the correct solution, though we should still be doing leader election of course. Ultimately, we need the control loops to be properly idempotent.

To be honest, I'm not that familiar with how IOR is supposed to work. Is it always one Gateway results in one Route? I think `GenerateName` is intended for one to many relationships based on label selectors, e.g. Deployment > ReplicaSets > Pods. If it's one to one, then I agree we should just name it predictably.

dgn:
+1 for deterministic names. We can use e.g. the name and a hash of the Gateway resource as part of the name.

Update: actually that would result in new routes being created whenever we change a resource.. so maybe just namespace+name of the gateway resource?

bison:
> Update: actually that would result in new routes being created whenever we change a resource.. so maybe just namespace+name of the gateway resource?

Yeah, exactly. I think it's as simple as `${NAMESPACE}-${NAME}` assuming they go in a different namespace than the corresponding Gateway. If they go in the same namespace, then literally just give them the exact same name.

jwendell:
Ok folks I'll work on that. Just a quick note it can't be just "ns+gw name" as a gateway can have more than one host, which means more than one route. Probably append the hostname too. I'll think about it and update this pr. Thanks.

jwendell:
Just pushed the fix. I've updated the commit/PR description to reflect the changes.

pilot/pkg/bootstrap/configcontroller.go
jwendell: done

dgn: all other components that use this seem to use a unique `electionID`, should we do the same?

jwendell: Not unique. `leaderelection.StatusController` is used more than one time for example. It's ok to use it as well.

rcernich:
instead of hash, maybe the uid for the gateway object?

jwendell:
> instead of hash, maybe the uid for the gateway object?

A gateway can have more than 1 hostname. So, 1 Gateway -> N Routes

pilot/pkg/bootstrap/configcontroller.go
dgn: really? I could only find it being used once, in `pilot/pkg/bootstrap/configcontroller.go:399`

dgn: Nevermind I was looking at the maistra-2.1 branch, where they're unique. I'd still prefer us to use our own electionID, even though it's not the case on maistra-2.0 yet.

The reason I think it can be problematic is if the other controller (for whatever reason) stops, it will release the leader lock, leading to us also losing the leader lock. Something very simple like `leaderelection.StatusController + "-ior"` will be sufficient I think

jwendell: Done. Changed the id and tested. Worked fine.

brian-avery (approved):

dgn (approved):

PR #279, opened 2021-02-25
MAISTRA-2150: Prevent deadlock, do not hold edsClusterMutex when load…

knrc:
…ing assignments

### DIFF
--- a/pilot/pkg/proxy/envoy/v2/debug.go
+++ b/pilot/pkg/proxy/envoy/v2/debug.go
@@ -666,12 +666,18 @@ func (s *DiscoveryServer) edsz(w http.ResponseWriter, req *http.Request) {
 		AdsPushAll(s)
 	}
 
-	edsClusterMutex.RLock()
-	defer edsClusterMutex.RUnlock()
+	edsClusterMutex.Lock()
+	// Create a temp map to avoid locking the add/remove
+	cMap := make(map[string]*EdsCluster, len(edsClusters))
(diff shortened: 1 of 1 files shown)


brian-avery (approved):

jwendell (commented):
I'm confused. Here you're touching `edsz` function, but in the Jira you mix calls with `cdsz`. Is that right?

knrc:
> I'm confused. Here you're touching `edsz` function, but in the Jira you mix calls with `cdsz`. Is that right?

In the reproducer?  Yes, it's a simple invocation which demonstrates the deadlock within pilot.

jwendell:
> > I'm confused. Here you're touching `edsz` function, but in the Jira you mix calls with `cdsz`. Is that right?
> 
> In the reproducer? Yes, it's a simple invocation which demonstrates the deadlock within pilot.

How? They use different mutexes.

knrc:
> How? They use different mutexes.

It's a cascade, it goes deeper.  I'll add details in the JIRA.

knrc:
> It's a cascade, it goes deeper. I'll add details in the JIRA.

The explanation is in the JIRA, deadlock on edsClusterMutex cascades to adsClientsMutex

jwendell (approved):
Thanks for the clarification.

knrc:
> Thanks for the clarification.

No problem at all, your question prompted me to add in the details which would be necessary to understand the impact.  I should have done this before.

PR #280, opened 2021-02-25
MAISTRA-2151: ensure locks are released using defer

knrc:
Note the underlying issue which caused MAISTRA-2101 has already been fixed in the 2.0 branch (getResourceVersion not handling a nonce of "") so the reproducer from that issue is not valid.  This cherry picks the same changes with regards to unlocking.

### DIFF
--- a/pilot/pkg/proxy/envoy/v2/debug.go
+++ b/pilot/pkg/proxy/envoy/v2/debug.go
@@ -207,15 +207,19 @@ func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 	_, _ = fmt.Fprintln(w, "{}]")
 }
 
+func (s *DiscoveryServer) marshallEndpoints() ([]byte, error) {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	return json.MarshalIndent(s.EndpointShardsByService, " ", " ")
+}
+
 // Dumps info about the endpoint shards, tracked using the new direct interface.
 // Legacy registry provides are synced to the new data structure as well, during
 // the full push.
 func (s *DiscoveryServer) endpointShardz(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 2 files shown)


jwendell (approved):

pilot/pkg/proxy/envoy/v2/debug.go
```diff
@@ -214,8 +214,8 @@ func (s *DiscoveryServer) endpointShardz(w http.ResponseWriter, req *http.Reques
 	_ = req.ParseForm()
 	w.Header().Add("Content-Type", "application/json")
 	s.mutex.RLock()
+	defer s.mutex.RUnlock()
```
bison: I may be misunderstanding the flow here, but this seems off. The lock is intended to protect access to `s.EndpointShardsByService`, right? Holding it while writing the response to the wire seems bad.

knrc: I completely agree, however that's the current implementation and this is just guarding against the unlock not happening because of a panic during the writing of the response.

bison: Ah, yeah, fair enough. Did this come from upstream or something, or is this our implementation? Haven't seen any mention of upstream issues, or whether this needs to be addressed in the 2.1 branch.

Anyway, would a `getEndpointShardsJSON()` method, similar to how the other bits were split out, be better? I just worry that this may introduce a new, different, possibility for a DOS.

knrc: It's upstream code and nothing changes functionally, all this change does is remove the possibility of a DoS by ensuring the lock is always released regardless of how the function terminates.

knrc: Actually I see what you mean and will update it

bison (approved):

dgn (approved):

PR #283, opened 2021-03-02
MAISTRA-2153: Disable namespace informer if MemberRoll is used

bison:
This causes istiod to skip creating the namespace informer in the
Kubernetes service registry controller if a MemberRoll is specified,
because Maistra cannot read namespace objects.  This will affect the
behavior of multi-network deployments that do not use mesh networks
for configuration, i.e. a namespace cannot be labeled with a default
network now.

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -307,7 +307,8 @@ func NewController(kubeClient kubelib.Client, options Options) *Controller {
 		syncInterval:                options.GetSyncInterval(),
 	}
 
-	if options.SystemNamespace != "" {
+	// Don't start the namespace informer if Maistra's MemberRoll is in use.
+	if options.SystemNamespace != "" && options.MemberRollName == "" {
 		c.nsInformer = informers.NewSharedInformerFactoryWithOptions(c.client, options.ResyncPeriod,
 			informers.WithTweakListOptions(func(listOpts *metav1.ListOptions) {
 				listOpts.FieldSelector = fields.OneTermEqualSelector("metadata.name", options.SystemNamespace).String()


bison:
This partially breaks the multi-network stuff. Does that affect our plans for federation or anything? cc: @dgn

bison:
I guess we could call this handler when the MemberRoll changes. Do we want to do that, or does this just not matter for us?

rcernich:
I wonder if a simple watch on the namespace is enough, as it appears it's only watching the `systemNamespace` anyway (I assume that's what the filter is doing).

dgn:
afaik we can't even watch our own namespace

dgn:
we're not really using the multi-network stuff at this point, and the labelling is also not strictly required for multi-cluster/federation

bison:
> I wonder if a simple watch on the namespace is enough, as it appears it's only watching the systemNamespace anyway (I assume that's what the filter is doing).

Ah, true. So no need to connect this to MemberRoll then, but like Daniel said, I don't think we can watch namespaces at all. So I guess Maistra just can't support this.

dgn (approved):

dgn (changes requested):
looking through the code, this isn't the only place namespaces are watched. there's a namespace controller in `pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go` that is initialized in `pilot/pkg/bootstrap/server.go:1093`

bison:
We handle that in the namespace controller by using the MemberRoll instead.

https://github.com/maistra/istio/blob/maistra-2.1/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go#L120-L122

bison:
There's also a separate issue I'm working on for a list of namespaces that happens elsewhere: [MAISTRA-2152](https://issues.redhat.com/browse/MAISTRA-2152)

dgn (approved):
That's right, sorry I was looking at the wrong branch here

PR #284, opened 2021-03-02
MAISTRA-2051 fixes for maistra install

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -461,6 +461,9 @@ func (c *client) RunAndWait(stop <-chan struct{}) {
 	c.metadataInformer.Start(stop)
 	c.istioInformer.Start(stop)
 	c.serviceapisInformers.Start(stop)
+	if c.memberRoll != nil {
+		c.memberRoll.Start(stop)
+	}
 	if c.fastSync {
 		// WaitForCacheSync will virtually never be synced on the first call, as its called immediately after Start()
 		// This triggers a 100ms delay per call, which is often called 2-3 times in a test, delaying tests.
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -1100,6 +1100,27 @@ func potentialPodName(metadata *metav1.ObjectMeta) string {
 	return ""
 }
 
+// rewriteCniPodSpec will check if values from the sidecar injector Helm
+// values need to be inserted as Pod annotations so the CNI will apply
+// the proper redirection rules.
+func rewriteCniPodSpec(annotations map[string]string, spec *SidecarInjectionSpec) {
+
+	if spec == nil {
+		return
+	}
+	if len(spec.PodRedirectAnnot) == 0 {
+		return
+	}
+	for k := range AnnotationValidation {
+		if spec.PodRedirectAnnot[k] != "" {
+			if annotations[k] == spec.PodRedirectAnnot[k] {
+				continue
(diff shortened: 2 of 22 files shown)


rcernich:
@bison, @dgn FYI.  These are the changes I've had to make so far to get pilot working with maistra.  For whatever reason, it doesn't appear that endpoint/pod watch is working for namespaces outside the system namespace (e.g. istio-system).

pkg/kube/client.go
```diff
@@ -461,6 +461,7 @@ func (c *client) RunAndWait(stop <-chan struct{}) {
 	c.metadataInformer.Start(stop)
 	c.istioInformer.Start(stop)
 	c.serviceapisInformers.Start(stop)
+	c.memberRoll.Start(stop)
```
rcernich: I'm not sure if this is correct, but it seemed the member roll controller needed to be started somewhere.

bison: Oops. Yeah, but we need to check if it is `nil` before starting it.

pkg/kube/inject/inject.go
```diff
@@ -1100,6 +1100,28 @@ func potentialPodName(metadata *metav1.ObjectMeta) string {
 	return ""
 }
 
+
+// rewriteCniPodSpec will check if values from the sidecar injector Helm
+// values need to be inserted as Pod annotations so the CNI will apply
+// the proper redirection rules.
+func rewriteCniPodSpec(annotations map[string]string, spec *SidecarInjectionSpec) {
```
bison, line 1106: Oops. I had purposefully skipped this stuff, cause I think it's fixed upstream, but part of it may have only made it into the 1.9 release.

pkg/kube/inject/webhook.go
```diff
@@ -714,6 +714,10 @@ func injectPod(req InjectionParameters, partialInjection bool) ([]byte, error) {
 		annotations[k] = v
 	}
 
+	if len(spec.PodRedirectAnnot) != 0 {
```
jwendell: nit: this `if` is redundant with the one inside the function

tools/istio-iptables/pkg/cmd/root.go
```diff
@@ -192,6 +194,16 @@ func init() {
 	}
 	viper.SetDefault(constants.ProxyGID, "")
 
+	rootCmd.Flags().Bool(constants.RedirectDNS, false,
```
dgn, line 197: why do we need this redirectDNS option? I think if we want to change the behaviour around DNS interception we should do it in a separate PR, it shouldn't be required to get an install to work?

rcernich: This isn't a change.  The logic for enabling this used to use an environment variable to enable/disable this (`ISTIO_META_DNS_CAPTURE`), but that doesn't work for CNI, so I added a command line option.  Viper should be initializing this option using the value from the env, if specified, when running in the init container.  This is basically synchronizing the existing iptables capabilities used by init container with cni.

rcernich: fwiw, this change is encapsulated in a separate commit.

dgn (approved):

PR #285, opened 2021-03-04
MAISTRA-2152: Get namespaces from MemberRoll in gateway controller

bison:
This avoids listing namespaces in the gateway controller, which
Maistra can't do, by having the controller take an optional
NamespaceSet argument that is registered with the MemberRoll
controller.  If that is present, the list of namespaces is taken from
it rather than from the API server.

Note that the NamespaceSet will only have access to namespace names
and no other metadata, so this may not work correctly for anything
needing things like annotations, labels, etc. from namespaces.

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	xnsinformers "github.com/maistra/xns-informer/pkg/informers"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/keepalive"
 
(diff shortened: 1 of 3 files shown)


bison:
This diff is unfortunate, and seems a bit brittle -- does anything use other metadata, or might something use it in the future? But I'm not sure what else we can do.

bison:
@jwendell maybe you could build this and try to reproduce the issue you reported?

jwendell:
@bison Sorry, I didn't find this issue while *running* istio, I found it while *reading* the code, so I thought of opening a Jira issue for us to keep track of it, and fix it if it's indeed an issue. IIRC @rcernich said this wouldn't be a problem because that code is never reached in Maistra/OSSM. He can confirm.

dgn:
It needs the Gateway objects because it does labelSelector magic. I don't think we can replace it like this.

dgn:
Then again, as has been said before, we probably don't even need this component. It's experimental upstream iirc

bison:
Ah, yeah, it does use the namespace labels for matching at some point. Oh well.

I guess can @rcernich confirm we're not planning to enable this or provide an option to any time soon? Should we just close this?

bison:
Closing this. Don't think it's necessary for now.

PR #292, opened 2021-03-12
MAISTRA-2194 && MAISTRA-2195 Federation Service Discovery v1 (incl. /watch endpoint)

dgn:
This does not yet use CRDs to configure exports, and it doesn't yet offer the option to change service names on import (this depends on the routing part, and also CRDs). Instead, it currently uses a service label (`maistra.io/exportAs`) to determine what the service will be exported as.

The routing is also not final. Instead of mTLS termination at egress/ingress, this was tested using the eastwest-gateway arch that upstream uses for multi-primary deployments.

Now also includes /watch endpoint and client code (in the ServiceRegistry)

If you want to test this, you can do the following:
- check out the `federation` branch of https://github.com/dgn/istio-multicluster
- run `./setup.sh install` - this will setup two kind clusters (you'll need a 1.8.x istioctl)
- mark a service in one of the clusters as exported, e.g. by doing `kubectl --context kind-cluster2 -n sample edit svc helloworld` and adding a `maistra.io/exportAs: helloworld.sample.svc.cluster2.local` label
- check the logs of the other istiod instance to see that it has received the exported service
- try connecting from cluster1's sleep pod to cluster2's now-exported helloworld service: `kubectl --context kind-cluster1 exec -it -n sample <SLEEP_POD_NAME> -- curl -v helloworld.sample.svc.cluster2.local:5000/hello`

Next steps:
- use mTLS termination routing (Rob has been working on this part)
- add CRDs (we should also investigate MCS serviceexport/import here)
- add integration tests

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff"
+
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/serviceregistry"
 	"istio.io/istio/pkg/config/constants"
@@ -37,6 +39,8 @@ var _ model.ServiceDiscovery = &Controller{}
 var _ model.Controller = &Controller{}
 var _ serviceregistry.Instance = &Controller{}
 
+var logger = log.RegisterScope("federation-registry", "federation-registry", 0)
+
 // Controller aggregates data across different registries and monitors for changes
 type Controller struct {
 	meshHolder        mesh.Holder
@@ -44,6 +48,7 @@ type Controller struct {
 	networkName       string
 	clusterID         string
 	resyncPeriod      time.Duration
+	backoffPolicy     *backoff.ExponentialBackOff
 
 	xdsUpdater model.XDSUpdater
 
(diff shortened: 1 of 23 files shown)


rcernich:
Hey @dgn, this is still not allowing pilot to come up (watch fails, no sync, no hasSynced=true).

dgn:
> Hey @dgn, this is still not allowing pilot to come up (watch fails, no sync, no hasSynced=true).

Sorry, looks like I pushed an empty changeset there. It's up now.

I also fixed a bug that caused service deletion to not be applied immediately (if you watched my demo you know what I mean :wink: ) and added an exponential backoff for watch connection retries.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+}
+
+func (c *Controller) handleEvent(e *federation.WatchEvent) {
+	if e.Service == nil {
+		checksum := c.resync()
+		if checksum != e.Checksum {
+			// this shouldn't happen
+			logger.Error("checksum mismatch after resync")
+		}
```
rcernich, line 313: I think there needs to be a `return` here.

rcernich:
I think we should merge this, as I think it provides a good starting point.

pilot/pkg/serviceregistry/federation/controller.go
dgn: thanks, good catch

dgn: I added a return below this line

rcernich: yep

rcernich (approved):

jwendell (approved):

PR #295, opened 2021-03-16
MAISTRA-2164: Refactor IOR internals

jwendell:
Instead of doing lots of API calls on every event - this
does not scale well with lots of namespaces - keep the state
in memory, by doing an initial synchronization on start up and
updating it when receiving events.

The initial synchronization is more complex, as we have to deal with
asynchronous events (e.g., we have to wait for the Gateway store to
be warmed up). Once it's initialized, handling events as they arrive
becomes trivial.

Tests that make sure we do not make more calls to the API server than
the necessary were added, to avoid regressions.

### DIFF
--- a/pilot/pkg/config/kube/ior/fake.go
+++ b/pilot/pkg/config/kube/ior/fake.go
@@ -26,6 +26,8 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	"istio.io/istio/pkg/servicemesh/controller"
 )
 
 // FakeRouter implements routev1.RouteInterface
@@ -84,6 +86,8 @@ func (rc *FakeRouterClient) Routes(namespace string) routev1.RouteInterface {
 	if _, ok := rc.routesByNamespace[namespace]; !ok {
 		rc.routesByNamespace[namespace] = NewFakeRouter()
 	}
+
+	countCallsIncrement("routes")
 	return rc.routesByNamespace[namespace]
 }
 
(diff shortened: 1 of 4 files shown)


dgn (approved):
lgtm in general, but two questions:
1. shouldn't the List() calls be unproblematic as they will hit the client-go cache?
2. how do we make sure we're not keeping stale state in memory, ie what happens when we miss an event? should we add a periodical resync?

jwendell:
> lgtm in general, but two questions:
> 
>     1. shouldn't the List() calls be unproblematic as they will hit the client-go cache?

Is it always the case it hits the cache? Also, I'm more worried about a regression here. We should make sure to not make more api calls than necessary in future commits. Sync() function may call List(), but other calls are suspect.

> 
>     2. how do we make sure we're not keeping stale state in memory, ie what happens when we miss an event? should we add a periodical resync?

I didn't know it was possible to miss an event. Does this happen often? Is it worth to refactor it to do a periodic sync()? Note that sync() relies on the Gateway store/cache as the source of truth. If the store is broken (eg, by not sending an event), syncing again will not have any effect.

dgn:
> > 2. how do we make sure we're not keeping stale state in memory, ie what happens when we miss an event? should we add a periodical resync?
> > ```
> 
> I didn't know it was possible to miss an event. Does this happen often? Is it worth to refactor it to do a periodic sync()? Note that sync() relies on the Gateway store/cache as the source of truth. If the store is broken (eg, by not sending an event), syncing again will not have any effect.

Looking into it, I think `client-go` makes sure these sorts of things can't happen. So probably nothing to worry about.

brian-avery (approved):

PR #301, opened 2021-03-25
MAISTRA-2223: Include istiod support for affinity and tolerations

knrc:


manifests/charts/istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -201,6 +201,16 @@ spec:
           - name: extracacerts
             mountPath: /cacerts
           {{- end }}
+      affinity:
+      {{- include "nodeaffinity" . | indent 6 }}
+      {{- include "podAntiAffinity" . | indent 6 }}
+{{- if .Values.pilot.tolerations }}
+      tolerations:
+{{ toYaml .Values.pilot.tolerations | indent 6 }}
+{{- else if .Values.global.defaultTolerations }}
```
shamsher31, line 210: The default value for `global.defaultTolerations` is not defined in `values.yaml`. Actually, the global values are not present in the `2.0` charts so not sure if this condition is applicable here.

knrc: It's still applicable, the default is defined in the global.yaml file at a higher level

shamsher31: Nevermind, found that `defaultTolerations` is already in `global.yaml` file. 
https://github.com/maistra/istio/blob/034f7b7c61ab7027808fccb249857196c0ecf069/manifests/charts/global.yaml#L310

And I think values from `global.yaml` file are now combined into `values.yaml` in`2.1` charts.

shamsher31: Ref: Upstream PR for combining global values into `values.yaml`. https://github.com/istio/istio/issues/25184

rcernich (approved):

jwendell (approved):

PR #302, opened 2021-03-31
MAISTRA-2231: Seed MemberRoll listeners with system namespace

bison:

In Maistra 2.1 this actually prevents the full install from completing
in some cases, e.g. the namespace controller may never create the
ConfigMap containing the CA certificate, which prevents the gateway
pods from starting.

namespace if no SMMR resource is found.

### DIFF
--- a/pkg/servicemesh/controller/controller.go
+++ b/pkg/servicemesh/controller/controller.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
@@ -38,6 +39,7 @@ type serviceMeshMemberRollController struct {
 	started        bool
 	lock           sync.Mutex
 	cacheWarmed    bool
+	cacheLock      sync.RWMutex
 }
 
 type MemberRollListener interface {
@@ -74,23 +76,24 @@ func (smmrc *serviceMeshMemberRollController) Start(stopCh <-chan struct{}) {
 	smmrc.lock.Lock()
 	defer smmrc.lock.Unlock()
 
-	if !smmrc.started {
-		go smmrc.informer.Run(stopCh)
-		smmrc.started = true
-
-		smmrLog.Debug("Controller started, waiting for cache to warm up")
-		go func() {
-			for {
-				if cache.WaitForNamedCacheSync("SMMR", stopCh, smmrc.informer.HasSynced) {
-					smmrLog.Debug("Cache warmed up. From now on will send the initial update to listeners")
-					smmrc.cacheWarmed = true
-					return
-				}
-				smmrLog.Debug("Cache not synced, trying again")
-				time.Sleep(time.Second)
-			}
-		}()
+	if smmrc.started {
+		return
 	}
+
+	go smmrc.informer.Run(stopCh)
+	smmrc.started = true
+
+	smmrLog.Debug("Controller started, waiting for cache to warm up")
+	go func() {
+		if cache.WaitForNamedCacheSync("smmr", stopCh, smmrc.informer.HasSynced) {
+			smmrLog.Debug("Cache synced. Will update listeners.")
+
+			smmrc.cacheLock.Lock()
+			defer smmrc.cacheLock.Unlock()
+
+			smmrc.cacheWarmed = true
+		}
+	}()
 }
 
 func (smmrc *serviceMeshMemberRollController) Register(listener MemberRollListener, name string) {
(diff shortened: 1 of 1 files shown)


pkg/servicemesh/controller/controller.go
```diff
@@ -121,19 +107,38 @@ func (smmrc *serviceMeshMemberRollController) newServiceMeshMemberRollListener(l
...
+				// No need to send initial update.  The informer will do it.
+				return
+			}
 
-		var members []string
-		for _, item := range smmrc.informer.GetIndexer().List() {
-			smmr := item.(*v1.ServiceMeshMemberRoll)
-			members = smmr.Status.ConfiguredMembers
+			smmrLog.Infof("Seeding listener %q with system namespace.", name)
+			handler.updateNamespaces("seed", smmrc.memberRollName, smmrc.getNamespaces(nil))
```
dgn: you should be able to call this with `members: nil`; the `updateNamespaces` func calls `getNamespaces` anyway

bison: Thanks. Fixed.

pkg/servicemesh/controller/controller.go
```diff
@@ -121,19 +107,38 @@ func (smmrc *serviceMeshMemberRollController) newServiceMeshMemberRollListener(l
...
+	// were registered that included only the Istio system namespace.  That
+	// caused problems with some controllers, e.g. IOR, because they expect the
+	// callback to be called with the full authoritative set of namespaces and
+	// may remove resources for namespaces not in the list.
+	//
+	// This instead waits for the informer's cache to sync, then sends an
+	// initial update only if the expected SMMR is not found in the cache.
+	// Unfortunately, there is a race here between sending this initial update
+	// and a user creating an SMMR resource.
+	go func() {
```
dgn, line 136: this will now create one goroutine per listener that'll poll the `HasSynced()` func... not sure if that's ideal. It does avoid having to use a mutex, but I'm not sure about the trade-off there. before the cache is synced, we're now creating ~20 goroutines all hammering the SMMR informer. it's probably not a big deal, but do you think it's better than a single goroutine + mutex?

bison: I see what you mean, but I don't expect it to be a problem in practice. It's not that many threads, and they are polling with a reasonable interval, and the `HasSynced()` method on informers is not doing any heavy lifting -- it's not hitting the API or anything. This seemed fine and easier to read to me since it's all in one place.

dgn: yeah I agree, it's not a ton of goroutines and they really don't do much

dgn (approved):
lgtm, thanks! For a longer-term solution, I think your idea of having an Add/Remove interface instead of SetNamespaces would be worth exploring, if people agree I'll file a ticket and we can look into it for 2.1

pkg/servicemesh/controller/controller.go
```diff
@@ -121,19 +107,38 @@ func (smmrc *serviceMeshMemberRollController) newServiceMeshMemberRollListener(l
 		name:              name,
 	}
 
-	if smmrc.cacheWarmed {
-		smmrLog.Debugf("Listener for %q created. Ready to send an initial update", name)
+	// WARNING: This is a bit of a hack.  It is mostly intended as a temporary
+	// workaround to enable 2.1 testing to begin.
```
jwendell: Please, file an issue in 2.1 bucket and link it here

pkg/servicemesh/controller/controller.go
```diff
@@ -121,19 +107,38 @@ func (smmrc *serviceMeshMemberRollController) newServiceMeshMemberRollListener(l
...
+			if exists {
+				// No need to send initial update.  The informer will do it.
+				return
+			}
 
-		var members []string
-		for _, item := range smmrc.informer.GetIndexer().List() {
-			smmr := item.(*v1.ServiceMeshMemberRoll)
-			members = smmr.Status.ConfiguredMembers
+			smmrLog.Infof("Seeding listener %q with system namespace.", name)
```
dgn: nit: the `updateNamespaces` func is already logging (at `debug` level), do we need this log line?

pkg/servicemesh/controller/controller.go
jwendell: I prefer the single goroutine+lock here, but since it's a temporary fix, that's ok.

pkg/servicemesh/controller/controller.go
bison: Don't have a strong opinion, but this is only logged in the case where the SMMR doesn't exist. Seemed kind of nice to know without having to up the log level. Happy to make it debug or remove it though.

jwendell:
likely the timeout on that test needs to be increased... prow is too slow... it takes 2 seconds on my laptop. 1 minute on prow is not enough.

bison:
Going to pause this until we make a decision on #308.

bison:
We merged #308 into the 2.0 branch, so I've updated this to match it. I think this is ready now.

pkg/servicemesh/controller/controller.go
bison: This uses a flag and a RW mutex now, though it's not a single go routine. We still have to poll in the listeners.

dgn (approved):

pkg/servicemesh/controller/controller.go
bison: Updated to match what is in 2.0 -- this was already for 2.1 -- don't think there's anything to do here.

jwendell (approved):

PR #305, opened 2021-04-05
MAISTRA-2246: Rewrite multi-namespace listwatcher to address performa…

knrc:
…nce issues

This PR implements a two layer cache for multiple namespaces, with the lower level caches handling individual namespaces, resync etc and being aggregated into a single upper layer cache for consumption by other parts of the code.

The lower level consumes events from an informer and adapts them as Watch events, these watch events are then consumed by the upper level cache to create an aggregated set of watch events for consumers.

Note this does not suffer from the thrashing seen in the original implementation, it also removes the need to track the resource versions for each namespace as they are now handled independently of each other.

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -446,8 +446,8 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
 			return fmt.Errorf("failed creating kube config: %v", err)
 		}
 		s.kubeClient, err = kubelib.CreateClientset(args.Config.KubeConfig, "", func(config *rest.Config) {
-			config.QPS = 20
-			config.Burst = 40
+			config.QPS = 100
+			config.Burst = 200
 		})
 		if err != nil {
 			return fmt.Errorf("failed creating kube client: %v", err)
--- a/pkg/listwatch/listinformer.go
+++ b/pkg/listwatch/listinformer.go
@@ -1,0 +1,160 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Provides multiple namespace listerWatcher. This implementation is Largely from
+// https://github.com/coreos/prometheus-operator/pkg/listwatch/listwatch.go
+
+package listwatch
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+)
+
+/*
+ * This type adapts the events from the informer into events which would come
+ * from a Watch, in this way we can use this as a lower level cache handling
+ * individual namespaces to feed a higher level cache aggregating across all
+ * the namespaces.
+ *
+ * This type will also support draining in that once the namespace has been removed
+ * from the aggregated cache it will issue DELETED events for each entry within this
+ * namespace so that the upper level cache is consistent.
+ */
+type listerInformer struct {
+	namespace  string
+	informer   cache.SharedInformer
+	lllw       *listLenListerWatcher
+	events     func(*watch.Event)
+	stopped    chan struct{}
+	hasStopped bool
+	draining   bool
+	drained    func(string)
+	lock       sync.RWMutex
+}
+
+func newListerInformer(namespace string, f func(string) cache.ListerWatcher, exampleObject runtime.Object,
+	resyncPeriod time.Duration, events func(*watch.Event), drained func(string)) *listerInformer {
+
+	lllw := newListLenListerWatcher(f(namespace))
+	informer := cache.NewSharedInformer(lllw, exampleObject, resyncPeriod)
+
+	li := &listerInformer{
+		namespace: namespace,
+		informer:  informer,
+		lllw:      lllw,
+		events:    events,
+		stopped:   make(chan struct{}),
+		drained:   drained,
+	}
+	informer.AddEventHandler(li)
+	go informer.Run(li.stopped)
+	return li
+}
+
+// The lister informer is considered synced if the underlying
+// informer has synced *and* we have passed sufficient ADDED events
+// to the higher level cache to match the underlying List count.
+func (li *listerInformer) hasSynced() bool {
+	return li.informer.HasSynced() && li.lllw.hasReachedListCount()
+}
+
+func (li *listerInformer) isDraining() bool {
+	li.lock.RLock()
+	defer li.lock.RUnlock()
+	return li.draining
+}
+
+func (li *listerInformer) isStopped() bool {
+	li.lock.RLock()
+	defer li.lock.RUnlock()
+	return li.hasStopped
+}
+
+func (li *listerInformer) drain() {
+	// If we are already draining then there is nothing to do
+	shouldDrain := func() bool {
+		li.lock.Lock()
+		defer li.lock.Unlock()
+		if li.draining {
+			return false
+		}
+		li.draining = true
+		// We are draining.  Stop the underlying informer so it
+		// cannot interfere with our Delete events
+		li.stopInformer()
+		return true
+	}()
+	if shouldDrain {
+		go func() {
+			// Issue Delete events for each entry remaining in the store
+			store := li.informer.GetStore()
+			resourcesToDrain := store.List()
+			for _, resource := range resourcesToDrain {
+				li.OnDelete(resource)
+			}
+			li.lock.Lock()
+			defer li.lock.Unlock()
+			if !li.hasStopped {
+				li.hasStopped = true
+				li.drained(li.namespace)
+			}
+		}()
+	}
+}
+
+func (li *listerInformer) stopInformer() {
+	close(li.stopped)
+}
+
+func (li *listerInformer) stop() {
+	li.lock.Lock()
+	defer li.lock.Unlock()
+	li.hasStopped = true
+	li.stopInformer()
+}
+
+func (li *listerInformer) newWatchEvent(eventType watch.EventType, obj interface{}) *watch.Event {
+	return &watch.Event{
+		eventType, obj.(runtime.Object),
+	}
+}
+
+func (li *listerInformer) sendEvent(event *watch.Event) {
+	li.lock.RLock()
+	defer li.lock.RUnlock()
+	if !li.hasStopped {
+		li.events(event)
+	}
+}
+
+// Adapt an OnAdd event into a watch ADDED event
+func (li *listerInformer) OnAdd(obj interface{}) {
+	li.sendEvent(li.newWatchEvent(watch.Added, obj))
+	li.lllw.incAddCount()
+}
+
+// Adapt an OnUpdate event into a watch MODIFIED event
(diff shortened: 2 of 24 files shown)


pkg/listwatch/listwatch.go
```diff
@@ -40,216 +43,253 @@ type NamespaceListerWatcher interface {
...
+			liMap[namespace] = dli
+			delete(mlw.drainingMap, namespace)
+		} else {
+			// Do we need to watch the namespace?
+			liMap[namespace] = mlw.addNamespace(namespace)
+		}
+	}
+
+	// Start to drain all removed namespaces
+	for _, namespace := range mlw.namespaces {
```
dgn: isn't this draining all namespaces, no matter if they have just been removed? what I mean is: it says "drain all removed namespaces" but it actually drains all namespaces

pkg/listwatch/listinformer.go
```diff
@@ -0,0 +1,196 @@
...
+		li.draining = true
+		li.stopInformer()
+		return true
+	}()
+	if shouldDrain {
+		go func() {
+			store := li.informer.GetStore()
+			resourcesToDrain := store.List()
+			for _, resource := range resourcesToDrain {
+				li.OnDelete(resource)
```
dgn, line 109: I like this part, sending explicit delete events for things that fall out of the watch. might solve a few issues we encountered. we should look at implementing the same in xns-informers

pkg/listwatch/listwatch_test.go
```diff
@@ -458,30 +521,37 @@ func TestMultiWatchDoesNotDeadlock(t *testing.T) {
...
+	}, &v1.Pod{}, 0)
 
 	if _, err := m.List(metav1.ListOptions{}); err != nil {
 		t.Errorf("Error while listing: %v", err)
 	}
 
-	if _, err := m.Watch(metav1.ListOptions{}); err == nil {
-		t.Error("Expected error while watching, got nil")
+	<-errorRaised
+	<-errorRaised
```
dgn, line 542: why wait for two errors to be raised?

pkg/servicemesh/controller/extension/controller.go
```diff
@@ -68,14 +68,15 @@ func NewControllerFromConfigFile(kubeConfig string, namespaces []string, mrc mem
...
+		&v1alpha1.ServiceMeshExtension{}, resync,
 	)
 	if mrc != nil {
 		mrc.Register(mlw, "extensions-controller")
 	}
 	store := make(map[string]*v1alpha1.ServiceMeshExtension)
 	informer := cache.NewSharedIndexInformer(
 		mlw, &v1alpha1.ServiceMeshExtension{},
-		resync, cache.Indexers{})
+		0, cache.Indexers{})
```
dgn, line 79: why does the resync period have to be `0` everywhere now? I see we're passing it into the `mlw` because it has its own informers now, just curios. does it break the `mlw`?

pkg/listwatch/listwatch_test.go
```diff
@@ -30,69 +32,61 @@ import (
...
-				t.Errorf("newMultiWatch should not panic when number of resource versions matches ListerWatchers; got: %v", r)
-			}
-		}()
-		// Create a multiWatch from 2 ListerWatchers and pass 2 resource versions.
-		_, _ = setupMultiWatch(t, []string{"1", "2"}, map[string]string{"1": "resource1", "2": "resource2"})
-	}()
+	return ws, &wsLock, mlw
 }
 
 func TestMultiWatchResultChan(t *testing.T) {
```
dgn, line 75: can we add a test case for the issue that this is meant to solve? and if not, an acceptance test that runs your reproducer?

pilot/pkg/bootstrap/server.go
```diff
@@ -446,8 +446,8 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
 			return fmt.Errorf("failed creating kube config: %v", err)
 		}
 		s.kubeClient, err = kubelib.CreateClientset(args.Config.KubeConfig, "", func(config *rest.Config) {
-			config.QPS = 20
-			config.Burst = 40
```
dgn, line 450: why do these need to be increased?

pkg/listwatch/listinformer.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+type listLenListerWatcher struct {
+	listChan      chan int
+	listerWatcher cache.ListerWatcher
+}
+
+func (lllw *listLenListerWatcher) List(options metav1.ListOptions) (runtime.Object, error) {
+	lllw.listChan <- -1
+	<-lllw.listChan
```
dgn: ?? why are pushing to this channel then immediately read from it and discard the result

pkg/listwatch/listinformer.go
```diff
@@ -0,0 +1,196 @@
...
+	result, err := lllw.listerWatcher.List(options)
+	if err != nil {
+		return result, err
+	}
+	lllw.listChan <- meta.LenList(result)
```
dgn: we're pushing the length of the list to the channel, but in line 95 we'll push `0` to the channel because we just pushed -1?

dgn (commented):
first pass... this is going to take some time. I would have appreciated a few comments, this is quite complex

bison:
> first pass... this is going to take some time. I would have appreciated a few comments, this is quite complex

Huge +1 to what @dgn said here. I'm just starting to look at this, but I'd just like to, in general, beg that we don't introduce these big changes to core parts of the codebase with *zero* commentary on the design in the commit message or PR description, and even not a single comment in the `listinformer.go` file.

pkg/listwatch/listwatch.go
knrc: Good catch, that should be working through the remaining entries in the map.  I'll update the tests to catch this and then fix.

pkg/listwatch/listwatch_test.go
knrc: It waits for two errors to be raised so we can ensure that the watch is being retried, I can add a comment about it.

pkg/servicemesh/controller/extension/controller.go
knrc: The resync is handled by the lower level caches, there's no need to do it again in the higher cache

pkg/listwatch/listwatch_test.go
knrc: I'm not sure what you mean, this was a test which was no longer applicable so was removed

pilot/pkg/bootstrap/server.go
knrc: They needed to be increased to allow IOR to handle leader election, as these changes were causing more requests to be created we were seeing errors from the underlying client code applying throttling.  We still have some issues with API server communication and throttling which are being investigated.

pkg/listwatch/listinformer.go
knrc: This allows us to sync between this function and a goroutine which emulates the sync behaviour through the events.

pkg/listwatch/listinformer.go
knrc: The values which matter are from the List to the goroutine
-1 => list is starting so reset the expected counts to emulate sync
<= 0 sync point after goroutine has finished initialising expected counts
n => (0 or above) list size to configure the expected count for sync

knrc:
> > first pass... this is going to take some time. I would have appreciated a few comments, this is quite complex
> 
> Huge +1 to what @dgn said here. I'm just starting to look at this, but I'd just like to, in general, beg that we don't introduce these big changes to core parts of the codebase with _zero_ commentary on the design in the commit message or PR description, and even not a single comment in the `listinformer.go` file.

I can add comments, although in listinformer it's pretty straight forward other than the sync emulation.

bison:
> I can add comments, although in listinformer it's pretty straight forward other than the sync emulation.

Thanks, that sounds like a good start. Can I ask specifically that you edit the PR description and/or the commit message and describe the intended design in plain English, so that we, and others in the future, can evaluate the changes in that context? I think any open source project would reject a change of this size with no PR description, and I don't know why we should be any different.

pkg/listwatch/listinformer.go
knrc: I'm going to rework this bit to make it easier to follow.

knrc:
@dgn I fixed that issue and reworked those channels so they are no longer used, it has simplified the design.  I've also added comments into the code to explain some of the non-obvious areas, let me know what you think.

knrc:
@bison this is the two layer cache approach that I originally asked for nearly a year ago, individual caches for each namespace isolating the aggregated (consumable) cache from the network issues and reconnects.  I can add something in the description.

pkg/listwatch/listwatch_test.go
dgn: no I mean we should add a test that verifies that this PR solves the performance issue

knrc: We already have that being done in QE, Yuanlin has a reproducer matching the original issue.  This PR tidies up the behaviour of the caching and handles the thrashing with exceptions being propagated up to the aggregated cache, I can add a test for that specifically.  We may still be throttled by the API server, which is separate from this and being investigated.

knrc: I've added a test for the error condition which caused problems with the cache, it validates that only one namespace is impacted by the changes and doesn't force it on others

dgn (approved):
LGTM. xns-informer also solves the noisy neighbour problem that is fixed here, so I would have preferred a back-port of the xns-informer implementation over a rewrite of the MLW, but I think it's going to work. xns-informer is also complex, so that's not my concern - rather that we now have two very different code bases that each are hard to debug if complicated issues arise. I'll hope that we won't have to touch this again for 1.1 and 2.0 (crossing lots of fingers here).

Diving back into client-go actually made me realize two things we should verify for xns-informer (not related to this PR):
- what is our resync behaviour wrt to top-level and lower-level informers?
- are we sending delete events for resources when namespaces are removed from the multinamespaceinformer?

knrc:
> LGTM. xns-informer also solves the noisy neighbour problem that is fixed here, so I would have preferred a back-port of the xns-informer implementation over a rewrite of the MLW, but I think it's going to work. xns-informer is also complex, so that's not my concern - rather that we now have two very different code bases that each are hard to debug if complicated issues arise. I'll hope that we won't have to touch this again for 1.1 and 2.0 (crossing lots of fingers here).

That would also have been my preference however @bison said it was too risky to include those changes within the 2.0. and 1.1 streams.  For now we will have to live with having two implementations.

bison:
I guess I'm pretty much with @dgn here. If I'm honest, I think this is a pretty complex solution to the problem that won't be fun to maintain. I would also have preferred to adapt xns-informers here, and while that definitely has its complicated bits, I think it's simpler over all personally. Nevertheless, this looks like it should work and hopefully help with namespace churn.

> That would also have been my preference however @bison said it was too risky to include those changes within the 2.0. and 1.1 streams. For now we will have to live with having two implementations.

I think I meant too risky compared to leaving things as-is and without first having tried it out in the 2.1 release -- not in comparison to completely rewriting MLW. This is clearly just as risky. I really wish we had all just sat down and had a thorough dedicated discussion around this.

bison (approved):

knrc:
> I guess I'm pretty much with @dgn here. If I'm honest, I think this is a pretty complex solution to the problem that won't be fun to maintain. I would also have preferred to adapt xns-informers here, and while that definitely has its complicated bits, I think it's simpler over all personally. Nevertheless, this looks like it should work and hopefully help with namespace churn.

I asked about backporting the xns-informers for 2.0 and you said you were against it as it was too risky, that's why we have gone down this route.  We can certainly look to backport it for 2.0/1.1 after this release.

knrc:
> I think I meant too risky compared to leaving things as-is and without first having tried it out in the 2.1 release -- not in comparison to completely rewriting MLW. This is clearly just as risky. I really wish we had all just sat down and had a thorough dedicated discussion around this.

Any changes here are certainly a risk but the status quo is not sustainable, something needed to be done and this approach was discussed a long time back.  We should certainly evaluate backporting the xns-informer codebase to 2.0/1.1, it may take time but we now have that.

PR #307, opened 2021-04-07
MAISTRA-681: Rewrite multi-namespace listwatcher to address performan…

knrc:
…ce issues

This PR implements a two layer cache for multiple namespaces, with the lower level caches handling individual namespaces, resync etc and being aggregated into a single upper layer cache for consumption by other parts of the code.

The lower level consumes events from an informer and adapts them as Watch events, these watch events are then consumed by the upper level cache to create an aggregated set of watch events for consumers.

Note this does not suffer from the thrashing seen in the original implementation, it also removes the need to track the resource versions for each namespace as they are now handled independently of each other.

### DIFF
--- a/pkg/listwatch/listinformer.go
+++ b/pkg/listwatch/listinformer.go
@@ -1,0 +1,160 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Provides multiple namespace listerWatcher. This implementation is Largely from
+// https://github.com/coreos/prometheus-operator/pkg/listwatch/listwatch.go
+
+package listwatch
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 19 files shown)


pkg/listwatch/listinformer.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (li *listerInformer) hasSynced() bool {
+	return li.informer.HasSynced() && li.hasReachedListCount()
+}
+
+func (li *listerInformer) isDraining() bool {
+	li.lock.RLock()
+	defer li.lock.RUnlock()
+	defer li.lock.RUnlock()
```
rcernich: Do we need both?

pkg/listwatch/listinformer.go
```diff
@@ -0,0 +1,198 @@
...
+	return li.draining
+}
+
+func (li *listerInformer) isStopped() bool {
+	li.lock.RLock()
+	defer li.lock.RUnlock()
+	defer li.lock.RUnlock()
```
rcernich: Again, both?

pkg/listwatch/listinformer.go
knrc: good catch, I'll double check the rest for this slip.

knrc: Fixed

pkg/listwatch/listinformer.go
knrc: Fixed

dgn (approved):
lgtm, for comments see #305

bison (approved):
Ditto. Haven't compared line-for-line, but trusting this is the same as #305 as far as I can tell -- that's where the discussion is. LGTM.

knrc:
> Ditto. Haven't compared line-for-line, but trusting this is the same as #305 as far as I can tell -- that's where the discussion is. LGTM.

+1, there is one minor difference due to the refactoring of the MemberRollController (different import and name) but otherwise identical

PR #308, opened 2021-04-08
MAISTRA-2234: Seed MemberRoll listeners with system namespace

bison:

namespace if no SMMR resource is found.  Unfortunately, it's possible
for this to race a user creating an SMMR, in which case the state may
be inconsistent until the next update or full sync.

### DIFF
--- a/pkg/servicemesh/controller/memberroll/controller.go
+++ b/pkg/servicemesh/controller/memberroll/controller.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
@@ -39,6 +40,7 @@ type serviceMeshMemberRollController struct {
 	started        bool
 	lock           sync.Mutex
 	cacheWarmed    bool
+	cacheLock      sync.RWMutex
 }
 
 type Listener interface {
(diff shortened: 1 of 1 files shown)


bison:
I am not 100% convinced we should do this for the 2.0 branch. The issue on 2.1 is that changes in istiod mean this bug prevents things from starting at all in some cases. In my testing on the latest 2.0 builds, that is not the case, regardless of whether you've created the initial SMMR or not.

I do think this has the potential to introduce a race between listeners seeing the cache has synced and looking for the SMMR, and one being created. I think it's possible, but maybe unlikely, that we then send add events out-of-order with this initial "seed" event. If that's the case, is it worth introducing that if we haven't seen this cause any actual issues on this branch?

knrc (changes requested):
The race can be handled in a similar way to how @jwendell handled things in https://github.com/maistra/istio/pull/312, could you please make similar changes to this code?

dgn (commented):
I think just adding a mutex around the `cacheWarmed` variable will be sufficient here. IMHO we should not fix a problem nobody has (ie 'no SMMR present causes strange behaviour') if it adds complexity - if a customer runs into an issue here we can still tackle it then

knrc:
> I think just adding a mutex around the `cacheWarmed` variable will be sufficient here. IMHO we should not fix a problem nobody has (ie 'no SMMR present causes strange behaviour') if it adds complexity - if a customer runs into an issue here we can still tackle it then

We know the problem exists, and while we may not currently be sure it has any impact it is much better to fix it now and not have it impact any customers.

bison:
Updated to use a channel in a similar way to #312 to be sure we don't process events from the informer until we've sent the initial update when necessary.

knrc (approved):

dgn:
> We know the problem exists, and while we may not currently be sure it has any impact it is much better to fix it now and not have it impact any customers.

Alright, that's fine with me, too

dgn (approved):

jwendell (approved):

PR #309, opened 2021-04-08
WIP: MAISTRA-2167 add CRD and controller for federating meshes

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


rcernich:
@bison or @dgn, can either of you help me figure out why the integration tests are failing?  It looks like it's because pilot isn't starting, but I don't know where to look for logs.

bison:
I don't have access to Prow. I think @dgn can probably hook you up with that. I'm guessing you've run them locally and they pass there?

Could you delete the boilerplate in the PR description and write something there? Even just a single line.

dgn:
I'm afraid I don't have access to pod logs-- for runs that happen on prow, I don't think we store them. Do the tests pass locally for you? Istio should store logs in the work dir

rcernich:
I use podman, and I always have problems running locally.  These are failing the integration tests.  I suspect there's an issue at startup, so I'll try a local install with the unadulterated charts.

pkg/kube/controller/controller.go
```diff
@@ -0,0 +1,144 @@
+// Copyright Red Hat, Inc.
```
dgn, line 1: accidental commit?

dgn: nevermind I see now this is being used somewhere

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -383,10 +432,13 @@ func (c *Controller) handleEvent(e *federation.WatchEvent) {
...
 func (c *Controller) watch(eventCh chan *federation.WatchEvent, stopCh <-chan struct{}) error {
-	req, err := http.NewRequest("GET", c.discoveryEndpoint+"/watch", nil)
+	url := c.dicsoveryURL + "/watch"
+	req, err := http.NewRequest(http.MethodGet, url, nil)
 	if err != nil {
-		return err
+		logger.Errorf("Failed to create request: '%s': %s", url, err)
+		return nil
 	}
+	req.Header.Add("discovery-address", c.networkAddress)
```
dgn, line 441: why send the discovery address?

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -58,29 +64,32 @@ type Controller struct {
...
 	return &Controller{
-		meshHolder:        opt.MeshHolder,
-		discoveryEndpoint: opt.DiscoveryEndpoint,
-		networkName:       opt.NetworkName,
-		clusterID:         opt.ClusterID,
-		xdsUpdater:        opt.XDSUpdater,
-		resyncPeriod:      opt.ResyncPeriod,
-		backoffPolicy:     backoffPolicy,
+		meshHolder:     opt.MeshHolder,
+		dicsoveryURL:   fmt.Sprintf("%s://%s:%d", federation.DiscoveryScheme, opt.EgressService, federation.DiscoveryPort),
```
dgn: typo

pkg/servicemesh/federation/controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+var _ model.ConfigStoreCache = (*Controller)(nil)
+
+// NewController creates a new Aggregate controller
+func NewController(opt Options) *Controller {
+	opt.validate()
+
+	cs := clientsetservicemeshv1alpha1.NewForConfigOrDie(opt.Client.RESTConfig())
+
+	var informer cache.SharedIndexInformer
+	if opt.Client.GetMemberRoll() == nil {
```
dgn: I think this should be `!= nil`, right?

pilot/pkg/serviceregistry/federation/controller.go
rcernich: I did this to support routing, in case the gateway is being used to route traffic to multiple external meshes.  The source of the traffic is istiod, which is not running a proxy and is just sending a raw (non-istio) http request.  Internally, the request is sent directly to the egress gateway with this header.  The header is used in a VirtualService to route to a specific discovery server in another mesh.

pkg/servicemesh/federation/controller/controller.go
rcernich: That could be why the integration tests are failing.  Great catch.

rcernich: i don't know if this is even necessary, as I'm currently only watching istio-system.

rcernich:
Replaced by #323

PR #310, opened 2021-04-08
Increase the timeout for a test

jwendell:
It's currently flaky, depending on the load of the prow node.
Time is not much important here, but the quantity of API calls is.

### DIFF
--- a/pilot/pkg/config/kube/ior/ior_test.go
+++ b/pilot/pkg/config/kube/ior/ior_test.go
@@ -316,7 +316,7 @@ func TestPerf(t *testing.T) {
 	mrc.setNamespaces(generateNamespaces(qty)...)
 
 	// It takes ~ 2s on my laptop, it's slower on prow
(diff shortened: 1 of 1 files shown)


pilot/pkg/config/kube/ior/ior_test.go
```diff
@@ -316,7 +316,7 @@ func TestPerf(t *testing.T) {
 	mrc.setNamespaces(generateNamespaces(qty)...)
 
 	// It takes ~ 2s on my laptop, it's slower on prow
-	_, ignore := getRoutes(t, routerClient, "istio-system", qty, time.Minute)
+	_, ignore := getRoutes(t, routerClient, "istio-system", qty, 3*time.Minute)
```
brian-avery, line 319: Should the other uses of getRoutes be updated as well?

rcernich (approved):

rcernich:
guess it didn't fix it :(

jwendell:
This is a legit failure. I'll work on a proper fix: https://issues.redhat.com/browse/MAISTRA-2255

PR #311, opened 2021-04-09
MAISTRA-2255: Make sure IOR respond to all events

jwendell:
Even those that arrive while IOR is performing its initial
sync. Previously those events were ignored.

### DIFF
--- a/pilot/pkg/config/kube/ior/route.go
+++ b/pilot/pkg/config/kube/ior/route.go
@@ -57,16 +57,15 @@ type syncRoutes struct {
 
 // route manages the integration between Istio Gateways and OpenShift Routes
 type route struct {
-	pilotNamespace  string
-	routerClient    routev1.RouteV1Interface
-	kubeClient      kubernetes.Interface
-	store           model.ConfigStoreCache
-	gatewaysMap     map[string]*syncRoutes
-	gatewaysLock    sync.Mutex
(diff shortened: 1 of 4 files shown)


brian-avery (approved):

pilot/pkg/config/kube/ior/route.go
```diff
@@ -328,24 +317,16 @@ func (r *route) SetNamespaces(namespaces ...string) {
...
-			IORLog.Debug("Waiting for the Gateway store cache to sync before performing our initial sync")
-			cache.WaitForNamedCacheSync("Gateways", r.stop, r.store.HasSynced)
-			IORLog.Debug("Gateway store cache synced. Performing our initial sync now")
-
-			if err := r.initialSync(); err != nil {
-				IORLog.Errora(err)
-			}
+		// But only after gateway store cache is synced
+		IORLog.Debug("Waiting for the Gateway store cache to sync before performing our initial sync")
+		cache.WaitForNamedCacheSync("Gateways", r.stop, r.store.HasSynced)
```
bison: Shouldn't we be checking the return value here? If it returns false, it means the stop channel was closed and we probably don't want to proceed with the sync.

jwendell: @bison Done.

knrc (approved):

PR #313, opened 2021-04-11
MAISTRA-2269: Increase hardcoded QPS/Burst to a more reasonable limit

knrc:
Increase QPS/Burst to a more reasonable limit

### DIFF
--- a/pilot/pkg/bootstrap/server.go
+++ b/pilot/pkg/bootstrap/server.go
@@ -446,8 +446,8 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
 			return fmt.Errorf("failed creating kube config: %v", err)
 		}
 		s.kubeClient, err = kubelib.CreateClientset(args.Config.KubeConfig, "", func(config *rest.Config) {
-			config.QPS = 100
-			config.Burst = 200
+			config.QPS = float32(features.PilotAPIServerQPS)
+			config.Burst = features.PilotAPIServerBurst
 		})
 		if err != nil {
 			return fmt.Errorf("failed creating kube client: %v", err)
--- a/pilot/pkg/config/kube/crd/controller/client.go
+++ b/pilot/pkg/config/kube/crd/controller/client.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"istio.io/istio/pilot/pkg/features"
+
 	"github.com/hashicorp/go-multierror"
 
 	"istio.io/api/label"
@@ -202,6 +204,9 @@ func NewClient(config string, context string, schemas collection.Schemas, domain
 	if err != nil {
 		return nil, err
 	}
+
+	cfg.QPS = float32(features.PilotAPIServerQPS)
+	cfg.Burst = features.PilotAPIServerBurst
 
 	return NewForConfig(cfg, schemas, domainSuffix, configLedger, revision)
 }
--- a/pilot/pkg/features/pilot.go
+++ b/pilot/pkg/features/pilot.go
@@ -351,4 +351,16 @@ var (
 	PilotEnableLoopBlockers = env.RegisterBoolVar("PILOT_ENABLE_LOOP_BLOCKER", true,
 		"If enabled, Envoy will be configured to prevent traffic directly to the inbound/outbound "+
 			"ports (15001/15006). This prevents traffic loops. This option will be removed, and considered always enabled, in 1.9.").Get()
+
+	PilotAPIServerQPS = env.RegisterFloatVar(
+		"PILOT_API_SERVER_QPS",
+		500,
+		"The default pilot API server QPS",
+	).Get()
+
+	PilotAPIServerBurst = env.RegisterIntVar(
+		"PILOT_API_SERVER_BURST",
+		1000,
+		"The default pilot API server Burst",
+	).Get()
 )


luksa:
How were these numbers determined? They seem somewhat high. In OpenShift, the default QPS/burst for the kubelet is 50/100, so you get 2000/4000 only in 40-node clusters. Wouldn't something like 250/500 be a more reasonable default?

knrc:
> How were these numbers determined? They seem somewhat high. In OpenShift, the default QPS/burst for the kubelet is 50/100, so you get 2000/4000 only in 40-node clusters. Wouldn't something like 250/500 be a more reasonable default?

I tried with 1000/2000 and was seeing throttling in the order of seconds, increasing this to 2000/4000 brought the throttling down to microseconds/nanoseconds.  This was for a 500 namespace test.

As per the JIRA this is just a temporary change until we get a more comprehensive solution in a later release

jwendell:
Does this have some side effect? In other words, why OpenShift values are so low if they can be increased without worries?

knrc:
> Does this have some side effect? In other words, why OpenShift values are so low if they can be increased without worries?

It will put more load on the API server for a short period of time, the tradeoff for customers who need this is a more responsive service mesh and available applications.  With the original defaults we were seeing throttling delaying requests for 4 minutes and longer.

Normal clients do not need to perform the type of intensive API server access we are doing within istiod/pilot.

luksa:
> Does this have some side effect? In other words, why OpenShift values are so low if they can be increased without worries?

The default values that OpenShift uses for the kubelet suffice when running a limited number of pods on each node. Users are encouraged to increase these values if required, but this does cause a higher load on the API server. 

I'd say these settings depend highly on the cluster, so it's hard to know what the appropriate values are. I'm concerned that too high values could affect the other components running in the cluster. But I don't really know what would be considered a high value. If Pilot needs values this high for normal operation, then we may just need to use them and require users to beef up their API servers. 

My gut feeling is that 2000/4000 is too high, but I can't back it up. I'd feel much better with lower values and instructing users who run 100+ namespaces to increase them in the config.

knrc:
> My gut feeling is that 2000/4000 is too high, but I can't back it up. I'd feel much better with lower values and instructing users who run 100+ namespaces to increase them in the config.

What values would you suggest?  I can rerun the test with those to see the impact.  Our current issue is that we have no way of configuring these in any config, I've opened another issue to deal with this but in the meantime we need to set the defaults to a more reasonable value.

luksa:
> What values would you suggest? I can rerun the test with those to see the impact. Our current issue is that we have no way of configuring these in any config, I've opened another issue to deal with this but in the meantime we need to set the defaults to a more reasonable value.

I'd feel okay with 250/500, which is the equivalent of 5 kubelets/nodes. Or maybe keep the QPS relatively low and just increase the Burst to ensure that Pilot can do what it needs without being blocked. IIRC you mentioned we need 2000/4000 for _500 namespaces_.  That doesn't sound like a typical scenario, so if we reduce the QPS to several hundred, we would still support around 100 namespaces which should be more than enough for most users (?).

knrc:
> I'd feel okay with 250/500, which is the equivalent of 5 kubelets/nodes. Or maybe keep the QPS relatively low and just increase the Burst to ensure that Pilot can do what it needs without being blocked.

okay, I can try that and see how it looks although with 1000/2000 it was still showing significant throttling.

> IIRC you mentioned we need 2000/4000 for _500 namespaces_. That doesn't sound like a typical scenario, so if we reduce the QPS to several hundred, we would still support around 100 namespaces which should be more than enough for most users (?).

Not usual, however it does happen and we have customers who are deploying meshes with around 300 namespaces.

jwendell:
> Not usual, however it does happen and we have customers who are deploying meshes with around 300 namespaces.

This is awesome. I'm guessing upstream doesn't have customers doing this, therefore they never had encountered this issue. This is worth to raise upstream and backport this PR and discuss there as well, so that we don't have to carry this on with us forever.

knrc:
> okay, I can try that and see how it looks although with 1000/2000 it was still showing significant throttling.

Not so good with those figures, we have 6825 requests (out of 13382 requests) which are being throttled by > 1s with everything ramping up until we hit 10s throttles per requests.

brian-avery:
Google customers are running some pretty large meshes. Not sure how many namespaces they have though.

knrc:
After discussing with @luksa and others we settled on defaults of 500 for QPS and 1000 for Burst but with the addition of environment variables should this need to be overridden

luksa (approved):

jwendell (commented):
Why not put them in the already existing `pilot/pkg/features/pilot.go` file?

knrc:
> Why not put them in the already existing `pilot/pkg/features/pilot.go` file?

Good idea, will do that

knrc:
> > Why not put them in the already existing `pilot/pkg/features/pilot.go` file?
> 
> Good idea, will do that

should be done

jwendell (approved):

dgn:
> > Not usual, however it does happen and we have customers who are deploying meshes with around 300 namespaces.
> 
> This is awesome. I'm guessing upstream doesn't have customers doing this, therefore they never had encountered this issue. This is worth to raise upstream and backport this PR and discuss there as well, so that we don't have to carry this on with us forever.

No, they do have very large customers, but they don't use multi-tenancy so they don't have the same problem. We have to send 1 request per namespace per resource; they watch globally so they only send 1 request per resource

PR #316, opened 2021-04-13
MAISTRA-1992 and others

jwendell:
This covers a bunch of things:

- MAISTRA-1992: Adds support to `pullPolicy` and `pullSecrets` fields.
  Although `pullSecrets` is not used directly, but it's used by image
  stream: Upon pulling an image, it will pick all docker secrets in the
  namespace automatically. In order to use this feature, MEC now creates
  the image stream in the same namespace as the `ServiceMeshExtension`
  object, not in the control plane as before.

- MAISTRA-2248: If the `ServiceMeshExtension` ready status is false,
  MEC now always tries to pull the image again. This is to avoid the
  case when image stream is stuck with a bad reference to an image.
  Example:
    User creates a `ServiceMeshExtension` pointing to a protected
    registry. Image stream fails to pull it due to lack of credentials.
    User realizes this and creates a secret (or make the repo public).
    Currently MEC will not attempt to pull it again. With this change,
    MEC will try to pull it again in the next sync event.

- MAISTRA-2252: Avoid process all `UPDATE` events. Only process them if
  `generation` changes.

- Dedicated logging for each component


mec/cmd/mec/main.go
```diff
@@ -47,56 +47,69 @@ var (
 	registryURL    string
 	resyncPeriod   string
 	namespace      string
+
+	mainlog        = log.RegisterScope("main", "Main function", 0)
```
dgn, line 51: nit: it's a matter of taste, but I prefer just using `logger` everywhere instead of `<scope>log`

mec/pkg/pullstrategy/ossm/strategy.go
```diff
@@ -72,9 +71,6 @@ func (p *ossmPullStrategy) createImageStreamImport(imageStreamName string, image
 						Kind: "DockerImage",
 						Name: image.String(),
 					},
-					To: &corev1.LocalObjectReference{
```
dgn, line 75: why did this have to be removed?

mec/pkg/model/interfaces.go
```diff
@@ -14,9 +14,13 @@
...
 package model
 
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
 type ImagePullStrategy interface {
 	// PullImage will go and Pull and image from a remote registry
-	PullImage(image *ImageRef) (Image, error)
+	PullImage(image *ImageRef, namespace string, pullPolicy corev1.PullPolicy, pullSecrets []corev1.LocalObjectReference) (Image, error)
```
dgn, line 23: why move the handling of the `pullPolicy` into the `ImagePullStrategy`? shouldn't the worker make the decision based on status of the resource + return value of `strategy.GetImage()`?

mec/cmd/mec/main.go
jwendell: We could have that. There are some clashes however. For instance, in the package server there are two loggers, http and worker.

mec/pkg/pullstrategy/ossm/strategy.go
jwendell: This field is meant to host the tag, not the image name. If omitted, it defaults to the tag, which is what we want, and is less confusing.

mec/pkg/model/interfaces.go
jwendell: GetImage() only works if the image is in the format registry/repo@sha256... which is not the case in most cases. In fact, when the user uses that format, pullPolicy doesn't matter.

mec/pkg/pullstrategy/ossm/strategy.go
dgn: gotcha, thanks

mec/pkg/model/interfaces.go
dgn: you're right, I forgot that the pullPolicy is relevant to the `ImageStream`

pkg/servicemesh/apis/servicemesh/v1alpha1/servicemeshextension_types.go
```diff
@@ -15,16 +15,19 @@
 package v1alpha1
```
dgn: we should probably bump the version as well? Let's use v1

mec/cmd/mec/main.go
```diff
@@ -47,56 +47,69 @@ var (
...
 				mrc,
 				resyncPeriod,
 			)
 			if err != nil {
-				log.Errorf("Failed to create Extension Controller: %v", err)
+				return fmt.Errorf("failed to create Extension Controller: %v", err)
 			}
 
-			p, err := ossm.NewOSSMPullStrategy(config, namespace)
+			p, err := ossm.NewOSSMPullStrategy(config /*, namespace*/)
```
dgn: lets' remove the comment

mec/cmd/mec/main.go
```diff
@@ -47,56 +47,69 @@ var (
...
 			if err != nil {
-				log.Errorf("Failed to create OSSMPullStrategy: %v", err)
+				return fmt.Errorf("failed to create OSSMPullStrategy: %v", err)
 			}
-			w, err := server.NewWorker(config, p, baseURL, serveDirectory)
+
+			w, err := server.NewWorker(config, p, baseURL, serveDirectory, nil)
```
dgn, line 108: why set the `errorChan` to `nil` here? doesn't that mean we'll never log any errors?

pkg/servicemesh/apis/servicemesh/v1alpha1/servicemeshextension_types.go
```diff
@@ -15,16 +15,19 @@
...
-	Phase            *FilterPhase     `json:"phase"`
-	Priority         *int             `json:"priority,omitempty"`
-	Config           string           `json:"config,omitempty"`
+	Image            string                        `json:"image,omitempty"`
+	ImagePullPolicy  corev1.PullPolicy             `json:"imagePullPolicy,omitempty"`
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+	WorkloadSelector WorkloadSelector              `json:"workloadSelector,omitempty"`
+	Phase            *FilterPhase                  `json:"phase"`
+	Priority         *int                          `json:"priority,omitempty"`
+	Config           string                        `json:"config,omitempty"`
```
dgn: this is a `struct{}` in the new API -- it'll have to be converted to json by istiod

dgn (commented):
lgtm so far, a few smaller comments.

what's still missing to close the ticket:
- bump API version to v1
- copy changes over to maistra/istio-operator (we don't have maistra/api yet)
- implement new config format in istiod parsing

mec/cmd/mec/main.go
jwendell: nope. We log the error on line 254 of worker.go. The channel is used primarily for unit tests so that they can inspect the errors.

dgn: I see, yeah I get it now

pkg/servicemesh/apis/servicemesh/v1alpha1/servicemeshextension_types.go
jwendell: In your API PR this is a `google.protobuf.Struct` in .proto (which maps to a proto struct in go). What do you have in mind for Maistra? Keep it string? map[string]interface{}? It's currently passed straight to envoy here: https://github.com/maistra/istio/blob/ea314b69a97fde3102b758ff5dea592cf34c5713/pkg/servicemesh/extension/apply_extension.go#L169

pkg/servicemesh/apis/servicemesh/v1/servicemeshextension_types.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ServiceMeshExtensionList contains a list of ServiceMeshExtension
+type ServiceMeshExtensionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ServiceMeshExtension `json:"items"`
+}
+
+type ServiceMeshExtensionConfig struct {
+	Data map[string]interface{} `json:"-"`
```
dgn, line 96: why not keep this inline? I guess it has to do with some code-gen weirdness?

pkg/servicemesh/extension/apply_extension.go
```diff
@@ -148,46 +155,59 @@ func ApplyListenerListPatches(
...
-											"seconds": {Kind: &structpb.Value_NumberValue{NumberValue: 30}},
+			TypedConfig: util.MessageToAny(&udpa.TypedStruct{
+				TypeUrl: "type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm",
+				Value: &structpb.Struct{
+					Fields: map[string]*structpb.Value{
+						"config": {Kind: &structpb.Value_StructValue{StructValue: &structpb.Struct{Fields: map[string]*structpb.Value{
+							"name":   {Kind: &structpb.Value_StringValue{StringValue: extension.Name}},
+							"rootId": {Kind: &structpb.Value_StringValue{StringValue: extension.Name + "_root"}},
+							"configuration": {Kind: &structpb.Value_StructValue{StructValue: &structpb.Struct{Fields: map[string]*structpb.Value{
+								"@type": {Kind: &structpb.Value_StringValue{StringValue: "type.googleapis.com/google.protobuf.Struct"}},
```
dgn, line 187: does this actually work? I thought we'd have to pass a `string`

mec/cmd/mec/main.go
dgn: bump :)

dgn (approved):
~a few small things, but~ lgtm

pkg/servicemesh/apis/servicemesh/v1/servicemeshextension_types.go
jwendell: yes, you are right. The tooling complains.

It's similar to this: https://github.com/maistra/istio-operator/blob/maistra-2.1/pkg/apis/maistra/v1/helmvalues.go#L12-L17

pkg/servicemesh/extension/apply_extension.go
jwendell: Works like a charm :)

pkg/servicemesh/apis/servicemesh/v1/servicemeshextension_types.go
dgn: alright, thank you!

pkg/servicemesh/extension/apply_extension.go
dgn: awesome, so we can even hand off the marshaling to proto

manifests/charts/base/crds/maistra.io_servicemeshmemberrolls.yaml
```diff
@@ -0,0 +1,69 @@
+
```
rcernich, line 1: We'll have to clean this up when we move stuff to maistra/api

rcernich (commented):
The api generation stuff looks good to me.

jwendell:
> The api generation stuff looks good to me.

feel free to approve it so it gets merged

rcernich (approved):

PR #318, opened 2021-04-14
MAISTRA-2275: Workaround client-go issue 91073

knrc:
work around client-go issue


dgn:
Interesting. I think in the past we set `resourceVersion = 0` which instead of a full list from the API server would fetch the latest from the cache. Could it be that this is also a valid workaround? Or did we just not notice this issue previously?

jwendell (approved):

dgn (approved):

brian-avery (approved):

PR #319, opened 2021-04-14
MAISTRA-2280: Delay SMMR Start until caches have synced

knrc:
Handle SMMR processing to avoid races with initialisation.  Note this PR supercedes MAISTRA-2197 and MAISTRA-2234

### DIFF
--- a/pkg/servicemesh/controller/memberroll/controller.go
+++ b/pkg/servicemesh/controller/memberroll/controller.go
@@ -19,7 +19,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
@@ -37,10 +36,9 @@ type serviceMeshMemberRollController struct {
 	informer       cache.SharedIndexInformer
 	namespace      string
 	memberRollName string
+	seedNamespaces []string
 	started        bool
-	lock           sync.Mutex
-	cacheWarmed    bool
-	cacheLock      sync.RWMutex
+	lock           sync.RWMutex
 }
 
 type Listener interface {
@@ -50,6 +48,23 @@ type Listener interface {
 type Controller interface {
 	Register(listener Listener, name string)
 	Start(stop chan struct{})
+}
+
+func getServiceMeshMemberRoll(obj interface{}) *v1.ServiceMeshMemberRoll {
+	serviceMeshMemberRoll, ok := obj.(*v1.ServiceMeshMemberRoll)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			smmrLog.Errorf("Couldn't get object from tombstone %#v", obj)
+			return nil
+		}
+		serviceMeshMemberRoll, ok = tombstone.Obj.(*v1.ServiceMeshMemberRoll)
+		if !ok {
+			smmrLog.Errorf("Tombstone contained object that is not a service mesh member roll %#v", obj)
+			return nil
+		}
+	}
+	return serviceMeshMemberRoll
 }
 
 func NewControllerFromConfigFile(kubeConfig string, namespace string, memberRollName string, resync time.Duration) (Controller, error) {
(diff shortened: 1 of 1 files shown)


luksa (approved):

bison:
Does this fix some non-obvious bug in the previous implementation? There is zero "why" info here or JIRA.

knrc:
> Does this fix some non-obvious bug in the previous implementation? There is zero "why" info here or JIRA.

See JIRA https://issues.redhat.com/browse/MAISTRA-2280 for the problem, there is a race happening which means the informers can list at cluster scope instead of with the appropriate namespaces

pkg/servicemesh/controller/memberroll/controller.go
```diff
@@ -78,31 +93,37 @@ func newMemberRollSharedInformer(restClient rest.Interface, namespace string, re
 }
 
 func (smmrc *serviceMeshMemberRollController) Start(stop chan struct{}) {
-	smmrc.lock.Lock()
-	defer smmrc.lock.Unlock()
+	if func() bool {
```
jwendell: Why an anonymous function instead of regular workflow?

pkg/servicemesh/controller/memberroll/controller.go
```diff
@@ -121,56 +142,53 @@ func (smmrc *serviceMeshMemberRollController) getNamespaces(namespaces []string)
...
+func (smmrc *serviceMeshMemberRollController) OnUpdate(oldObj, newObj interface{}) {
+	if serviceMeshMemberRoll := getServiceMeshMemberRoll(newObj); serviceMeshMemberRoll != nil {
+		smmrc.setSeedNamespaces(serviceMeshMemberRoll.Status.ConfiguredMembers)
+	}
+}
 
-	return handler
+func (smmrc *serviceMeshMemberRollController) OnDelete(obj interface{}) {
+	if serviceMeshMemberRoll := getServiceMeshMemberRoll(obj); serviceMeshMemberRoll != nil {
+		smmrc.setSeedNamespaces(serviceMeshMemberRoll.Status.ConfiguredMembers)
```
jwendell: Should it be set to nil regardless?

jwendell (commented):
The original issue was:
- IOR registers and gets an initial ADD event with only istio-system
- After a while, it would get additional ADD's from the actual members of the SMMR

If I understand it correctly, you fixed this by holding Start() until the cache is warmed.
Is it OK to hold Start()? From the point of view of who calls it?

pkg/servicemesh/controller/memberroll/controller.go
```diff
@@ -78,31 +93,37 @@ func newMemberRollSharedInformer(restClient rest.Interface, namespace string, re
...
-	smmrc.started = true
-
 	smmrLog.Debug("Controller started, waiting for cache to warm up")
-	go func() {
-		if cache.WaitForNamedCacheSync("smmr", stop, smmrc.informer.HasSynced) {
-			smmrLog.Debug("Cache synced. Will update listeners.")
-
-			smmrc.cacheLock.Lock()
-			defer smmrc.cacheLock.Unlock()
+	cache.WaitForNamedCacheSync("smzmr", stop, smmrc.informer.HasSynced)
```
jwendell: nit: typo

bison:
> There is now a race with some informers between the new changes in the SMMR controller and the start of the informers, in some circumstances some of the informers can be started with a single namespace of "" (in one case deliberately and in other cases possible through configuration) and updates to the namespace can now arrive after the original list has occurred.

Can you point out an example? Don't we register the MultiListWatchers with the MemberRoll controller? In what case are they getting an empty string?

knrc:
> If I understand it correctly, you fixed this by holding Start() until the cache is warmed.

That's correct, it now waits until cache has synced with the control plane namespace and uses the initial values as the 

> Is it OK to hold Start()? From the point of view of who calls it?

With the 1.1 and 2.0 codebases this is safe to do, I have not looked at the 2.1 codebase so not sure about that.

pkg/servicemesh/controller/memberroll/controller.go
knrc: It's scoping the lock and ensuring it is released before the next section.  I was thinking about moving this into a separate function, it would likely improve the readability.

pkg/servicemesh/controller/memberroll/controller.go
knrc: oops, good catch

pkg/servicemesh/controller/memberroll/controller.go
knrc: If there are no configured namespaces then the update will see only the control plane namespace which would be appropriate, as other namespaces are created/reconciled they will receive further updates.

knrc: Ah sorry, I follow what you are saying and it should be nil

bison:
> > There is now a race with some informers between the new changes in the SMMR controller and the start of the informers, in some circumstances some of the informers can be started with a single namespace of "" (in one case deliberately and in other cases possible through configuration) and updates to the namespace can now arrive after the original list has occurred.
> 
> Can you point out an example? Don't we register the MultiListWatchers with the MemberRoll controller? In what case are they getting an empty string?

Is this because the MultiNamespaceListWatcher takes a list of namespaces in its initializer before being registered with the MemberRoll controller? If so, does this only fix the race if the MemberRoll controller is always started before the informers using the MLW?

bison:
This kind of seems like a known property of the system that was just not handled in mixer. For instance, in istiod, if a MemberRoll is specified, we ensure that `WatchedNamespaces` defaults to the system namespace, and that is what is passed to the ListWatchers. So if the informers are started before the MemberRoll controller, they only watch that.

Is this a problem in many places other than mixer? If not, why not just have mixer check if the namespace set is empty and if there's a MemberRoll -- if those are both true, set the namespace set to the system namespace like we do in istiod. Otherwise, we're rewriting half of the MemberRoll controller and the problem still exists if you don't start separate components in the right order. Let me know if I'm misunderstanding the problem.

knrc:
> This kind of seems like a known property of the system that was just not handled in mixer.

It's a new property introduced by the recent changes, since the existing behaviour always initialised the namespaces albeit to istio-system (or equivalent initial configs) before the original List took place.

> For instance, in istiod, if a MemberRoll is specified, we ensure that `WatchedNamespaces` defaults to the system namespace, and that is what is passed to the ListWatchers. So if the informers are started before the MemberRoll controller, they only watch that.

Now yes, but in the original code it was always sending the update of the namespace but in the new code this happens async hence the race.

> Is this a problem in many places other than mixer?

Yes, although potentially with different outcomes

> If not, why not just have mixer check if the namespace set is empty and if there's a MemberRoll -- if those are both true, set the namespace set to the system namespace like we do in istiod. Otherwise, we're rewriting half of the MemberRoll controller and the problem still exists if you don't start separate components in the right order. Let me know if I'm misunderstanding the problem.

The race exists regardless of whether it is empty or not, this is what I mentioned earlier in the team call.  In the team call I described the difference in expectations between the multi namespace watcher implementation and the new SMMR controller, since it is seeded by the config (sometimes hardcoded to "" otherwise as config) then we still end up in the situation where the informer could be initialised with the multi namespace watcher's seed namespace (istio-system, "" or otherwise) and then updated with the complete list from the SMMR at a later date.  In the 2.0.2 code it was always initialised from the SMMR before it could list.

The alternative I mentioned in the call was to set the initial namespaces for the multi namespace watcher to an empty array, however this deviates further from the 2.0.2 behaviour.  This change simplifies some of the new changes and sticks with the previous behaviour of always sending some update before the informers can list.

pkg/servicemesh/controller/memberroll/controller.go
knrc: done

pkg/servicemesh/controller/memberroll/controller.go
knrc: done

pkg/servicemesh/controller/memberroll/controller.go
knrc: I tidied this up and made it easier to read

jwendell (approved):

PR #321, opened 2021-04-15
MAISTRA-2286: Handle QPS/Burst for galley, mixer and citadel

knrc:
Updated QPS/Burst for other control plane components


pkg/features/api_client_config.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2020 Red Hat, Inc.
```
jwendell, line 1: Not 2020 :)
Istio policy is not to put the year. Anyway, as this is just a move between files, I think it should retain the original value.

jwendell (approved):

luksa (approved):

PR #323, opened 2021-04-20
MAISTRA-2293 add CRD and controller for federating meshes 

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


rcernich:
I'm not sure what's up with the unit test failures.  It's a different test every time.  Is it an issue with the execution environment?

dgn:
They do seem to be quite flaky. I'll take a look

pkg/servicemesh/federation/server.go
```diff
@@ -30,7 +30,9 @@ import (
 )
 
 const (
-	ExportLabel = "maistra.io/exportAs"
+	ExportLabel     = "maistra.io/exportAs"
+	DiscoveryScheme = "http"
+	DiscoveryPort   = 8188
```
dgn: maybe just keep it as `DefaultDiscoveryPort`? If users want to open a different port on their gateway they should be able to do that IMO

I'm fine with hard-coding for now and creating an issue to make configurable later, though

pkg/servicemesh/federation/controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+func (c *Controller) delete(_ context.Context, instance *v1alpha1.MeshFederation) error {
+	registry := c.getRegistry(instance.Name)
+	if registry != nil {
+		// make sure it's one of ours
+		if registry.Provider() != serviceregistry.Federation {
+			return fmt.Errorf("cannot delete Federation registry for %s, registry belongs to another provider (%s)", instance.Name, registry.Provider())
+		}
+		// unregister federation registry
+		logger.Infof("Removing registry for Federation cluster %s", instance.Name)
+		c.sc.DeleteRegistry(registry.Cluster())
```
dgn: shouldn't we also delete the discovery resources?

pkg/servicemesh/federation/controller/discovery.go
```diff
@@ -0,0 +1,510 @@
...
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			// XXX: We don't support upgrade, so ignore this for now
+			err = nil
+		} else {
+			vs = nil
+			return
+		}
+	}
+
```
dgn, line 223: maybe we should store the resource names in a `map[string]string[]` (MeshFederation name -> resourceName[]) so we can easily find them again for deletion

pkg/servicemesh/apis/servicemesh/v1alpha1/meshfederation_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// the remote.  This is also used to validate certificates used by the
+	// remote services (both client and server certificates).
+	// XXX: maybe this is only used to initiate a connection, with the actual
+	// certs stored in the status field, as retrieved from the remote mesh, or
+	// maybe this identifies an endpoint used to retrieve a cert chain, a la jwks
+	// +optional
+	CertificateChain string `json:"certificateChain,omitempty"`
+
+	// AllowDirectInbound determines whether or not external service
+	// invocations will be terminated at the ingress the gateway.
```
dgn:
```suggestion
	// invocations will be terminated at the ingress gateway.
```

pkg/servicemesh/apis/servicemesh/v1alpha1/meshfederation_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// XXX: should this also be configurable per exported service?
+	// +optional
+	AllowDirectInbound bool `json:"allowDirectInbound,omitempty"`
+
+	// AllowDirectOutbound determines whether or not external service
+	// invocations will be proxied through the gateway
```
dgn:
```suggestion
	// invocations will be proxied through and terminated at 
	// the egress gateway
```

pkg/servicemesh/federation/controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+	"istio.io/pkg/log"
+)
+
+func init() {
+	schemasBuilder := collection.NewSchemasBuilder()
+	schemasBuilder.MustAdd(collections.IstioNetworkingV1Alpha3Destinationrules)
+	schemasBuilder.MustAdd(collections.IstioNetworkingV1Alpha3Virtualservices)
+	schemasBuilder.MustAdd(collections.IstioNetworkingV1Alpha3Gateways)
+	schemasBuilder.MustAdd(collections.IstioSecurityV1Beta1Authorizationpolicies)
+	// XXX: we should consider adding this directly to the service registry
```
dgn, line 51: +1 we might want to create two registry entries directly for proxied services. should we file an issue for it?

pkg/servicemesh/federation/controller/controller_test.go
```diff
@@ -0,0 +1,97 @@
...
+			opt: Options{
+				Client:            nil,
+				ServiceController: &aggregate.Controller{},
+				XDSUpdater:        &v1alpha3.FakeXdsUpdater{},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+		})
```
dgn: I'm guessing this needs some test code ;)

pkg/servicemesh/federation/controller/controller.go
rcernich: Yes.  Originally, these were created as k8s resources, with an owner reference, hence no original cleanup code.  Good catch.

pkg/servicemesh/federation/controller/controller.go
rcernich: Yeah.  I think it complicates things because the federation controller then also becomes a service registry itself, so it will need to integrate as both a config store and service registry.

pkg/servicemesh/federation/server.go
rcernich: I've been thinking about running this through 15443, so it's no different than any other inbound request.

pkg/servicemesh/federation/controller/controller_test.go
rcernich: done

pkg/servicemesh/apis/servicemesh/v1alpha1/meshfederation_types.go
rcernich: done

pkg/servicemesh/apis/servicemesh/v1alpha1/meshfederation_types.go
rcernich: done

pkg/servicemesh/federation/controller/discovery.go
rcernich: the names are static and based on the cr, so it was simpler to just create the names and delete them (three strings).

pkg/servicemesh/federation/controller/controller.go
rcernich: done, with unit test coverage

pkg/servicemesh/federation/server.go
rcernich: changed the name to DefaultDiscoveryPort, but left everything else as-is.

pkg/servicemesh/federation/controller/controller.go
rcernich: I've created an issue for this, MAISTRA-2298

pkg/servicemesh/federation/server.go
dgn: > I've been thinking about running this through 15443, so it's no different than any other inbound request.

I had the same idea :) if we have a 'default' service name for it (e.g. just `discovery`) we would only need to exchange gateway information in the future, and we get mTLS for discovery for free

pkg/kube/controller/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+	c.logger.Info("Starting workers")
+	wait.Until(c.worker, c.resyncPeriod, stopChan)
+}
+
+func (c *Controller) HasSynced() bool {
+	return c.informer.HasSynced()
+}
+
+func (c *Controller) worker() {
```
dgn, line 120: why do we need `worker()`? can't we use `processNextItem()` directly?

pkg/servicemesh/maistra.mk
```diff
@@ -53,11 +55,18 @@ maistra-gen-k8s-client:
...
 	@$(informer_gen) --input-dirs $(kube_api_packages_v1) --versioned-clientset-package $(kube_clientset_package_v1)/$(kube_clientset_name) --listers-package $(kube_listers_package_v1) --output-package $(kube_informers_package_v1) -h $(kube_go_header_text)
+	@$(deepcopy_gen) -i  $(kube_api_packages_v1alpha1) -O zz_generated.deepcopy -h $(kube_go_header_text)
 	@$(client_gen) --clientset-name $(kube_clientset_name) --input-base "" --input  $(kube_api_packages_v1alpha1) --output-package $(kube_clientset_package_v1alpha1) -h $(kube_go_header_text)
 	@$(lister_gen) --input-dirs $(kube_api_packages_v1alpha1) --output-package $(kube_listers_package_v1alpha1) -h $(kube_go_header_text)
 	@$(informer_gen) --input-dirs $(kube_api_packages_v1alpha1) --versioned-clientset-package $(kube_clientset_package_v1alpha1)/$(kube_clientset_name) --listers-package $(kube_listers_package_v1alpha1) --output-package $(kube_informers_package_v1alpha1) -h $(kube_go_header_text)
 	@$(move_generated)
 
+# this is manual for now, but should be moved into a separate maistra/api project
+.PHONY: maistra-gen-k8s-client
+maistra-gen-crd-resources:
```
dgn, line 66: can we add this as a prereq for the `gen` target in `Makefile.core.mk`? That way it is checked automatically by prow

rcernich: prow was failing because controller-gen is not available and I didn't want to dive in and figure out how to make it work.  The real solution is to move all this stuff to the api repo anyway, so this was the simplest solution for the short term.

dgn: agreed. ignore my comment then

dgn (approved):

jwendell (approved):

PR #329, opened 2021-04-30
MAISTRA-2283: Add integration test suite for Maistra 

bison:
This adds the beginnings of an integration test suite for Maistra specific features. It deploys an Istio control-plane with Maistra's options for limiting cluster-scoped access, and provides helpers to manage the ServiceMeshMemberRoll for the control-plane.

I had really hoped to have more here, but it turned out to be a bit more work than I thought, and I'm running out of time. I want to get something up that at least provides a framework and some initial value.


dgn (approved):
let's get this merged and extend it in follow-ups

brian-avery:
This will have to be rewritten for 2.1

dgn:
closing as it is out of date

PR #334, opened 2021-05-10
MAISTRA-2325 Distribute trust bundles over SDS 

dgn:
I extended the istio CA protocol to also send a list of trust bundles -- I'm not sure this is the way to go, as it might end up breaking our compatibility with upstream. It does simplify rotation though, because we can reuse existing code paths that check cert expiry time.

Another option I'm investigating is moving the GetTrustBundles() func into its own gRPC API. I'll double-check to see whether there's already work upstream so we can mimic whatever it is they're doing


dgn:
I had to update go-control-plane to get this working, I'll split that into its own commit to make future cherry-picks easier (the update of go-control-plane will likely not be required in future istio versions, current `master` already has it)

dgn:
looks as though this failing test is pulling from dockerhub.. potentially we're being ratelimited

dgn:
This should now be mostly ready for review (there might still be some unforeseen test or lint failures). I split the part that updates go-control-plane into its own commit so we can skip over it during the next rebase. It's also not very nice to review - focus on the other commit.

I changed this today so that I don't have to fork `istio/api` to get it to work. Instead, we're now running our own gRPC service (on the existing TLS port used by the other gRPC services istiod hosts) that does nothing but return all remote trust bundles. The agent then adds its own root cert (which it knows because it's part of the cert chain it receives) and has the full picture. This splitting allowed me to also implement a Ticker that regularly polls istiod for trust bundles, so that you don't have to restart the pods to update root certs. This actually gives us an edge over upstream when it comes to things such as root cert rotation, which we can now do automatically, whereas in upstream you need to restart all proxies. Pretty cool actually.

Anyway, there are still a few references to my own containers and also a rewire back to the upstream envoy build which I left in here for now in case you want to test this stuff. I'll remove all that before removing the WIP mark. I want to get the maistra/api PR merged first anyway.

Gotta step away from the screen now, have a nice day!

dgn:
Removed the WIP here now. It still depends on https://github.com/maistra/api/pull/11 but it is ready to be reviewed. Focus on the second commit when reviewing

rcernich:
This looks okay to me.  A few comments:

* Do we want to keep the cert definition in MeshFederation, or would it be better to have MeshFederation reference a Secret?  I'm guessing we'll eventually want to pass trust information through the discovery interface, and this would allow istiod to update that information should it change.  We could also update the status of MeshFederation to include any updated cert.  dunno
* I assume there is no push mechanism when it comes to distributing certs, but it seems like it would be advantageous if we pushed changes based on a notification rather than polling (e.g. something similar to endpoints or services).

rcernich (approved):

dgn:
> This looks okay to me. A few comments:
> 
>     * Do we want to keep the cert definition in MeshFederation, or would it be better to have MeshFederation reference a Secret?  I'm guessing we'll eventually want to pass trust information through the discovery interface, and this would allow istiod to update that information should it change.  We could also update the status of MeshFederation to include any updated cert.  dunno

I think a Secret isn't necessary, as the cert is public information anyway. I do agree that it might sense to store it externally so we can easily update it - maybe we should write to a ConfigMap, like istiod does for its own root CA cert? I would keep the field in the MeshFederation around though, so it can be used for easy bootstrapping.

>     * I assume there is no push mechanism when it comes to distributing certs, but it seems like it would be advantageous if we pushed changes based on a notification rather than polling (e.g. something similar to endpoints or services).

That's right, there's currently no push mechanism- I will file an issue for that. I already have an idea how to do that by pushing a custom struct{} through xDS. I'd prefer to do that work upstream though, and then backport it. It's useful for 'normal' (non-federated) deployments too, as it allows root cert rotations without having to restart proxies.

jwendell (approved):
You have some conflicts

PR #335, opened 2021-05-10
MAISTRA-2330: Use maistra/api

jwendell:
Split in several commits for easier review. Basically this:
- Depends on maistra/api: Change code to use this repo
- Remove local stuff that moved to maistra/api (pkg/servicemesh/apis and pkg/servicemesh/client)
- Updates to vendor and license directories
- Makefile changes


Makefile.core.mk
```diff
@@ -510,4 +511,11 @@ include tests/integration/tests.mk
 
 include common/Makefile.common.mk
 
-include pkg/servicemesh/maistra.mk
+#-----------------------------------------------------------------------------
```
rcernich, line 513: I think I'd prefer to keep the include, and have all maistra specific stuff in a separate makefile.  The goal being to minimize the changes in core istio to prevent merge conflicts in the future.  The more changes we make that diverge from upstream, the more likely we are to have to manually process future merges with upstream.

go.mod
```diff
@@ -19,6 +19,16 @@ replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 // See https://github.com/kubernetes/kubernetes/issues/92867, there is a bug in the library
 replace github.com/evanphx/json-patch => github.com/evanphx/json-patch v0.0.0-20190815234213-e83c0a1c26c8
 
+replace k8s.io/client-go => k8s.io/client-go v0.19.3
+
+replace k8s.io/code-generator => k8s.io/code-generator v0.19.3
```
rcernich: Is this still necessary, if we're no longer generating code in maistra/istio?  Also, aren't all these imports resolved transitively?  Is there a need for us to pin them in maistra/istio?  As above, the more changes we make here, the more likely we are to have issues merging in the future (although, I think that's the nature of go.mod and go.sum :/).

Makefile.core.mk
jwendell: Currently we only have `vendor` target which is specific to maistra (this is unlikely to generate conflicts as it's an addition). We are already making other changes to Makefile, so I thought of being simple here.

rcernich: And what about the other changes to this file?

go.mod
jwendell: I can remove `code-generator`. Indeed we no longer use it. `client-go` is used.
`go mod` is like black magic. If I don't put those `replace`s, it will use `v0.20.2`, I have no idea why.

Makefile.core.mk
jwendell: They are minimal as well. My point is that for such minimal changes, having a separate file is not worth, especially given we already have made other changes to this file.

Anyway, I'll move it back to its own file.

manifests/charts/base/templates/crds.yaml
```diff
@@ -1,7 +1,4 @@
 {{- if .Values.base.enableCRDTemplates }}
 {{ .Files.Get "crds/crd-all.gen.yaml" }}
 {{ .Files.Get "crds/crd-operator.yaml" }}
-{{ .Files.Get "crds/maistra.io_meshfederations.yaml" }}
-{{ .Files.Get "crds/maistra.io_serviceexports.yaml" }}
-{{ .Files.Get "crds/maistra.io_servicemeshextensions.yaml" }}
 {{- end }}
```
jwendell: @rcernich Should we keep https://github.com/jwendell/istio-maistra/blob/MAISTRA-2330/manifests/charts/base/templates/role.yaml#L22 in this repo? Or delete it from here as well and move it to operator?

Makefile.core.mk
rcernich: In that case, I guess it doesn't matter.  I didn't realize that we made other changes beyond include maistra.mk.

manifests/charts/base/templates/crds.yaml
rcernich: I think these should stay here.  I think, ideally, we'd like to get to a point where the operator is pulling in the charts without modification.  This may not be an achievable goal, but the closer we can get, the better.

dgn:
Strange failures. Rerunning

/test integration

dgn (approved):

rcernich (approved):

PR #336, opened 2021-05-11
MAISTRA-2295 create CRD for federation ServiceImport

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes


pkg/servicemesh/federation/imports/controller.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return internalNewController(cs, mrc, opt), nil
+}
+
+// allows using a fake client set for testing purposes
+func internalNewController(cs maistraclient.Interface, mrc memberroll.MemberRollController, opt Options) *Controller {
+	logger := common.Logger.WithLabels("component", controllerName)
+	var informer cache.SharedIndexInformer
+	// Currently, we only watch istio system namespace for MeshFederation resources, which is why this block is disabled.
+	if mrc != nil && false {
```
jwendell, line 74: LOL

Is this something we want to handle in the future? If so I'd say to create a ticket and link it here with a `FIXME`

rcernich: yes/no.  I think it makes sense to use a unique namespace for federation, but we have issues deploying gateways to namespaces outside istio-system, so constraining things to istio-system was the cleanest at this point.

jwendell: I don't understand why this block is here if it's never going to be executed and we don't have plans to enable it.

rcernich: So it doesn't get lost?

jwendell: yeah, that's what I thought. That's why I suggested to create a Jira for it so it doesn't get lost. If it's something we want to tackle before 2.1 final, I'd say put it in the sprint with blocker severity so every sprint we remember it :)
(in order to avoid this https://github.com/maistra/istio-operator/blob/maistra-2.1/build/patch-charts.sh#L449)

go.mod
```diff
@@ -113,7 +113,7 @@ require (
 	k8s.io/client-go v0.19.3
 	k8s.io/kubectl v0.19.3
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
-	maistra.io/api v0.0.0-20210511174324-2c8a7ef09df8
+	maistra.io/api v0.0.0-20210518232929-7bcc8ff51069
```
jwendell: this needs to point to `9f28759d3cb`

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -165,51 +224,50 @@ func (c *Controller) convertExportedService(s *federationmodel.ServiceMessage) (
 	if c.useDirectCalls {
 		serviceVisibility = visibility.Public
 	}
-	serviceName := fmt.Sprintf("%s.%s.remote", s.Name, s.Namespace)
-	return c.createService(s, serviceName, s.Hostname, c.clusterID, c.networkName, serviceVisibility, c.gatewayStore)
+	serviceName := fmt.Sprintf("%s.%s.%s.remote", s.Name, s.Namespace, c.clusterID)
```
dgn, line 227: not entirely sure about `.remote` here - it's not a reserved TLD so might be used at some point in the future

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -165,51 +224,50 @@ func (c *Controller) convertExportedService(s *federationmodel.ServiceMessage) (
...
 	serviceVisibility := visibility.Public
-	return c.createService(s, serviceName, importedName.Hostname, c.clusterID, network, serviceVisibility, gateways)
+	return c.createService(s, serviceName, serviceNamespace, importedName.Hostname,
+		c.clusterID, c.localNetwork, serviceVisibility, c.egressGateways, c.egressSAs)
 }
 
-func (c *Controller) createService(s *federationmodel.ServiceMessage, serviceName, hostname, clusterID, network string,
-	serviceVisibility visibility.Instance, networkGateways []*model.Gateway) (*model.Service, []*model.ServiceInstance) {
+func (c *Controller) createService(s *federationmodel.ServiceMessage, serviceName, serviceNamespace, hostname, clusterID, network string,
+	serviceVisibility visibility.Instance, networkGateways []*model.Gateway, sas []string) (*model.Service, []*model.ServiceInstance) {
```
dgn, line 248: nit: any reason not to use `serviceAccounts` instead of `sas`?

pkg/servicemesh/federation/common/util.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/spiffe"
+)
+
+func EndpointsForService(client kube.Client, name, namespace string) (*corev1.Endpoints, error) {
+	return client.KubeInformer().Core().V1().Endpoints().Lister().Endpoints(namespace).Get(name)
+}
+
+func ServiceAccountsForService(client kube.Client, name, namespace string) (map[string]string, error) {
```
jwendell, line 32: It would be great to have comments for exported/public functions, e.g., so that we know what to expect as a result. For instance, on error, why not return nil instead of an empty map?

pkg/servicemesh/federation/server/routing.go
```diff
@@ -74,14 +74,15 @@ func (s *meshServer) removeServiceFromAuthorizationPolicy(target *federationmode
 	}
 	for index, host := range apSpec.Rules[0].To[0].Operation.NotHosts {
 		if host == target.Hostname {
-			apSpec.Rules[0].To[0].Operation.NotHosts = append(apSpec.Rules[0].To[0].Operation.NotHosts[:index], apSpec.Rules[0].To[0].Operation.NotHosts[index+1:]...)
+			apSpec.Rules[0].To[0].Operation.NotHosts = append(apSpec.Rules[0].To[0].Operation.NotHosts[:index],
+				apSpec.Rules[0].To[0].Operation.NotHosts[index+1:]...)
```
jwendell, line 78: Is there any risk of an OutOfRange expection here and in the range statement above? Better add guards?

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -84,34 +98,79 @@ type Options struct {
...
+	}
+}
+
+func (c *Controller) UpdateImportConfig(importConfig *v1alpha1.ServiceImports) {
+	func() {
+		c.storeLock.Lock()
+		defer c.storeLock.Unlock()
+		c.importNameMapper = common.NewServiceImporter(importConfig, c.defaultNameMapper, c.defaultDomainSuffix, c.localDomainSuffix)
+	}()
+	c.resync()
```
jwendell, line 164: Is there any risk of this be so quick that the annon function above doesn't have time to finish?

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -84,34 +98,79 @@ type Options struct {
...
+}
+
+func (c *Controller) UpdateDefaultImportConfig(importConfig *v1alpha1.ServiceImports) {
+	func() {
+		c.storeLock.Lock()
+		defer c.storeLock.Unlock()
+		c.importNameMapper.UpdateDefaultMapper(common.NewServiceImporter(importConfig, nil, c.defaultDomainSuffix, c.localDomainSuffix))
+	}()
+	c.resync()
```
jwendell, line 173: ditto

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -547,67 +650,107 @@ func (c *Controller) updateXDS(hostname, namespace string, instances []*model.Se
...
+			return nil, err
+		}
+
+		// delete the old imported service
+		c.removeServiceFromStore(existing.localName)
+		updatedConfigs[model.ConfigKey{
+			Kind:      gvk.ServiceEntry,
+			Name:      existing.localName.Hostname,
+			Namespace: existing.Namespace,
+		}] = struct{}{}
```
jwendell, line 735: Curious why not just delete the element here and below

pkg/servicemesh/federation/common/namemapping.go
```diff
@@ -0,0 +1,139 @@
...
+func newNameMatcher(mapping *v1alpha1.ServiceNameMapping, domainSuffix string) NameMapper {
+	var alias *v1alpha1.ServiceName
+	// if it's nil or matches anything, it may as well be nil
+	if mapping.Alias == nil ||
+		((mapping.Alias.Namespace == v1alpha1.MatchAny || mapping.Alias.Namespace == "") &&
+			(mapping.Alias.Name == v1alpha1.MatchAny || mapping.Alias.Name == "")) {
+		alias = nil
+	} else {
+		alias = &v1alpha1.ServiceName{}
+		*alias = *mapping.Alias
```
jwendell: Better use DeepCopy? It would be safer if the api object changes to more complex structures in the future.

jwendell (approved):
lgtm, just a few comments and nits

pilot/pkg/serviceregistry/federation/controller.go
rcernich: it's not `go func()...`, so it will be executed like a normal inline function call.  i did it this way to use the `defer` functionality to ensure the lock is freed (basically, nesting scope).

pkg/servicemesh/federation/common/namemapping.go
rcernich: good call.  i'll update to use `DeepCopy()`

pkg/servicemesh/federation/server/routing.go
rcernich: see above:
```
	if len(apSpec.Rules) != 1 || len(apSpec.Rules[0].To) != 1 {
		return fmt.Errorf("invalid AuthorizationPolicy for service export")
	}
```

pilot/pkg/serviceregistry/federation/controller.go
rcernich: This is used to trigger xds, so here we're saying that the service needs to be updated (it's been deleted).  I believe this helps constrain the updates that end up being pushed out to the sidecars (i.e. so we're only pushing updates related to things that have actually changed; changed meaning added, updated, or deleted from the registry).

pkg/servicemesh/federation/imports/controller.go
dgn: +1 to adding a ticket

dgn (approved):
lgtm, just some minor things. we can look at those later on

pilot/pkg/serviceregistry/federation/controller.go
rcernich: This is used in the name field for the `model.Service`, which really isn't used for much, but does appear in telemetry and kiali.  The actual hostname is stored in the `Hostname` field of the service.

PR #342, opened 2021-05-18
MAISTRA-2370: Handle tombstones when handling events from the namespa…

knrc:
…ce specific caches

Handles tombstones when forwarding events from the namespace specific cache to the aggregated cache

### DIFF
--- a/pkg/listwatch/listinformer.go
+++ b/pkg/listwatch/listinformer.go
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"time"
 
+	"istio.io/pkg/log"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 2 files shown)


pkg/listwatch/listinformer.go
```diff
@@ -130,6 +130,10 @@ func (li *listerInformer) stop() {
 }
 
 func (li *listerInformer) newWatchEvent(eventType watch.EventType, obj interface{}) *watch.Event {
+	tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+	if ok {
+		obj = tombstone.Obj
+	}
 	return &watch.Event{
```
jwendell: Worth to add another check like 
```go
if _, ok := obj.(runtime.Object); !ok {
   print err; return nil
}
```
before trying to use it below?

jwendell (approved):

brian-avery (approved):

pkg/listwatch/listinformer.go
knrc: Sounds like a good idea

knrc: @jwendell should be handled now

knrc:
It looks as if something is up with prow, lots of deployments are timing out and not deploying.

knrc:
Prow is not healthy, I'm going to merge this manually

dgn:
I guess it doesn't act on closed PRs :(

PR #348, opened 2021-05-21
MAISTRA-2309 create CRD for FederationStatus

rcernich:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

### DIFF
--- a/pkg/servicemesh/federation/discovery/controller.go
+++ b/pkg/servicemesh/federation/discovery/controller.go
@@ -23,6 +23,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
@@ -40,6 +41,7 @@ import (
 	memberroll "istio.io/istio/pkg/servicemesh/controller"
 	"istio.io/istio/pkg/servicemesh/federation/common"
 	"istio.io/istio/pkg/servicemesh/federation/server"
+	"istio.io/istio/pkg/servicemesh/federation/status"
 )
 
 const controllerName = "federation-discovery-controller"
@@ -51,6 +53,7 @@ type Options struct {
 	Env               *model.Environment
 	ConfigStore       model.ConfigStoreCache
 	FederationManager server.FederationManager
+	StatusManager     status.Manager
 	LocalNetwork      string
 	LocalClusterID    string
 }
@@ -64,6 +67,7 @@ type Controller struct {
 	cs                maistraclient.Interface
 	env               *model.Environment
 	federationManager server.FederationManager
+	statusManager     status.Manager
 	sc                *aggregate.Controller
 	xds               model.XDSUpdater
 	mu                sync.Mutex
@@ -133,6 +137,7 @@ func internalNewController(cs maistraclient.Interface, mrc memberroll.MemberRoll
 		xds:               opt.XDSUpdater,
 		federationManager: opt.FederationManager,
 		trustBundles:      map[string]string{},
+		statusManager:     opt.StatusManager,
 	}
 	internalController := kubecontroller.NewController(kubecontroller.Options{
 		Informer:     informer,
@@ -161,6 +166,9 @@ func (c *Controller) HasSynced() bool {
 func (c *Controller) reconcile(resourceName string) error {
 	c.Logger.Debugf("Reconciling MeshFederation %s", resourceName)
 	defer func() {
+		if err := c.statusManager.PushStatus(); err != nil {
+			c.Logger.Errorf("error pushing FederationStatus for mesh %s: %s", resourceName, err)
+		}
 		c.Logger.Debugf("Completed reconciliation of MeshFederation %s", resourceName)
 	}()
 
@@ -246,7 +254,9 @@ func (c *Controller) update(ctx context.Context, instance *v1alpha1.MeshFederati
 			c.Logger.Errorf("error retrieving ServiceImports associated with MeshFederation %s: %s", instance.Name, err)
 			return err
 		}
-		if err := c.federationManager.AddMeshFederation(instance, exportConfig); err != nil {
+
+		statusHandler := c.statusManager.FederationAdded(types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace})
+		if err := c.federationManager.AddMeshFederation(instance, exportConfig, statusHandler); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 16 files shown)


morvencao:
Hi @rcernich I'm interested with mesh federation feature, will it be available in maistra 2.1? What's the corresponding openshift version we can get this feature?

rcernich:
@jwendell, @dgn, this just needs a couple of reviews and it can be merged.

rcernich:
@dgn, @jwendell, could you please review this?

pkg/servicemesh/federation/status/manager.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	oldBytes, err := json.Marshal(&v1alpha1.FederationStatus{
+		Status: m.status,
+	})
+	if err != nil {
+		return err
+	}
+
+	m.logger.Debugf("old bytes:\n%s\n", string(oldBytes))
+	m.logger.Debugf("new bytes:\n%s\n", string(newBytes))
```
jwendell, line 185: Do we really need these? Is this leaking any confidential info?

pkg/servicemesh/federation/status/manager.go
```diff
@@ -0,0 +1,250 @@
...
+
+	fedStatus, err := m.cs.CoreV1alpha1().FederationStatuses(m.name.Namespace).
+		Patch(context.TODO(), m.name.Name, types.MergePatchType, patch, metav1.PatchOptions{}, "status")
+	if err == nil {
+		m.missingCRDErrorLogged = false
+		m.status = *fedStatus.Status.DeepCopy()
+		return nil
+	} else if !(errors.IsGone(err) || errors.IsNotFound(err)) {
+		return err
+	} else if meta.IsNoMatchError(err) {
```
jwendell, line 205: shouldn't this `if` be above the previous one?

pkg/servicemesh/federation/status/manager.go
```diff
@@ -0,0 +1,250 @@
...
+		// The FederationStatus resource is not installed.  We won't error in this case.
+		if !m.missingCRDErrorLogged {
+			m.logger.Errorf("status will not be updated: FederationStatus CRD is not installed")
+			m.missingCRDErrorLogged = true
+		}
+		return nil
+	}
+
+	// if we get here, it means the resource doesn't exist
+	pilotPod, err := m.client.CoreV1().Pods(m.name.Namespace).Get(context.TODO(), m.name.Name, metav1.GetOptions{})
```
jwendell: do we have this info in istiod cache so that we don't need to reach the api server? Maybe this is not an issue at all, I'm just remembering the issues we had with a large number of namespaces...

pkg/servicemesh/federation/federation.go
```diff
@@ -72,6 +75,11 @@ func New(opt Options) (*Federation, error) {
 	if err := opt.validate(); err != nil {
 		return nil, err
 	}
+	name := types.NamespacedName{Name: os.Getenv("POD_NAME"), Namespace: os.Getenv("POD_NAMESPACE")}
```
jwendell: Do you have access to pilotArgs here? If yes, this struct already have these variables for free: https://github.com/maistra/istio/blob/8693f7adc22cdb8ee2db0854519320bece8bad84/pilot/pkg/bootstrap/options.go#L152

rcernich: I thought I investigated this, but I'll check again.

pkg/servicemesh/federation/status/manager.go
rcernich: This is helpful for debugging.  As this is the status that we'll be writing out, no leakage.

pkg/servicemesh/federation/status/manager.go
rcernich: Yes it should.  Good catch!

pkg/servicemesh/federation/status/manager.go
rcernich: This is well scoped, but yes, I think using the lister would be more efficient.

pkg/servicemesh/federation/federation.go
rcernich: updated

pkg/servicemesh/federation/status/manager.go
rcernich: done

pkg/servicemesh/federation/status/manager.go
rcernich: done

jwendell (approved):

pkg/servicemesh/federation/server/server.go
```diff
@@ -486,6 +508,9 @@ func (s *meshServer) resync() {
 			s.addService(svcKey, svcMessage)
 		}
 	}
+	if err := s.statusHandler.Flush(); err != nil {
```
dgn, line 511: do we need to manually `Flush()` from outside the status handler? couldn't this be a private func as well?

dgn: nevermind I see now that we flush manually here, I guess because there might be >1 event at a time. I think I'd prefer the flush() method to be private and triggering it automatically (using debounce) from within statushandler - but that's an optimization (reason is simply that it's not obvious to me when I need to flush)

pkg/servicemesh/federation/discovery/controller.go
```diff
@@ -161,6 +166,9 @@ func (c *Controller) HasSynced() bool {
 func (c *Controller) reconcile(resourceName string) error {
 	c.Logger.Debugf("Reconciling MeshFederation %s", resourceName)
 	defer func() {
+		if err := c.statusManager.PushStatus(); err != nil {
```
dgn, line 169: ~why `PushStatus()` here? doesn't `Flush()` do the same if it's required?~

okay I should have read more carefully.. PushStatus() is on the manager, Flush() on the handler- now it makes sense

pkg/servicemesh/federation/status/handler.go
```diff
@@ -0,0 +1,448 @@
...
+		connections = append(connections, connection)
+	}
+	h.discoveryDirty = h.discoveryDirty || len(connections) > 0
+	for _, connection := range connections {
+		delete(h.discovery, connection)
+	}
+}
+
+// Write status
+func (h *handler) Flush() error {
```
dgn, line 375: nit: I think this could use a comment when to use `Flush()` and when it's not required (some funcs seem to call `Flush()` but some don't)

dgn (approved):

pkg/servicemesh/federation/server/server.go
rcernich: I chose to have the status automatically written for important events, like connect/disconnect and full sync, and to let the periodic updates write out status for other events.  The main reason I did this was to avoid spamming the api server when processing a full sync.  This means the user will immediately be aware of the connection status, as well as the imported/exported services after full syncs.  Interim events will not be immediately reflected in the status (e.g. a new service is created in the mesh and gets exported).

PR #352, opened 2021-05-26
MAISTRA-2389: Ensure the status code can handle Forbidden errors

knrc:
Fix status update code to handle Forbidden errors.

### DIFF
--- a/pilot/pkg/status/state.go
+++ b/pilot/pkg/status/state.go
@@ -174,8 +174,8 @@ func (c *DistributionController) writeStatus(ctx context.Context, config Resourc
 	// should this be moved to some sort of InformerCache for speed?
 	current, err := resourceInterface.Get(ctx, config.Name, metav1.GetOptions{ResourceVersion: config.ResourceVersion})
 	if err != nil {
-		if errors.IsGone(err) || errors.IsNotFound(err) {
-			// this resource has been deleted.  prune its state and move on.
+		if errors.IsGone(err) || errors.IsNotFound(err) || errors.IsForbidden(err) {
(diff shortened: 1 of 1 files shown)


pilot/pkg/status/state.go
```diff
@@ -174,8 +174,8 @@ func (c *DistributionController) writeStatus(ctx context.Context, config Resourc
 	// should this be moved to some sort of InformerCache for speed?
 	current, err := resourceInterface.Get(ctx, config.Name, metav1.GetOptions{ResourceVersion: config.ResourceVersion})
 	if err != nil {
-		if errors.IsGone(err) || errors.IsNotFound(err) {
-			// this resource has been deleted.  prune its state and move on.
+		if errors.IsGone(err) || errors.IsNotFound(err) || errors.IsForbidden(err) {
```
rcernich, line 177: I think this works around the problem, but should it be happening at all if the controller has the correct list of members?

knrc: It has nothing to do with the controller, this is going direct to the API server which is why we get the problem

knrc: Note the way this works in that there is a configmap created in the control plane namespace containing status updates for resources which have been propagated, the leader then updates those resources directly by going straight to the API server.  There is no integration with SMMR updates.

knrc: I also checked other parts of the codebase to see if this was used elsewhere but could not find anything.

rcernich (approved):

jwendell (approved):

PR #357, opened 2021-06-04
MAISTRA-2390 Push trust bundle updates through xDS

dgn:
This replaces a bunch of code that was committed as part of #334 - looking back, I should have implemented it using a push mechanism from the start.

This completely removes the gRPC service that was previously used for polling. I could have even simplified the proto in maistra/api, but we can remove that in 2.2 anyway, so I'll keep it as it is.

### DIFF
--- a/pilot/pkg/xds/tbds.go
+++ b/pilot/pkg/xds/tbds.go
@@ -1,0 +1,68 @@
+// Copyright Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package xds
+
+import (
+	v1 "maistra.io/api/security/v1"
+
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pilot/pkg/networking/util"
+	fedmodel "istio.io/istio/pkg/servicemesh/federation/model"
+)
+
+// TbdsGenerator generates trust bundle configuration for proxies to consume
+type TbdsGenerator struct {
+	TrustBundleProvider fedmodel.TrustBundleProvider
+}
+
+var _ model.XdsResourceGenerator = &TbdsGenerator{}
+
+func tbdsNeedsPush(req *model.PushRequest) bool {
+	if req == nil {
+		return true
+	}
+
+	if !req.Full {
+		return false
+	}
+
+	if len(req.ConfigsUpdated) == 0 {
+		return true
+	}
+
+	return false
+}
+
+// Generate returns protobuf containing TrustBundle for given proxy
+func (e *TbdsGenerator) Generate(proxy *model.Proxy, push *model.PushContext, w *model.WatchedResource, req *model.PushRequest) model.Resources {
+	if !tbdsNeedsPush(req) {
+		return nil
+	}
+	if e.TrustBundleProvider == nil {
+		return nil
+	}
+	tb := &v1.TrustBundleResponse{}
+	trustBundles := e.TrustBundleProvider.GetTrustBundles()
+	if len(trustBundles) == 0 {
+		return nil
+	}
+	for td, cert := range trustBundles {
+		tb.TrustBundles = append(tb.TrustBundles, &v1.TrustBundle{
+			TrustDomain: td,
+			RootCert:    cert,
+		})
+	}
+	return model.Resources{util.MessageToAny(tb)}
+}
--- a/pilot/pkg/xds/v3/model.go
+++ b/pilot/pkg/xds/v3/model.go
@@ -25,6 +25,8 @@ const (
 	RouteType     = resource.RouteType
 	SecretType    = resource.SecretType
 	NameTableType = "type.googleapis.com/istio.networking.nds.v1.NameTable"
+	// maistra xDS
+	TrustBundleType = "type.googleapis.com/maistra.security.v1.TrustBundleResponse"
 )
 
 // GetShortType returns an abbreviated form of a type, useful for logging or human friendly messages
@@ -42,6 +44,8 @@ func GetShortType(typeURL string) string {
 		return "SDS"
 	case NameTableType:
 		return "NDS"
+	case TrustBundleType:
+		return "TBDS"
 	default:
 		return typeURL
 	}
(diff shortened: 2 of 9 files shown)


jwendell (commented):
Is it too much effort to write an integration test for this feature?

dgn:
> Is it too much effort to write an integration test for this feature?

Good question.. It does require setting up a federation, which in turn requires our Maistra CRDs to be present. It should be possible, but we need some groundwork

pilot/pkg/xds/tbds.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// TbdsGenerator generates trust bundle configuration for proxies to consume
+type TbdsGenerator struct {
+	TrustBundleProvider fedmodel.TrustBundleProvider
+}
+
+var _ model.XdsResourceGenerator = &TbdsGenerator{}
+
+func tbdsNeedsPush(req *model.PushRequest) bool {
```
jwendell, line 32: I'm too noob here :) The way I interpret this function is the opposite: if req is null, there's no need to push. Similarly, if len(req.ConfigsUpdated) == 0 we don't need to push :/

pilot/pkg/xds/v3/model.go
```diff
@@ -42,6 +44,8 @@ func GetShortType(typeURL string) string {
 		return "SDS"
 	case NameTableType:
 		return "NDS"
+	case TrustBundleType:
+		return "TBDS"
```
jwendell, line 48: sorry, I can't avoid to read "To be defined" when I see TBD :)

pilot/pkg/xds/tbds.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	v1 "maistra.io/api/security/v1"
+
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pilot/pkg/networking/util"
+	fedmodel "istio.io/istio/pkg/servicemesh/federation/model"
+)
+
+// TbdsGenerator generates trust bundle configuration for proxies to consume
+type TbdsGenerator struct {
```
rcernich, line 26: nit: isn't go standard `TBDSGenerator`?  I guess that runs counter to `XdsGenerator`

pkg/istio-agent/xds_proxy.go
```diff
@@ -319,6 +330,46 @@ func (p *XdsProxy) HandleUpstream(ctx context.Context, con *ProxyConnection, xds
...
+					} else if certExpireTime.Before(expireTime) {
+						expireTime = certExpireTime
+					}
+					trustBundles[bundle.TrustDomain] = certBytes
+				}
+
+				sc, ok := p.agent.WorkloadSecrets.(*cache.SecretCache)
+				if !ok || sc == nil {
+					proxyLog.Errorf("failed to access secret cache", err)
+					return nil
```
jwendell: Should you return the error here?

pilot/pkg/xds/tbds.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if !req.Full {
+		return false
+	}
+
+	if len(req.ConfigsUpdated) == 0 {
+		return true
+	}
+
+	return false
```
rcernich, line 45: If it's a full push, do we always need to push? Is it possible to check the ConfigsUpdated for a TB kind?

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -201,21 +200,33 @@ func (sc *SecretCache) setRootCert(rootCert []byte, rootCertExpr time.Time) {
...
+			earliestExpiry = sc.rootCertExpireTime
+		}
+	}
 	sc.rootCertMutex.RUnlock()
 	return trustBundles, earliestExpiry
 }
 
-// setRootCert sets root cert into cache. This method is thread safe.
-func (sc *SecretCache) setTrustBundles(trustBundles map[string][]byte, earliestExpiry time.Time) {
+// setRootCert stores trust bundles into the cache. This method is thread safe.
```
jwendell: wrong function name?

pilot/pkg/xds/tbds.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	tb := &v1.TrustBundleResponse{}
+	trustBundles := e.TrustBundleProvider.GetTrustBundles()
+	if len(trustBundles) == 0 {
+		return nil
+	}
+	for td, cert := range trustBundles {
+		tb.TrustBundles = append(tb.TrustBundles, &v1.TrustBundle{
+			TrustDomain: td,
+			RootCert:    cert,
```
rcernich, line 64: Not related to this PR, but will we need to include a complete cert chain at some point?

rcernich (approved):
lgtm.  one question, other than validating discovery communication in the ingress gateway, do we need to distribute trust bundles for remote meshes if we're not supporting direct communication?  (sorry, just thought of this.)

security/pkg/nodeagent/cache/secretcache.go
```diff
@@ -201,21 +200,33 @@ func (sc *SecretCache) setRootCert(rootCert []byte, rootCertExpr time.Time) {
...
+func (sc *SecretCache) SetTrustBundles(trustBundles map[string][]byte, earliestExpiry time.Time) {
 	sc.rootCertMutex.Lock()
-	sc.trustBundleMap = trustBundles
-	sc.trustBundleExpireTime = earliestExpiry
+	trustBundlesUpdated := !compareTrustBundles(sc.trustBundleMap, trustBundles)
```
jwendell, line 221: probably not in the scope of this pr, but this function should better be called `trustBundlesEqual` or something like that. If trustBundle were a type, we could have a `func (tb TrustBundle) Equals(another TrustBundle)`.

pilot/pkg/xds/tbds.go
dgn: agreed. for some reason though this is the style they use for all xDS generators

pilot/pkg/xds/v3/model.go
dgn: yeah.. note that the whole TBDS stuff is just a stopgap until PCDS (Proxy Config Discovery Service) lands (coming in 1.10), that supports pushing trust bundles. All of this is mostly a blatant copy of PCDS

pkg/istio-agent/xds_proxy.go
dgn: I should. will fix

pilot/pkg/xds/tbds.go
dgn: I don't think so - if the presented cert chain can be traced back to a trusted root, it will be accepted

pilot/pkg/xds/tbds.go
dgn: again, copied from PCDS -- I was a bit confused about this part as well, but that's how they do it upstream: https://github.com/istio/istio/blob/master/pilot/pkg/xds/pcds.go

dgn: I guess they didn't want to introduce a fake config type (that is usually bound to a GroupVersionKey which we don't have for this stuff...)

security/pkg/nodeagent/cache/secretcache.go
dgn: yup. will fix

jwendell (approved):

PR #361, opened 2021-06-09
MAISTRA-2249: Remove outdated comment

jwendell:
This is not an issue anymore. With the new podman shipped in our
base images (ubi8, based on rhel 8.4), this command (`podman image ls
-nq`) works fine. This is also valid for 2.0, because both use the same
ubi8 base image.


brian-avery (approved):

brian-avery:
Who knew a comment could break unit tests? 😛 
/test unit

knrc (approved):

PR #362, opened 2021-06-10
MAISTRA-2412 Fix ensureNamespaceExists() in ior/route.go

luksa:
Previously, ensureNamespaceExists() sometimes never actually checked
whether the namespace is present in the list, because the timeout
check was performed before the namespace check. This only happened
in unit tests, because handleEventTimeout is set to only 1ms, whereas
in production, it's set to 10s. This caused the tests TestPerf()
and TestConcurrency() in ior_test.go to fail on occasion.

This commit moves the timeout check after the namespace check so that
the latter is performed at least once.


jwendell (approved):
Thanks for looking into this!

brian-avery (approved):

PR #363, opened 2021-06-11
MAISTRA-2254: MEC: Garbage collects Image Streams

jwendell:
Set the `OwnerReferences` field of the Image Streams we create so
that they are automatically removed when the corresponding
`ServiceMeshExtension` resource is removed.

### DIFF
--- a/mec/pkg/model/interfaces.go
+++ b/mec/pkg/model/interfaces.go
@@ -16,11 +16,18 @@ package model
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 type ImagePullStrategy interface {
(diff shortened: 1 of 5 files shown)


rcernich:
I don't understand this.  Why do we not add the OwnerReference when we create the ImageStream?  If we aren't the one who created the ImageStream, we should not be adding an OwnerReference at all.

jwendell:
We create ImageStreams not directly, but indirectly via ImageStreamImport. If we add the ownerReference to ImageStreamImport, it gets lost when the ImageStream is created. So, we wait for it to be created and then patch it.

knrc (approved):

shamsher31 (approved):

dgn:
Another solution could have been for us to create the `ImageStream` prior to doing any imports

PR #379, opened 2021-06-16
Rebase to Istio 1.9.5

luksa:


samples/bookinfo/platform/kube/bookinfo.yaml
```diff
@@ -72,7 +72,7 @@ spec:
       serviceAccountName: bookinfo-details
       containers:
       - name: details
-        image: docker.io/istio/examples-bookinfo-details-v1:1.16.2
+        image: maistra/examples-bookinfo-details-v1:2.1.0
```
brian-avery, line 75: @luksa is this file wrong? The images below are 2.0, this is 2.1

luksa: I guess so. I've now changed all tags in bookinfo.yaml to 2.1.0 so they match those in the files with individual deployments (e.g. bookinfo-details.yaml, etc.). Thanks.

brian-avery: Thanks :)

jwendell:
@luksa if we don't use that feature, and it's blocking this to be merged just because the test is failing, I'd say just disable the failing test and create a Jira to keep track of it and re-enable it later.

luksa:
Yeah, it looks like only the TPROXY interception mode is causing issues. And I assume we don't use/support it, because it requires CAP_NET_ADMIN for the sidecar. But I'm close to fixing this in the code, so I probably won't need to disable the test. I'm running the integration tests locally now. If they pass, I'll update the PR and see if they pass in prow. My solution may not be perfect, but we can improve on it next week.

dgn (approved):
lgtm - at least I couldn't spot anything obvious. let's get this merged together with the operator changes and have the acceptance tests run against it

dgn:
Did you run `make lint` against your changes? I spotted some linter issues while working on my cherry-pick and they're not all in code I touched

luksa:
Yup, make lint doesn't complain when I run it against the final state. But I didn't run it for each individual commit.

dgn:
> Yup, make lint doesn't complain when I run it against the final state. But I didn't run it for each individual commit.

Thanks, looks like it's my fault then :)

PR #381, opened 2021-06-16
MAISTRA-2437: Push configuration as String

jwendell:
Envoy doesn't handle well passing a Struct as the configuration field.
When an extension retrieves its value, it gets binary data.

Passing string instead is safe. Extensions will get the value as string
and can unmarshal it into a proper JSON struct.

While on it, also fixed a potential `index out of range` error.


brian-avery (approved):

dgn (approved):

dgn (approved):

jwendell:
Changing the base branch didn't work well. Will create another PR.

PR #383, opened 2021-06-24
MAISTRA-2463: copy seed namespaces when set

knrc:
Ensure calls to setSeedNamespaces will not update the input slice.

### DIFF
--- a/pkg/servicemesh/controller/memberroll/controller.go
+++ b/pkg/servicemesh/controller/memberroll/controller.go
@@ -117,7 +117,7 @@ func (smmrc *serviceMeshMemberRollController) Start(stop chan struct{}) {
 	for _, obj := range smmrc.informer.GetStore().List() {
 		serviceMeshMemberRoll := getServiceMeshMemberRoll(obj)
 		if smmrc.memberRollName == serviceMeshMemberRoll.Name {
-			seedNamespaces := smmrc.getNamespaces(serviceMeshMemberRoll.Status.ConfiguredMembers)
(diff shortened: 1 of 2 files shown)


luksa (approved):

pkg/servicemesh/controller/memberroll/controller.go
```diff
@@ -117,7 +117,7 @@ func (smmrc *serviceMeshMemberRollController) Start(stop chan struct{}) {
 	for _, obj := range smmrc.informer.GetStore().List() {
 		serviceMeshMemberRoll := getServiceMeshMemberRoll(obj)
 		if smmrc.memberRollName == serviceMeshMemberRoll.Name {
-			seedNamespaces := smmrc.getNamespaces(serviceMeshMemberRoll.Status.ConfiguredMembers)
```
dgn, line 120: does `.ConfiguredMembers` now contain the control plane namespace? if not, this is changing behavior, right?

luksa: The control plane namespace gets added in https://github.com/Maistra/istio/blob/e25b1b454352642d6b52fd241344d6c00a1ae71a/pkg/servicemesh/controller/memberroll/controller.go#L216

dgn: yes, but that isn't called for seeding anymore now, right? maybe I'm overlooking something

dgn: oh nevermind it is called in the updateNamespaces func. ignore my comment

dgn (approved):

PR #388, opened 2021-06-30
Additional cherry-picks for 1.9

luksa:


dgn:
this looks like it only touched our code, so I assume the cherry-picks went smoothly? anything we should focus on? otherwise lgtm

dgn (approved):

PR #390, opened 2021-07-05
MAISTRA-2478 Honor enableIngressClassName in ingress status syncer

luksa:


luksa:
@shamsher31 After this is merged, you can revert the following change in the operator: https://github.com/maistra/istio-operator/commit/c2c78966f39ce507fccc02d04d3f0b318ef082ea#diff-c5f5b082b87676e0375856cd9a705a245141c481b76b35546eda9c0487646ac6R234-R237

pilot/pkg/config/kube/ingress/status_test.go
```diff
@@ -115,7 +115,7 @@ func makeStatusSyncer(t *testing.T) *StatusSyncer {
 
 	client := kubelib.NewFakeClient()
 	setupFake(t, client)
-	sync := NewStatusSyncer(fakeMeshHolder("istio-ingress"), client, false)
+	sync := NewStatusSyncer(fakeMeshHolder("istio-ingress"), client, false, true)
```
shamsher31, line 118: Does this means `IngressClassName` will be enabled by default?

shamsher31, line 118: `IngressClassName` is disabled by default right?

luksa: No. In Istio itself, it's enabled, but we disable it in Maistra via `enableIngressClassName=false`. Here, it has to be enabled for the unit tests to pass.

shamsher31 (approved):

brian-avery (approved):

brian-avery:
@luksa is this ready for an `ok-to-merge`?

PR #408, opened 2021-07-26
use -f test to not only check they are non-empty but also that they point to valid files

jmazzitelli:


jmazzitelli:
I was afraid of setting these to the wrong things - I looked to see if the script will tell me if I set them incorrectly, and there really isn't a check done by these scripts to tell me that they are really pointing to what it expected (and that is - the kubeconfig file). The `-z` just checks to see that you set them - but you should also check that they at least point to a valid file (and not, say, an invalid file or a directory instead of a file).

jwendell (approved):

luksa (approved):

PR #413, opened 2021-07-29
Get mesh2 ports from correct resource

FilipB:
This PR is fixing installation script for federation poc example. It was not getting ports for mesh2 correctly.

[X] Does not have any changes that may affect Istio users.


dgn (approved):
Thanks @FilipB!

luksa (approved):

PR #417, opened 2021-08-16
MAISTRA-2297 Support updates of federation resources

dgn:
Getting this working required quite a few changes to the federation serviceregistry and the discovery controller, including:
- addition of `restartWatch()` method to stop and restart a running watch (for when watched URL changes)
- untangling of controller.Options and settings derived from ServiceMeshPeer and ImportedServiceSet resource
- addition of UpdatePeerConfig() and UpdateImportConfig() methods to update CRD-derived controller fields
- refactor of discovery resource creation to remove some duplication
- when a watch is successfully established, we now trigger a resync first thing. this helps when the controller is started before the remote endpoints are reachable: previously, the watch would retry but the resync would just time out, meaning we'd have to wait for new events or the 5 minute resync timer to actually see exported services. in a cluster where no service import updates are happening (my test cluster, ahem), this was frustrating.
- addition of the 'remote' header in discovery requests. this is to avoid a race between an updated `spec.Remote` and a watch restart before the proxy would have applied the new configuration - we would restart the watch and connect to the old endpoint again. hashing the remote and including it in the header makes sure we never connect to an outdated endpoint.

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -15,10 +15,12 @@
 package federation
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
+	"reflect"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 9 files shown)


rcernich (approved):

jwendell (approved):

dgn:
I had to move the `FakeStatusHander` to a common package to avoid a race in the discovery controller's ReconcileTest that was only occurring on slow systems (prow :wink: )

PR #422, opened 2021-08-20
MAISTRA-2610 Prefix federation discovery endpoints with /v1/

dgn:
If we change the protocol, we can add v2 endpoints and stay backwards-compatible


rcernich (commented):
I guess we don't have to worry about falling back to from v2 to v1 until we add v2.  Do we need a service to get the supported protocol versions?

dgn:
> I guess we don't have to worry about falling back to from v2 to v1 until we add v2. Do we need a service to get the supported protocol versions?

Yeah, I guess we can think about that when we build v2. If we add that kind of endpoint, it can be `/v2/versions` or similar, and if it's not available, it means we need to fallback to v1

rcernich (approved):

jwendell (approved):

PR #424, opened 2021-08-24
WIP: MAISTRA-2438 Verify that multiple services cannot be exported with the same hostname

shamsher31:


rcernich:
Hey @shamsher31, I think we need to make the check when we export the service.  We should make sure the service export logic is consistent, so that if there are duplicate exports, we always export the same service each time, and log a warning for the other services each time we process.  (This might be something that should go into the status field at some point, but don't worry about that.)  Another alternative is to not export either service if there is a duplicate, although that could be trickier, if a service is already exported, then a duplicate occurs.  I think Istio generally tries to use the creation time on the resources to determine precedence.

PR #426, opened 2021-08-24
Fixes to automatic route creation

jwendell:
This PR addresses two issues:
- MAISTRA-2205: Add an option to opt-out for automatic route creation
- MAISTRA-2375: Do not create automatic routes for Federation Gateways

### DIFF
--- a/pilot/pkg/config/kube/ior/ior_test.go
+++ b/pilot/pkg/config/kube/ior/ior_test.go
@@ -79,6 +79,7 @@ func TestCreate(t *testing.T) {
 		expectedRoutes int
 		expectedError  string
 		tls            bool
+		annotations    map[string]string
 	}{
 		{
 			"One host",
@@ -88,6 +89,7 @@ func TestCreate(t *testing.T) {
 			1,
 			"",
 			false,
+			nil,
 		},
 		{
 			"Two hosts",
@@ -97,6 +99,7 @@ func TestCreate(t *testing.T) {
 			2,
 			"",
 			false,
+			nil,
 		},
 		{
 			"Wildcard 1",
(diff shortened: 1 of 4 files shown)


dgn (approved):
lgtm when linter is fixed. Thanks!

dgn:
Wait.. so your fix for us creating Routes for egress gateways is by adding this option? Can't we skip Gateways that attach to istio-egressgateway instances? They will never need a Route

egress gateways created with OSSM should always have the label:
`istio: egressgateway`

jwendell:
This label is not hardcoded. Users can create ingress/egress gateways with their own labels.

The first commit adds the ability to skip route creation based on a annotation. This fixes the general use-case.
The second one makes use of this new feature to not create routes for federation gateways (ingress or egress)

dgn:
> This label is not hardcoded. Users can create ingress/egress gateways with their own labels.
> 
> The first commit adds the ability to skip route creation based on a annotation. This fixes the general use-case.
> The second one makes use of this new feature to not create routes for federation gateways (ingress or egress)

They can explicitly override it, but this label is always added by default; in fact gateways don't even work without it (I tested that recently). So we should be able to use that to figure out if a gateway is an egress gateway. That way, at least the default egress gateways don't get routes created. We can then still document your way of adding the annotation to disable it for gateways without the label

dgn (approved):

rcernich (approved):

yxun:
@jwendell do we need to back port this fix into 2.0 and 1.1 branch ?

yxun:
@jwendell do we need a similar fix in 2.0 branch ?

PR #430, opened 2021-08-26
MAISTRA-2613 Update to latest Istio 1.9.8

shamsher31:


shamsher31:
@jwendell There was a conflict in istio.deps for proxy sha as you mentioned, so I have left it as it as we have a seperate proxy for maistra. because of which few tests are still failing.

shamsher31:
I think we need to update something once https://github.com/maistra/envoy/pull/104 is merged?

go.mod
```diff
@@ -105,9 +106,9 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	helm.sh/helm/v3 v3.4.2
 	honnef.co/go/tools v0.0.1-2020.1.5 // indirect
-	istio.io/api v0.0.0-20210503211644-902e709f95c4
-	istio.io/client-go v1.9.2
-	istio.io/gogo-genproto v0.0.0-20210420211914-9cbf6943c732
+	istio.io/api v0.0.0-20210607161139-ab961225b5e9
+	istio.io/client-go v1.9.5-0.20210607162355-6a6709ba5473
```
shamsher31, line 110: Is this also needs to be updated to https://github.com/istio/client-go/releases/tag/1.9.8?

I could not find any commit related to upgrading it to 1.9.8.

jwendell: don't worry about this right now. This can be updated by running `bin/update_deps.sh` later, after this is merged

jwendell:
> I think we need to update something once [maistra/envoy#104](https://github.com/maistra/envoy/pull/104) is merged?

nope. After this Envoy PR merges, automator will take care of the rest.

jwendell:
@shamsher31 All tests pass with the latest (1.9.8) proxy: https://github.com/maistra/istio/pull/429
Feel free to update istio.deps in this PR... Or get the PR above merged first (needs a second approval)...

shamsher31:
@jwendell Test failures are now fixed. PTAL.
I think there are only a few changes in charts, so the operator side will be quick to update.

luksa:
LGTM 

Do we want to merge this via a merge-commit so the changes are nicely separated from the maistra commits?

luksa (approved):

jwendell:
> Do we want to merge this via a merge-commit so the changes are nicely separated from the maistra commits?

@luksa What do you mean? I already added the `not squash` label. Isn't it enough?

tools/istio-iptables/pkg/cmd/run_test.go
```diff
@@ -418,24 +418,24 @@ func TestHandleInboundIpv6RulesWithOutboundPorts(t *testing.T) {
...
-		"ip6tables -t nat -A ISTIO_OUTPUT -o lo ! -d ::1/128 -m owner --uid-owner 1337 -j ISTIO_IN_REDIRECT",
-		"ip6tables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --uid-owner 1337 -j RETURN",
-		"ip6tables -t nat -A ISTIO_OUTPUT -m owner --uid-owner 1337 -j RETURN",
-		"ip6tables -t nat -A ISTIO_OUTPUT -o lo ! -d ::1/128 -m owner --gid-owner 1337 -j ISTIO_IN_REDIRECT",
-		"ip6tables -t nat -A ISTIO_OUTPUT -o lo -m owner ! --gid-owner 1337 -j RETURN",
-		"ip6tables -t nat -A ISTIO_OUTPUT -m owner --gid-owner 1337 -j RETURN",
-		"ip6tables -t nat -A ISTIO_OUTPUT -d ::1/128 -j RETURN",
-		"ip6tables -t nat -A ISTIO_OUTPUT -p tcp --dport 32000 -j ISTIO_REDIRECT",
-		"ip6tables -t nat -A ISTIO_OUTPUT -p tcp --dport 31000 -j ISTIO_REDIRECT",
+		"ip6tables -w -t nat -N ISTIO_INBOUND",
```
jwendell, line 421: why?
Why does it pass upstream and fail with us?

pilot/pkg/networking/core/v1alpha3/gateway_test.go
```diff
@@ -39,67 +37,10 @@ import (
 	"istio.io/istio/pilot/test/xdstest"
 	"istio.io/istio/pkg/config"
 	"istio.io/istio/pkg/config/schema/gvk"
-	tls_features "istio.io/istio/pkg/features"
 	"istio.io/istio/pkg/proto"
 )
 
 func TestBuildGatewayListenerTlsContext(t *testing.T) {
-	runTestBuildGatewayListenerTLSContext(t, &auth.TlsParameters{})
```
jwendell: why this commit was necessary?
I mean, this commit essentially makes this file equal to upstream. Does it mean we changed it before? Maybe in `fix test` commit? If so, it should be enough to remove it, right?

pkg/servicemesh/federation/status/manager_test.go
```diff
@@ -573,7 +573,7 @@ func TestStatusManager(t *testing.T) {
 			},
 			assertions: []func(t *testing.T, status *v1.ServiceMeshPeerStatus){
 				func(t *testing.T, status *v1.ServiceMeshPeerStatus) {
-					if status.DiscoveryStatus.Inactive[0].Watch.LastEvent.IsZero() {
+					if len(status.DiscoveryStatus.Inactive) > 0 && status.DiscoveryStatus.Inactive[0].Watch.LastEvent.IsZero() {
```
jwendell: I **think** this should be
```
if len(status.DiscoveryStatus.Inactive) == 0 || (len(status.DiscoveryStatus.Inactive) > 0 && status.DiscoveryStatus.Inactive[0].Watch.LastEvent.IsZero())
```

@rcernich what's your input here?

luksa:
Even with don't squash, we end up with a linear history. As much as I prefer linear history for day-to-day stuff, I think that in this particular case a merge commit would be better. This is what we'd end up with:
```
       U1---U2---U3 
     /              \
1.9.5---M1--M2--M3---X  maistra-2.1
```
where `U1..U3` are upstream commits (from 1.9.5. to 1.9.8), `M1-M3` are the existing Maistra commits that we added on top of 1.9.5, and `X` is the merge commit that merges the two linear histories. 

This will leave the resulting history slightly better organized, as it would keep our changes separate from those upstream.

tools/istio-iptables/pkg/cmd/run_test.go
shamsher31: I compared the test output and saw that the actual output needs `-w`. not sure what it does.

jwendell:
Ah, got it. We don't use merge strategy by default. Let me look if there's an option for that

pilot/pkg/networking/core/v1alpha3/gateway_test.go
shamsher31: Not sure if we added it before. The test was failing because [tlsParams](https://github.com/maistra/istio/pull/430/commits/a2dad5095041f24e3ee0fa4d095e29f664f4b9f0#diff-7abd8e9bc7226623690e85b5acc0280def2d72df7960b1324349b08dcb186ca1L458) was removed from upstream, which makes this option obsolete for most of the test. I tried to just pass empty params, but it wasn't used in the test. so lint throws a warning.

jwendell:
https://github.com/maistra/test-infra/pull/172

tools/istio-iptables/pkg/cmd/run_test.go
shamsher31: I think it has something to do with the output of the following?
https://github.com/maistra/istio/blob/9c6658fd6b16846d6cff4a525585a1dcb949f2a6/tools/istio-iptables/pkg/cmd/run_test.go#L58

jwendell: This was likely added by https://github.com/maistra/istio/pull/242
@luksa Any input here?

rcernich: I believe that was added to keep it from failing if it was trying to update iptables at the same time as the platform.

jwendell: True. My question was more like, Has this changed upstream between 1.9.5 and 1.9.8? Because those tests are passing currently and failing with the rebase.

rcernich: Looking at these changes, I think the test may have changed.  I'd check the gid and uid that are being set.  Should be easy to see from the test failures.  The only thing we do differently is allow uid and gid to be dynamically specified, as opposed to hard-coded (1337 is the default).

luksa: We add the `-w`, but upstream does not.

pkg/servicemesh/federation/status/manager_test.go
rcernich: Seems like `if len(status.DiscoveryStatus.Inactive) == 0 || status.DiscoveryStatus.Inactive[0].Watch.LastEvent.IsZero()` is enough.  I don't remember what that is doing, but I suspect if this panics, that's just a test failure, so all we're doing by cleaning things up is making sure the test doesn't crash?

jwendell (dismissed):

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -433,12 +457,6 @@ func buildGatewayListenerTLSContext(
 	ctx := &tls.DownstreamTlsContext{
 		CommonTlsContext: &tls.CommonTlsContext{
 			AlpnProtocols: util.ALPNHttp,
-			TlsParams: &tls.TlsParameters{
-				TlsMinimumProtocolVersion: convertTLSProtocol(server.Tls.MinProtocolVersion, tls_features.TLSMinProtocolVersion.Get()),
```
dgn: this is removing an OSSM feature (setting TLS defaults through env vars), we need this

dgn:
if we want to use merge strategy here, let's keep the required changes on top of the upstream commits to a single commit - we currently have 6 additional commits in here

shamsher31:
> if we want to use merge strategy here, let's keep the required changes on top of the upstream commits to a single commit - we currently have 6 additional commits in here

Thanks, I have squashed them into a single commit.

pilot/pkg/networking/core/v1alpha3/gateway.go
shamsher31: Reverted, Thanks.

pilot/pkg/networking/core/v1alpha3/gateway.go
```diff
@@ -514,38 +513,20 @@ func buildGatewayListenerTLSContext(
...
-	if len(server.Tls.CipherSuites) > 0 ||
-		server.Tls.MinProtocolVersion != networking.ServerTLSSettings_TLS_AUTO ||
-		server.Tls.MaxProtocolVersion != networking.ServerTLSSettings_TLS_AUTO {
-
-		ctx.CommonTlsContext.TlsParams = &tls.TlsParameters{
-			TlsMinimumProtocolVersion: convertTLSProtocol(server.Tls.MinProtocolVersion),
-			TlsMaximumProtocolVersion: convertTLSProtocol(server.Tls.MaxProtocolVersion),
-			CipherSuites:              filteredCipherSuites(server.Tls.CipherSuites),
-		}
+	if len(server.Tls.CipherSuites) > 0 {
```
shamsher31, line 517: This was added in https://github.com/maistra/istio/commit/443879e3cb83b724b6cbb0ab582b8bda13742e94#diff-7abd8e9bc7226623690e85b5acc0280def2d72df7960b1324349b08dcb186ca1R492

shamsher31:
Integration tests are failing due to timeout. It's almost taking more than 1hr to run the integration test.
```
traffic.go:150: call failed from TestRunner to https://172.18.255.155:443 (using https): 
timeout while waiting (last error: no responses received)
```

dgn (approved):

PR #432, opened 2021-09-01
MAISTRA-2439: Prevent federation from exporting services that are not visible to the federation gateway

jwendell:
By taking into consideration the service annotation
`networking.istio.io/exportTo`.

This annotation restricts where this service is visible: https://istio.io/latest/docs/reference/config/annotations/

If a service is not reachable from the federation gateway namespace due
to this annotation, it should not be exported.

### DIFF
--- a/pkg/servicemesh/federation/server/server.go
+++ b/pkg/servicemesh/federation/server/server.go
@@ -33,6 +33,7 @@ import (
 
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config/host"
+	"istio.io/istio/pkg/config/visibility"
 	"istio.io/istio/pkg/servicemesh/federation/common"
 	federationmodel "istio.io/istio/pkg/servicemesh/federation/model"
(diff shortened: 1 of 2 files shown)


dgn (approved):
thanks Jonh, looks good. Should we create an OSSMDOCS issue to make sure this is documented?

jwendell:
@dgn I'm not sure. Talking to @rcernich he said this would be the expected behavior. I mean, it's expected that users would understand this service would not be exported once they add such annotations.

That said, I think documentation is always good to have, even if "in obvious cases", right?

rcernich (approved):

dgn:
> @dgn I'm not sure. Talking to @rcernich he said this would be the expected behavior. I mean, it's expected that users would understand this service would not be exported once they add such annotations.
> 
> That said, I think documentation is always good to have, even if "in obvious cases", right?

+1 I think it doesn't hurt creating the ticket, albeit with maybe lower Priority. If we don't get to writing that part, so be it

PR #434, opened 2021-09-08
MAISTRA-2649: Update the config map when it already exists

jwendell:
If the creation fails with an `Already Exists` error, that might be
because the lister needs a sync and contains outdated information.

In such cases, try updating it instead. The update logic is smart enough
to only update it only when necessary - only if anything changed.

### DIFF
--- a/security/pkg/k8s/configutil.go
+++ b/security/pkg/k8s/configutil.go
@@ -51,7 +51,8 @@ func InsertDataToConfigMap(client corev1.ConfigMapsGetter, lister listerv1.Confi
 			// The lister may be outdated, thinking there's no config map when there is, so the attempt to create one will fail
 			// If this is the case, try updating it instead
 			if errors.IsAlreadyExists(err) {
-				return UpdateDataInConfigMap(client, configmap, data)
+				_, err := client.ConfigMaps(meta.Namespace).Update(context.TODO(), configmap, metav1.UpdateOptions{})
+				return err
 			}
 
 			return fmt.Errorf("error when creating configmap %v: %v", meta.Name, err)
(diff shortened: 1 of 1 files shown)


rcernich (approved):

dgn (approved):
I think this is a good solution for now, but if we have time to revisit in the future we should look at replacing it with waiting for the cache to sync by doing a `cache.WaitForCacheSync()` call to `informer.HasSynced()`, would spare us the extra call to the API server

luksa:
I don't think this will work. After the create operation fails, the update will probably fail with 409 Conflict, since the object passed into the update does not match the one in the API server. Let me check if that's the case.

luksa:
OK, looks like the update would succeed. Maybe if `resourceVersion` isn't set at all, the API server ignores it and simply updates the object without checking for conflicts.  

However, the update is never actually invoked because the `insertData()` function always returns false, since the configMap's data field matches the `data` map passed into it (see [1]). 

Instead of calling `UpdateDataInConfigMap()`, you should directly call `client.ConfigMaps().Update()`.

[1] https://github.com/Maistra/istio/blob/475b8e044fba92c0eed213aa1110cae188b9157c/security/pkg/k8s/configutil.go#L43

jwendell:
Thanks, @luksa . Updated the code.

security/pkg/k8s/configutil.go
```diff
@@ -51,7 +51,8 @@ func InsertDataToConfigMap(client corev1.ConfigMapsGetter, lister listerv1.Confi
 			// The lister may be outdated, thinking there's no config map when there is, so the attempt to create one will fail
 			// If this is the case, try updating it instead
 			if errors.IsAlreadyExists(err) {
-				return UpdateDataInConfigMap(client, configmap, data)
+				_, err := client.ConfigMaps(meta.Namespace).Update(context.TODO(), configmap, metav1.UpdateOptions{})
+				return err
```
luksa: It would probably be good to explain the error, like it's explained elsewhere in this function (e.g. `return fmt.Errorf("error when updating configmap %v: %v", meta.Name, err)`) to keep things consistent.

jwendell: done

luksa (approved):

PR #436, opened 2021-09-09
MAISTRA-2655: Fix the DNS capture with CNI

jwendell:
DNS capture is not correct when using CNI. It's capturing packets
with the destination of the server found in /etc/resolv.conf.

Since with CNI the iptables binary runs on the node, it's getting
the contents of resolv.conf from the node, not from the container.

Upstream fixed this with the addition of an istio-iptables CLI argument,
(`--capture-all-dns`) that is meant to capture all DNS traffic,
not only those targeting the DNS server listed in /etc/resolv.conf.

This is the full upstream PR: https://github.com/istio/istio/issues/29511

Since the code diverges a lot from our version and upstream,
this PR does a minimal manual backport of the upstream fix.

### DIFF
--- a/cni/cmd/istio-cni/iptables.go
+++ b/cni/cmd/istio-cni/iptables.go
@@ -56,7 +56,7 @@ func (ipt *iptables) Program(netns string, rdrct *Redirect) error {
 		"-k", rdrct.kubevirtInterfaces,
 	}
 	if rdrct.redirectDNS {
-		nsenterArgs = append(nsenterArgs, "--redirect-dns")
+		nsenterArgs = append(nsenterArgs, "--redirect-dns", "--capture-all-dns")
 	}
 	log.Infof("nsenter args: %s", strings.Join(nsenterArgs, " "))
 	out, err := exec.Command("nsenter", nsenterArgs...).CombinedOutput()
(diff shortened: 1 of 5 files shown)


jwendell:
Here's it in action (note the destination field):
```
# nsenter -n -t 1180038 iptables -t nat -vnL OUTPUT
Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
    7   420 ISTIO_OUTPUT  tcp  --  *      *       0.0.0.0/0            0.0.0.0/0           
   74  7288 RETURN     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp dpt:53 owner UID match 1000740001
    0     0 RETURN     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp dpt:53 owner GID match 1000740001
    8   828 REDIRECT   udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp dpt:53 redir ports 15053
```

The DNS requests are reaching the istio-agent, as seen in its log (in debug mode).

luksa (approved):

dgn (approved):
Awesome find, Jonh!

PR #439, opened 2021-09-22
MAISTRA-2682 Fix watch mechanism in federation

luksa:
Previously, no events were read from the watch response, because the read started with an endless loop that waited for data to be available in the decoder's buffer. This never happened, because the buffer is only written to when you call decoder.Decode(); this function was never called because the code waited for the buffer to have data.

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -15,9 +15,10 @@
 package federation
 
 import (
-	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"reflect"
@@ -92,7 +93,7 @@ type Controller struct {
 	lastMessage *federationmodel.ServiceListMessage
 	started     int32
 
-	watchDone   chan struct{}
+	cancelWatch context.CancelFunc
 	watchEvents chan *federationmodel.WatchEvent
 
 	peerConfigGeneration   int64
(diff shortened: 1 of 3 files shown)


pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -1020,27 +1033,14 @@ func (c *Controller) watch(eventCh chan *federationmodel.WatchEvent, stopCh <-ch
 	dec := json.NewDecoder(resp.Body)
 	defer resp.Body.Close()
 	for {
-		reader := dec.Buffered().(*bytes.Reader)
```
dgn, line 1023: whoops. I added this part in my last PR to make sure not to leave any goroutines around when stopping a watch. I have tested it extensively, though and I'm surprised you say it will never work

dgn: maybe I was testing an old image?

luksa: Hm, yeah, using an old image sounds like the most likely explanation. 

If you check the decoder code in golang/src/encoding/json/stream.go, you'll see that the buffer is only written to in the `Decode()` function.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -994,7 +1007,7 @@ func (c *Controller) watch(eventCh chan *federationmodel.WatchEvent, stopCh <-ch
 	}
 	req.Header.Add("discovery-service", c.discoveryServiceName)
 	req.Header.Add("remote", fmt.Sprint(common.RemoteChecksum(c.remote)))
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := http.DefaultClient.Do(req.WithContext(ctx))
```
dgn, line 1010: does this ensure we'll stop the http client when the context is cancelled?

dgn (approved):
Thank you, I'm glad you caught this. We should really use this client in the federation discovery server tests; I think we should factor out the code that does the watch and the polling into a Client and then use that in the unit tests to gain some more coverage and avoid things like this. I created [MAISTRA-2684](https://issues.redhat.com/browse/MAISTRA-2684) to track this effort.

pilot/pkg/serviceregistry/federation/controller.go
luksa: Yes. When the context is cancelled, the blocking read is aborted and the decoder returns the `io.EOF` error, which is handled below.

luksa:
Yeah, that's exactly what I'm doing today. :)

pilot/pkg/serviceregistry/federation/controller.go
dgn: perfect, thanks

rcernich (approved):

PR #443, opened 2021-10-11
Bare metal z -- provide example install scripts for bare metal and related federation installations

cfilleke:
provide example install scripts for bare metal and related federation installations

[ x] Docs
[ x] Installation
[ x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

[ X] Does not have any changes that may affect Istio users.


samples/bookinfo/platform/kube/bookinfo-db.yaml
```diff
@@ -49,7 +49,7 @@ spec:
     spec:
       containers:
       - name: mongodb
-        image: quay.io/maistra/examples-bookinfo-mongodb:2.1.0
+        image: quay.io/maistra/examples-bookinfo-mongodb:2.1-z
```
jwendell: This and below changes don't look right.

cfilleke: These are for the IBM system Z.  If we want to have a generic set of bookinfo images for Z, P and x86, we should make proper multi-arch manifests in quay.io/maistra.  Then 2.1.0 would be correct. As it is now, for the bookinfo yaml files to reference the 2.1-z images is just as correct as 2.1.0, since both need to be modified for one platform or another.

danshaik (approved):

jwendell:
@cfilleke You can try it local before pushing: `make lint`

samples/bookinfo/platform/kube/bookinfo-db.yaml
dgn: making this change would break everyone not on Z though, right? if so, we can't make that change in mainline

dgn (commented):
can we move this out of `example/config-poc/` and into its own subdirectory, e.g. `example/bare-metal/`? that way we'd have a clear separation between any special config needed for bare-metal/libvirt deployments and the simpler-to-configure cloud/single-cluster environments

cfilleke:
Closing: IBM System Z/Linux ONE and System P users, please refer to https://github.com/IBM/OSSM-federation-zap-config for latest updates relevant to libvirt IPI, z/VM UPI and hybrid istio installation and federation procedures.

PR #444, opened 2021-10-14
MAISTRA-2704 Print version when istiod starts

luksa:


pilot/cmd/pilot-discovery/main.go
```diff
@@ -62,6 +62,7 @@ var (
 			return nil
 		},
 		RunE: func(c *cobra.Command, args []string) error {
+			log.Info(version.Info.LongForm())
```
dgn, line 65: nit: maybe prefix with `istiod `

dgn (approved):
approving as there's a hold, find to merge this when 2.1 is out of code freeze

dgn:
@luksa I think this would be great to also push upstream. Can't imagine that anybody would not agree with this change

jwendell (approved):

PR #447, opened 2021-11-12
MAISTRA-2640 Add federation integration test

dgn:
This is a first `servicemesh` integration test. It covers setting up a federation that spans two clusters, exporting a service in one with an alias (ratings service), importing it in the second cluster, and then calling it from a `sleep` pod that lives in that second, importing cluster.

Depends on https://github.com/maistra/api/pull/26 which provides the CRD manifests through the go module system. That'll allow us to write more tests that require any of our CRDs (e.g. ServiceMeshExtension).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -118,7 +118,7 @@ require (
 	k8s.io/kube-openapi v0.0.0-20210312000640-9f480bcdc475
 	k8s.io/kubectl v0.20.1
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
-	maistra.io/api v0.0.0-20210823191737-259bd5854df6
+	maistra.io/api v0.0.0-20211119171546-348bbce3ca27
 	sigs.k8s.io/controller-runtime v0.7.0
 	sigs.k8s.io/service-apis v0.1.0
 	sigs.k8s.io/yaml v1.2.0
--- a/tests/integration/servicemesh/federation/federation_test.go
+++ b/tests/integration/servicemesh/federation/federation_test.go
@@ -1,0 +1,62 @@
+// +build integ
+// Copyright Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package servicemesh
+
+import (
+	"testing"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/istio"
+	"istio.io/istio/pkg/test/framework/label"
+	"istio.io/istio/tests/integration/servicemesh"
+)
+
+var (
+	i istio.Instance
+)
+
+// GetIstioInstance gets Istio instance.
+func GetIstioInstance() *istio.Instance {
+	return &i
+}
+
+func TestMain(m *testing.M) {
+	framework.
+		NewSuite(m).
+		RequireMinClusters(2).
+		Setup(servicemesh.ApplyServiceMeshCRDs).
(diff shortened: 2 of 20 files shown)


dgn:
I'm also updating `istio.deps` with the current proxy SHA in here now, because the old SHA that we shipped with 2.1.0 has since been deleted from our GCS bucket :/

jwendell:
> I'm also updating `istio.deps` with the current proxy SHA in here now, because the old SHA that we shipped with 2.1.0 has since been deleted from our GCS bucket :/

We should be able to manually upload it. Do you know why it got deleted?

dgn:
> > I'm also updating `istio.deps` with the current proxy SHA in here now, because the old SHA that we shipped with 2.1.0 has since been deleted from our GCS bucket :/
> 
> We should be able to manually upload it. Do you know why it got deleted?

everything's deleted after 60 days I believe

dgn:
Removing WIP but adding hold until https://github.com/maistra/api/pull/26 has been merged

tests/integration/servicemesh/util.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	for _, file := range files {
+		if !file.IsDir() && strings.HasSuffix(file.Name(), ".yaml") {
+			list = append(list, file.Name())
+		}
+	}
+	return
+}
+
+func InstallBookinfo(ctx framework.TestContext, c cluster.Cluster, namespace string) {
```
jwendell, line 77: perhaps you could leverage `pkg/test/framework/components/bookinfo`'s Deploy()'s? I see it hardcodes the cluster, but that could be easily patched to be given as part of the Config struct (and submitted upstream). It returns an uninstall() func which could be handy.

Same with sleep? (Add it and submit upstream if it's not there). Anyway, this could be done later in a follow up.

tests/integration/servicemesh/util.go
```diff
@@ -0,0 +1,123 @@
...
+		if _, err := kubetest.CheckPodsAreReady(kubetest.NewPodFetch(c, namespace, "app=sleep")); err != nil {
+			return fmt.Errorf("sleep pod is not ready: %v", err)
+		}
+		return nil
+	}, retry.Timeout(300*time.Second), retry.Delay(time.Second)); err != nil {
+		ctx.Fatal(err)
+	}
+}
+
+// TODO for some reason namespace.NewOrFail() doesn't work so I'm doing this manually
```
jwendell, line 105: weird, no clue at all? no logs? worth to create a ticket and add it to OSSM-401?

tests/integration/servicemesh/federation/federation_test.go
```diff
@@ -0,0 +1,62 @@
...
+// GetIstioInstance gets Istio instance.
+func GetIstioInstance() *istio.Instance {
+	return &i
+}
+
+func TestMain(m *testing.M) {
+	framework.
+		NewSuite(m).
+		RequireMinClusters(2).
+		Setup(servicemesh.ApplyServiceMeshCRDs).
```
jwendell, line 40: This is great :)

jwendell (approved):
Thanks for this Daniel

go.mod
```diff
@@ -125,3 +125,5 @@ require (
 )
 
 replace github.com/envoyproxy/go-control-plane => github.com/envoyproxy/go-control-plane v0.9.9
+
+replace maistra.io/api => github.com/dgn/maistra-api v0.0.0-20210903091813-ce11be2f0b74
```
jwendell: You can now remove this

tests/integration/servicemesh/util.go
dgn: it always threw `namespace already exists`. no idea why

tests/integration/servicemesh/util.go
dgn: +1 good suggestions. I'll look at it, but in a follow-up

tests/integration/servicemesh/util.go
luksa: `namespace.NewOrFail()` seems to loop through clusters (see [1]) and creates the namespace in each of them. Maybe the same cluster appears twice in the clusters list. If so, this could also be causing other issues. 

[1] https://github.com/Maistra/istio/blob/c45e902630dcb10138533a3ee255dd844c290ab0/pkg/test/framework/components/namespace/kube.go#L159

dgn: that's right, but I'm not setting up the list of clusters, that is coming from upstream

PR #455, opened 2021-11-25
Disable failing integration tests (until OSSM-755 is fixed)

dgn:
This is a stop-gap measure until OSSM-755 is fixed.

### DIFF
--- a/tests/integration/security/file_mounted_certs/p2p_mtls_test.go
+++ b/tests/integration/security/file_mounted_certs/p2p_mtls_test.go
@@ -46,6 +46,8 @@ const (
 )
 
 func TestClientToServiceTls(t *testing.T) {
+	// https://issues.redhat.com/browse/OSSM-755
+	t.Skip("https://github.com/istio/istio/issues/0")
 	framework.NewTest(t).
 		Features("security.peer.file-mounted-certs").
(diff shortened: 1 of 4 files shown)


tests/integration/security/file_mounted_certs/p2p_mtls_test.go
```diff
@@ -46,6 +46,8 @@ const (
 )
 
 func TestClientToServiceTls(t *testing.T) {
+	// https://issues.redhat.com/browse/OSSM-755
+	t.Skip("https://github.com/istio/istio/issues/0")
```
jwendell, line 50: We should fix this in upstream at some point. A regex should be ok, instead of hardcoding a url in the linter.

jwendell (approved):

tests/integration/security/file_mounted_certs/p2p_mtls_test.go
dgn: they do have a regex, but it only matches `github.com/istio/istio/issues/[0-9]*`. it's a tool that is compiled into their build image. we could clone and patch it, but I figured it's not really worth the effort

rcernich (approved):

PR #459, opened 2021-12-17
Remove redundant function ApplyServiceMeshCRDs

jewertow:
Function `ApplyServiceMeshCRDs` was added in #447, but it's redundant, because Maistra CRDs already exist in [manifests directory](https://github.com/maistra/istio/tree/maistra-2.1/manifests/charts/base/crds).

What's more, this function has no effect, because `os.Stat` returns `no such file or directory` for all files in `${IstioSrc}/vendor/maistra.io/api/manifests` and `kubectl apply` skips these files.

### DIFF
--- a/tests/integration/servicemesh/federation/federation_test.go
+++ b/tests/integration/servicemesh/federation/federation_test.go
@@ -1,4 +1,6 @@
+//go:build integ
 // +build integ
+
 // Copyright Red Hat, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


dgn:
I'm afraid we'll need to fix this function instead of remove it.

jewertow:
Why to keep this function, since it tries to apply already existing CRDs?

jewertow:
If you mean that I will need this function in #458, I didn't fix this function, because I saw this [comment](https://github.com/maistra/istio/blob/maistra-2.1/bin/update_maistra_deps.sh#L33), so I assumed I can copy required CRDs.

dgn:
Let's go with this one instead: https://github.com/maistra/istio/pull/461

dgn:
> Why to keep this function, since it tries to apply already existing CRDs?

Not all CRDs are in the manifests directory, and for a reason (some are installed by the operator, some by OLM - the software that installs the operator)

PR #461, opened 2021-12-17
OSSM-870 Fix ApplyServiceMeshCRDs function

dgn:


bin/update_maistra_deps.sh
```diff
@@ -23,25 +23,16 @@ go get -d "maistra.io/api@${UPDATE_BRANCH}"
...
 go mod vendor
 
-# Copy CRD's
-dir=$(mktemp -d)
-git clone "https://github.com/maistra/api.git" "${dir}"
-pushd "${dir}"
-git checkout "${UPDATE_BRANCH}"
-popd
-
 # FIXME: https://issues.redhat.com/browse/MAISTRA-2353
```
jewertow, line 26: What about this issue? Are we going to copy all of maistra CRDs?

dgn: no - as I commented on the issue, I don't think it makes sense to do that. we should discuss in the team what we want to do and then remove the FIXME when we address the issue.

jewertow (approved):

PR #462, opened 2022-01-14
OSSM-542 Use buffered channel to avoid cert watcher deadlock

dgn:
When we disabled webhook config management in Galley, we introduced a deadlock because we never started the goroutine that read from the unbuffered webhookServerReady channel. This meant that the code that reacted to watch events was never reached, because the goroutine was stuck sending to that channel.

The fix is pretty easy: just make the channel a buffered channel, and now the write is non-blocking.


dgn:
I had to fix a few things to get the 1.1 branch to build. I kept those changes in separate commits

bin/init_helm.sh
```diff
@@ -75,13 +75,5 @@ export ISTIO_OUT=${ISTIO_OUT:-${ISTIO_BIN}}
...
 if [ ! -f "${ISTIO_OUT}/version.helm.${HELM_VER}" ] ; then
-    TD=$(mktemp -d)
-    # Install helm. Please keep it in sync with .circleci
-    cd "${TD}" && \
-        ${DOWNLOAD_COMMAND} "${TD}/helm.tgz" "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VER}-${LOCAL_OS}-amd64.tar.gz" && \
-        tar xfz helm.tgz && \
-        mv ${LOCAL_OS}-amd64/helm "${ISTIO_OUT}/helm-${HELM_VER}" && \
-        cp "${ISTIO_OUT}/helm-${HELM_VER}" "${ISTIO_OUT}/helm" && \
-        rm -rf "${TD}" && \
-        touch "${ISTIO_OUT}/version.helm.${HELM_VER}"
```
jewertow, line 86: Why did you remove downloading helm?

jewertow (approved):

bin/init_helm.sh
dgn: the download failed- it doesn't matter anyway, because we pre-install helm in our builder image

jwendell (approved):

PR #470, opened 2022-02-28
Fix failing build of proxyv2 image

jewertow:


pilot/docker/Dockerfile.proxyv2
```diff
@@ -1,5 +1,11 @@
 FROM centos:8
```
jwendell: We should move this to `FROM quay.io/centos/centos:stream8`

PR #489, opened 2022-03-23
OSSM-1254: Use maistra proxy

jwendell:


oschaaf:
(@jwendell fyi -- I added 8ee9211 to test with an updated proxy sha)

oschaaf:
(Updated the proxy sha again)

oschaaf:
I'm surprised that the integration tests still fail. I double checked, retracing my steps with respect to the changes to proxy that the latest sha update bring in -- my local test run succeeds with these changes. But perhaps I am missing something ... looking into if that's the case.

oschaaf:
Looking at the CI flow, I now wonder -- was I too fast with updating the proxy sha? Maybe the new proxy binary wasn't pushed to its gcp bucket yet, and we ran with the old version. Lets try again:

/retest

jwendell:
> Looking at the CI flow, I now wonder -- was I too fast with updating the proxy sha? Maybe the new proxy binary wasn't pushed to its gcp bucket yet, and we ran with the old version. Lets try again:

Nope. If the new binary wasn't there, the test would fail quickly with a 404 error. It wouldn't run with the old version. Also note that the other integration test (istio-integration-maistra-mc-2.2) succeeded. The rule of the thumb is: When automator creates a PR updating istio.deps, this means that the new build was uploaded to GCS.

oschaaf:
So I also had `s/--config=release/-c dbg/` in `common.sh` for the proxy build; and I think I just proven that reverting this change brings back the issue in local test runs. So stripping the hardening flags only helps for debug builds -- release builds will still fail :(

rcernich (approved):

PR #504, opened 2022-04-14
OSSM-1279 Set UID/GID to 1337 in proxyv2 image

luksa:


jewertow (dismissed):

jewertow:
These tests work on my computer, so they should pass...

jewertow:
@luksa did you investigate why do these tests fail?

dgn:
Failures seem to be legit. From the proxy logs:

```
2022-04-21T11:53:20.799278Z	error	envoy config	listener '0.0.0.0_80' failed to bind or apply socket options: cannot bind '0.0.0.0:80': Permission denied
2022-04-21T11:53:20.800080Z	warning	envoy config	gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) 0.0.0.0_80: cannot bind '0.0.0.0:80': Permission denied
```

dgn:
It looks like the failing tests use custom gateway Services. You should be able to update the Services in the two failing tests to include a `targetPort > 1023`.

luksa:
Yeah, I fixed the tests by using a different port, but that made me think that changing the default UID may not be appropriate, since it would break existing deployments. And if this change isn't accepted upstream, we probably don't want our proxyv2 containers to use a different default UID, right?

luksa:
Closing. @dgn agrees that changing the UID in the image might not be acceptable in some cases.

PR #509, opened 2022-04-27
OSSM-1423 Preliminary support for Gateway API

dgn:
This adds support for Gateway API in a multi-tenant OSSM install, plus integration tests. It also improves upon our current test suite a bit. Note that it will still be disabled by default in OSSM- but enabling it will not break istiod anymore.

Caveats:
- we don't support namespace selectors in the AllowedRoutes field
- GatewayClasses are ignored (but specifying "istio" always works)

### DIFF
--- a/pilot/pkg/config/kube/crdclient/client.go
+++ b/pilot/pkg/config/kube/crdclient/client.go
@@ -501,6 +501,11 @@ func handleCRDAdd(cl *Client, name string, stop <-chan struct{}) {
 	resourceGVK := s.Resource().GroupVersionKind()
 	gvr := s.Resource().GroupVersionResource()
 
+	if cl.client.GetMemberRoll() != nil && resourceGVK == gvk.GatewayClass {
+		scope.Infof("Skipping CRD %v as it is not compatible with maistra multi-tenancy", s.Resource().GroupVersionKind())
+		return
+	}
+
 	cl.kindsMu.Lock()
 	defer cl.kindsMu.Unlock()
 	if _, f := cl.kinds[resourceGVK]; f {
--- a/pilot/pkg/config/kube/gateway/controller.go
+++ b/pilot/pkg/config/kube/gateway/controller.go
@@ -99,11 +99,16 @@ func NewController(client kube.Client, c model.ConfigStoreCache, options control
 			}
 		}, uint(features.StatusMaxWorkers))
 	}
-	nsInformer := client.KubeInformer().Core().V1().Namespaces().Informer()
+	var nsLister listerv1.NamespaceLister
+	var nsInformer cache.SharedIndexInformer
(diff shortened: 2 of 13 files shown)


jwendell (commented):
Where it's supposed to be enabled/disabled?

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -498,6 +498,11 @@ func handleCRDAdd(cl *Client, name string, stop <-chan struct{}) {
 	resourceGVK := s.Resource().GroupVersionKind()
 	gvr := s.Resource().GroupVersionResource()
 
+	if cl.client.GetMemberRoll() != nil && resourceGVK == gvk.GatewayClass {
+		scope.Infof("Skipping CRD %v as it is not compatible with maistra multi-tenancy", s.Resource().GroupVersionKind())
```
rcernich, line 505: I assume this is because GatewayClass is a cluster scoped resource?

dgn:
> Where it's supposed to be enabled/disabled?

We disable it by the default [in the operator](https://github.com/maistra/istio-operator/blob/78ef39ec9034ff0139f4d04e51152ae0e36d6280/build/patch-charts.sh#L169-L173). Before, in 2.1, it was still [disabled by default](https://github.com/maistra/istio/blob/e8478c0696a8c962d4ec3a991463b4289383f09a/pilot/pkg/features/pilot.go#L297-L299) upstream (it was still called Service API back then).

pilot/pkg/config/kube/gateway/controller.go
```diff
@@ -99,26 +99,33 @@ func NewController(client kube.Client, c model.ConfigStoreCache, options control
...
 
-	nsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
-			gatewayController.namespaceEvent(nil, obj)
-		},
-		UpdateFunc: func(oldObj, newObj interface{}) {
-			gatewayController.namespaceEvent(oldObj, newObj)
-		},
-	})
+	if client.GetMemberRoll() == nil {
```
jwendell: nit: I'd prefer a comparison like `if nsInformer != nil` here and below, but it's not a strong preference.

jwendell (approved):

rcernich (approved):

pilot/pkg/config/kube/crdclient/client.go
dgn: that's right

pilot/pkg/config/kube/gateway/controller.go
dgn: makes sense. I made the changes

jwendell (approved):

PR #510, opened 2022-04-27
OSSM-1430: CNI: Watch for modified files with a prefix

jwendell:
Because our CNI pod contains more than one container, and they write to
the same directory, and they watch for changes on those directories,
changes made by one container trigger the watch on the other, which will
responde by copying the files to the directory, which will in turn
trigger the watcher of the other container in an endless loop.

This leads to high CPU usage on the node.

This PR changes the logic to only monitor for files that have the
desired prefix. Thus, for example, the 2.2 container will only react to
changes to files whose names  start with "v2-2". This avoid this race
condition and achieve the same end result.

### DIFF
--- a/cni/pkg/util/util.go
+++ b/cni/pkg/util/util.go
@@ -19,21 +19,30 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/fsnotify/fsnotify"
 
 	"istio.io/istio/pkg/file"
 )
 
+type watchPrefix struct {
+	watcher *fsnotify.Watcher
+	prefix  string
+}
+
 // Creates a file watcher that watches for any changes to the directory
-func CreateFileWatcher(dirs ...string) (watcher *fsnotify.Watcher, fileModified chan bool, errChan chan error, err error) {
+// If `prefix` is non-empty, it will only notify of changes for files whose filename starts with `prefix`
+func CreateFileWatcher(prefix string, dirs []string) (watcher *fsnotify.Watcher, fileModified chan bool, errChan chan error, err error) {
 	watcher, err = fsnotify.NewWatcher()
 	if err != nil {
 		return
 	}
 
 	fileModified, errChan = make(chan bool), make(chan error)
-	go watchFiles(watcher, fileModified, errChan)
+	wp := &watchPrefix{watcher: watcher, prefix: prefix}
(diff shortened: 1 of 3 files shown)


cni/pkg/util/util.go
```diff
@@ -19,21 +19,30 @@ import (
...
 	"istio.io/istio/pkg/file"
 )
 
+type watchPrefix struct {
+	watcher *fsnotify.Watcher
+	prefix  string
+}
+
 // Creates a file watcher that watches for any changes to the directory
-func CreateFileWatcher(dirs ...string) (watcher *fsnotify.Watcher, fileModified chan bool, errChan chan error, err error) {
```
dgn, line 29: why replace `...string` with `[]string`?

dgn (approved):
LGTM to unblock 2.2. For upstream, we'll have to submit a different PR though, as they don't have the `CNIBinariesPrefix` field in `InstallConfig`. We should file a follow-up ticket to do that, so we don't have to carry over this change indefinitely

luksa (approved):

cni/pkg/util/util.go
jwendell: because if we have two arguments, first of type `string` and second of type `...string`, go interprets it as a single one as of type `...string`, and thus accepts only one argument as a result. https://go.dev/play/p/HuASlAWYUl2

jwendell:
> LGTM to unblock 2.2. For upstream, we'll have to submit a different PR though, as they don't have the `CNIBinariesPrefix` field in `InstallConfig`. We should file a follow-up ticket to do that, so we don't have to carry over this change indefinitely

That's interesting. We already carry over this whole prefix thing. I was wondering if we should submit all these changes to upstream as an enhancement. I think it's a good addition even if they don't use it. The fact that we use and are willing to maintain it is a good argument. @luksa any thougths?

luksa:
I don't think upstream will want the prefix stuff, since they don't need it and it makes things quite more complex. But I'm not against pushing all this upstream.

dgn:
> I don't think upstream will want the prefix stuff, since they don't need it and it makes things quite more complex. But I'm not against pushing all this upstream.

+1, I'm skeptical they will merge unused/dead code - I wouldn't want it either. This fix however could be relevant, even though the issue technically only materializes when you have multiple CNI plugins "misbehaving" like this (ie copying everything into the folders on watch events even though nothing on the ground has actually changed)

PR #512, opened 2022-04-29
OSSM-1252 Fix federation status updates

luksa:


dgn (approved):
I just saw your comment on the JIRA. Thank you for the explanation. I think this approach makes sense if it ends up allowing us to re-enable the test. Should we create a follow-up issue for the remaining stale object issue? Or wait and see what the failure rate will be?

luksa:
Yeah, this PR doesn't just fix the flaky test, but also the faulty status patches. Previously, the test failed about 10% of the time on my machine, now it fails around 1%, which is probably in-line with how flaky the test was before my recent changes to it.

I've created OSSM-1443.

PR #513, opened 2022-04-29
OSSM-1420 Ensure istiod can run with no cluster-scoped privileges

luksa:


pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -91,14 +86,8 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
...
-	client.KubeInformer().InformerFor(&appsv1.Deployment{}, func(k kubernetes.Interface, resync time.Duration) cache.SharedIndexInformer {
-		return appsinformersv1.NewFilteredDeploymentInformer(
-			k, metav1.NamespaceAll, resync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
-			func(options *metav1.ListOptions) {
-				options.LabelSelector = "gateway.istio.io/managed=istio.io-gateway-controller"
-			},
-		)
-	}).AddEventHandler(handler)
+	// TODO: filter deployments with label selector "gateway.istio.io/managed=istio.io-gateway-controller"
+	client.KubeInformer().Apps().V1().Deployments().Informer().AddEventHandler(handler)
```
dgn, line 90: we could consider using [`NewFilteredSecretInformer()`](https://github.com/maistra/xns-informer/blob/d7b9519d1378b4fd2cbdfbde61b1fbd72649e786/pkg/generated/kube/core/v1/secret.go#L59-L83) func of xns-informer

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -81,23 +79,7 @@ type authorizationResponse struct {
...
-				// large Secrets in clusters.
-				// This is a best effort optimization only; the code would behave correctly if we watched all Secrets.
-				options.FieldSelector = fields.AndSelectors(
-					fields.OneTermNotEqualSelector("type", "helm.sh/release.v1"),
-					fields.OneTermNotEqualSelector("type", string(v1.SecretTypeServiceAccountToken)),
-				).String()
-			},
-		)
-	})
+	informer := client.KubeInformer().Core().V1().Secrets().Informer()
```
dgn, line 82: same here, we could use [NewFilteredSecretInformer()](https://github.com/maistra/xns-informer/blob/d7b9519d1378b4fd2cbdfbde61b1fbd72649e786/pkg/generated/kube/core/v1/secret.go#L59-L83). let's create a follow-up ticket

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
```diff
@@ -202,3 +204,51 @@ func (d *discoveryNamespacesFilter) isSelected(labels labels.Set) bool {
 
 	return false
 }
+
+type maistraDiscoveryNamespacesFilter struct {
+	lock       sync.RWMutex
+	namespaces sets.String
+}
+
+func NewMaistraDiscoveryNamespacesFilter(mrc memberroll.MemberRollController) DiscoveryNamespacesFilter {
```
dgn, line 212: matches exactly what I have :) just called it `FakeDiscoveryNamespacesFilter`

dgn (approved):
lgtm. not 100% sure about the impact of leaving out the filters on the informers though. we should file a follow-up ticket

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -91,14 +86,8 @@ func NewDeploymentController(client kube.Client) *DeploymentController {
...
 	// For Deployments, this is the only controller watching. We can filter to just the deployments we care about
-	client.KubeInformer().InformerFor(&appsv1.Deployment{}, func(k kubernetes.Interface, resync time.Duration) cache.SharedIndexInformer {
-		return appsinformersv1.NewFilteredDeploymentInformer(
-			k, metav1.NamespaceAll, resync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
-			func(options *metav1.ListOptions) {
-				options.LabelSelector = "gateway.istio.io/managed=istio.io-gateway-controller"
-			},
-		)
-	}).AddEventHandler(handler)
+	// TODO: filter deployments with label selector "gateway.istio.io/managed=istio.io-gateway-controller"
```
jwendell, line 89: Worth to create a Jira and link it here? We might thinks this is a TODO from upstream and never look back here

dgn:
integration test failure is likely a flake. seen this before

rcernich:
Looks like pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go needs formatting.

pilot/pkg/credentials/kube/secrets.go
luksa: Here, the filtering is only done as a best-effort optimization (according to the comment, the code behaves correctly without it). And since we're only watching the secrets in member namespaces, whereas the upstream code uses a cluster-scoped watch, this optimization may not do much at all.

pilot/pkg/config/kube/gateway/deploymentcontroller.go
luksa: AFAICT, the NewFilteredSecretInformer is a more low-level function. If we were to use it, we'd have to keep the list of namespaces in sync using the SMMR controller. 

On the other hand, this filter here is only used for optimization.

pilot/pkg/config/kube/gateway/deploymentcontroller.go
luksa: I've created OSSM-1448, but I have the feeling that this is something that we'll never actually implement. We'll either move away from xns-informers or upstream will remove the label selector.

dgn:
@luksa is this ready to merge? I would like to update https://github.com/maistra/istio/pull/509 once it is in

PR #516, opened 2022-05-02
OSSM-1442: IOR: Ignore UPDATE events if resourceVersions are the same

jwendell:
For some obscure reason, it looks like we may receive UPDATE events with
the new object being equal to the old one. As IOR always delete and
recreate routes when receiving an UPDATE event, this might lead to some
service downtime, given for a few moments the route will not exist.

We guard against this behavior by comparing the `resourceVersion` field
of the new object and the one stored in the Route object.

### DIFF
--- a/pilot/pkg/config/kube/ior/ior_test.go
+++ b/pilot/pkg/config/kube/ior/ior_test.go
@@ -42,7 +42,8 @@ import (
 func initClients(t *testing.T,
 	stop <-chan struct{},
 	errorChannel chan error,
-	mrc memberroll.MemberRollController) (model.ConfigStoreCache, kube.Client, routev1.RouteV1Interface) {
+	mrc memberroll.MemberRollController,
+	register bool) (model.ConfigStoreCache, KubeClient, routev1.RouteV1Interface) {
 	t.Helper()
 
 	k8sClient := kube.NewFakeClient()
@@ -63,11 +64,13 @@ func initClients(t *testing.T,
 		return nil
 	}, retry.Timeout(time.Second))
 
-	if err := Register(iorKubeClient, routerClient, store, "istio-system", mrc, stop, errorChannel); err != nil {
-		t.Fatal(err)
+	if register {
+		if err := Register(iorKubeClient, routerClient, store, "istio-system", mrc, stop, errorChannel); err != nil {
+			t.Fatal(err)
+		}
 	}
 
-	return store, k8sClient, routerClient
+	return store, iorKubeClient, routerClient
 }
 
 func TestCreate(t *testing.T) {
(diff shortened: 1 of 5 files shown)


luksa:
> For some obscure reason, it looks like we may receive UPDATE events with the new object being equal to the old one.

Could this be caused by xns-informer? If so, we may have similar issues elsewhere.

jwendell:
I don't know. @rcernich suggested that this could possible indeed.

The problem with this PR is that we don't enforce the desired state upon an UPDATE event if the object's resourceVersion is the same as the one we have in memory. Quoting Rob:

> ... the reconciler should force the managed resource to match the desired state, so if we get an update request for a Gateway, we should still verify that the created Route reflects the desired state (e.g. the user hasn't modified the Route)

One possible fix to this is to add a second level comparison: If the objects are the same, perform an get() request on the OCP API just to make sure the object wasn't modified. This has the disadvantage of performing an API request, which could be too much at some point. A solution for this would be using a cache store, but this requires a bigger refactor of the IOR code.

Comments?

rcernich:
@luksa, are you saying that xns-informer is not generating the correct event types for resyncs?

dgn:
I spent some time looking at this and the error logs mentioned in the issue. I believe it might be caused by a reconnect in the reflector. But let me explain:
1. just before the UPDATE events come in, our leader election lock logs: `error retrieving resource lock servicemesh/ior-leader: etcdserver: leader changed` - it doesn't report that it lost leadership, though, so this is just the result of a periodic check if we're still the leader. one of those checks has failed because etcd leader changed.
2. Likely, our WATCH on Routes is also dropped as a result of that etcd leader change, forcing the reflector to resync
3. The reflector will re-list all resources from the last known resourceVersion, causing these UPDATE events

I'm not entirely sure if that's the cause, but if it is, then the method of comparing resourceVersion makes sense and is a viable fix. I will ask the customer for more debug logs.

What I still don't understand is that normal resyncs don't seem to cause this problem. In the customer logs, it only happens once over a period of three days. Maybe the new etcd leader has a lower resourceVersion than what the reflector is sending, and thus sending a `TooLargeResourceVersionError` [causing the reflector to re-list from scratch](https://github.com/kubernetes/client-go/blob/b4b8e5e1e565f72819da7105edd839bbdbba1386/tools/cache/reflector.go#L303-L304).

pilot/pkg/config/kube/ior/ior_test.go
```diff
@@ -363,6 +363,11 @@ func TestEdit(t *testing.T) {
 			validateRoutes(t, c.hosts, list, "gw", c.tls)
 		})
 	}
+
+	// Try to edit a gateway with the same resourceVersion
+	// No effect. UPDATE Event is not generated. I'll remove it.
+	IORLog.SetOutputLevel(log.DebugLevel)
+	editGateway(t, store, "istio-system", "gw", []string{"one.org.br"}, map[string]string{"istio": "ingressgateway"}, true, "6")
```
dgn: maybe we can replace this with a test that explicitly sends an event twice?

jwendell: done. It was not very straightforward but at least it's a start.

pilot/pkg/config/kube/ior/ior_test.go
```diff
@@ -459,6 +462,69 @@ func TestConcurrency(t *testing.T) {
...
+		t.Fatalf("expecting the error: %q, but got nothing", eventDuplicatedMessage)
+	}
+	if msg := err.Error(); msg != eventDuplicatedMessage {
+		t.Fatalf("expecting the error: %q, but got %q", eventDuplicatedMessage, msg)
+	}
+
+	func() {
+		r.gatewaysLock.Lock()
+		defer r.gatewaysLock.Unlock()
+		if len(r.gatewaysMap) != 1 {
```
dgn: shouldn't this test whether the OpenShift Route was re-created?

dgn: ah nevermind it checks the duplication error message above. I think this part can be left out then?

jwendell: Yeah, the check for the error above should be enough. Removed.

dgn (approved):

PR #519, opened 2022-05-09
WIP: OSSM-1211 Merge a locality setting from an imported service

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -136,6 +136,8 @@ func mergeLocality(locality *v1.ImportedServiceLocality, defaults *v1.ImportedSe
 		if locality == nil {
 			return nil
 		}
+		// When defaults is nil and locality is not nil
+		merged = *locality
 	} else if locality != nil {
 		merged = *locality
 	}


rcernich:
Is it possible to add a unit test that recreates the bug and verifies the fix?

yxun:
This change should help set a locality when the controller defaultLocality is empty. See the following line:
https://github.com/maistra/istio/blob/maistra-2.1.2/pilot/pkg/serviceregistry/federation/controller.go#L255

In our 2.1 code, we ignore the locality spec value from the importConfig when c.defaultLocality is nil. 
I think we should not ignore that.

rcernich:
Please add some unit tests.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -136,6 +136,8 @@ func mergeLocality(locality *v1.ImportedServiceLocality, defaults *v1.ImportedSe
 		if locality == nil {
 			return nil
 		}
+		// When defaults is nil and locality is not nil
+		merged = *locality
 	} else if locality != nil {
 		merged = *locality
 	}
```
jewertow, line 143: Couldn't this if-else condition be simplified to the following conditions?
```
if defaults == nil {
	defaults = &v1.ImportedServiceLocality{}
	if locality == nil {
		return nil
	}
}
if locality != nil {
	merged = *locality
}
```

yxun:
comment added in OSSM-1211. This PR is not related to the OSSM-1211 issue so close this one now.

PR #522, opened 2022-05-19
OSSM-1529 Improve federation example script

luksa:
Previously, the script would fall back to using nodeports when the load balancer IP wasn't set. This meant that if the provision of the load balancer took too long, the SMCPs would be misconfigured and you had to run the install script again.

With this change, the script now waits for the load balancer IP to appear. It never falls back to using node ports, because they never really worked (the nodes' hostnames typically aren't FQDN and the node ports are typically protected by firewalls).

If the user wants to expose the federation ingresses in a different way, they can now set the environment variables MESH1_ADDRESS, MESH1_DISCOVERY_PORT, and MESH2_SERVICE_PORT (likewise for MESH2) and run the script.


jwendell:
Do we need to change README to reflect these changes?

luksa:
That README is outdated actually. It was written for the initial version of the install.sh script. We can probably just remove it.

jwendell:
got it. I know it's a separate issue, but we should probably keep the README up to date. People are relying on this example to understand and even test federation.

luksa:
@jwendell I've updated the README.

pkg/servicemesh/federation/example/config-poc/install.sh
```diff
@@ -45,59 +45,53 @@ log "Retrieving root certificates"
...
+
+  while [ -z "$MESH2_ADDRESS" ]
+  do
+    MESH2_ADDRESS=$(oc2 -n mesh2-system get svc mesh1-ingress -o jsonpath="{.status.loadBalancer.ingress[].ip}")
     if [ -z "$MESH2_ADDRESS" ]; then
-      echo "LoadBalancer service mesh1-ingress has no externally reachable ip or hostname; falling back to node ports"
-      MESH2_ADDRESS=$(oc2 get nodes -o json | jq -r '.items[0].status.addresses[] | select (.type == "Hostname") | .address')
-      MESH2_DISCOVERY_PORT=$(oc2 -n mesh2-system get svc mesh1-ingress -o json | jq '.spec.ports[] | select (.name == "https-discovery") | .nodePort')
-      MESH2_SERVICE_PORT=$(oc2 -n mesh2-system get svc mesh1-ingress -o json | jq '.spec.ports[] | select (.name == "tls") | .nodePort')
+      MESH2_ADDRESS=$(oc2 -n mesh2-system get svc mesh1-ingress -o jsonpath="{.status.loadBalancer.ingress[].hostname}")
```
yxun, line 77: This change should be enough for fixing the OSSM-1529 issue and avoid using the nodeports. However,
@luksa This does not help resolve the LB creation check. I can get an AWS LB hostname by
MESH1_ADDRESS=$(oc1 -n mesh1-system get svc mesh2-ingress -o jsonpath="{.status.loadBalancer.ingress[].hostname}")

However, the actual AWS LB resource is still in Provisioning state. For example, the MESH1_ADDRESS is not empty but the LB Provisioning state takes another 10 minutes.
ServceMeshPeer connection will be false until an AWS LB resource state in Active

So your change and line 79 echo "Waiting for load balancer to be provisioned" will not wait the LB to be provisioned. It only checks when the LB started to be provisioned.

yxun (approved):

jwendell (approved):

PR #524, opened 2022-05-23
OSSM-993 Deprecate ServiceMeshExtension

yannuil:
Signed-off-by: Yann Liu <yannliu@redhat.com>

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/pkg/webhooks/validation/server/server.go
+++ b/pkg/webhooks/validation/server/server.go
@@ -210,6 +210,10 @@ func (wh *Webhook) validate(request *kube.AdmissionRequest) *kube.AdmissionRespo
 		return toAdmissionResponse(fmt.Errorf("cannot decode configuration: %v", err))
 	}
 
+	if request.Kind.Kind == "ServiceMeshExtension" {
+		return &kube.AdmissionResponse{Allowed: true, Warnings: []string{"ServiceMeshExtension is deprecated. Please use WasmPlugin instead."}}
+	}
+
 	gvk := obj.GroupVersionKind()
 
 	// TODO(jasonwzm) remove this when multi-version is supported. v1beta1 shares the same
--- a/pkg/webhooks/validation/server/server.go
+++ b/pkg/webhooks/validation/server/server.go
@@ -37,6 +37,8 @@ import (
 	"istio.io/pkg/log"
 )
 
+const smeDeprecationMessage = "ServiceMeshExtension is deprecated. It works now but will be removed in a future version. Please use WasmPlugin instead."
+
 var scope = log.RegisterScope("validationServer", "validation webhook server", 0)
 
 var (
@@ -211,7 +213,7 @@ func (wh *Webhook) validate(request *kube.AdmissionRequest) *kube.AdmissionRespo
 	}
 
 	if request.Kind.Kind == "ServiceMeshExtension" {
-		return &kube.AdmissionResponse{Allowed: true, Warnings: []string{"ServiceMeshExtension is deprecated. Please use WasmPlugin instead."}}
+		return &kube.AdmissionResponse{Allowed: true, Warnings: []string{smeDeprecationMessage}}
 	}
 
 	gvk := obj.GroupVersionKind()
--- a/pkg/webhooks/validation/server/server_test.go
+++ b/pkg/webhooks/validation/server/server_test.go
@@ -254,7 +254,6 @@ func TestAdmitPilot(t *testing.T) {
 
 	smeV1 := makeServiceMeshExtensionConfig(t, "v1")
 	smeV1alpha1 := makeServiceMeshExtensionConfig(t, "v1alpha1")
-	deprecatedWarning := "ServiceMeshExtension is deprecated. Please use WasmPlugin instead."
 
 	maistraCases := []struct {
 		name string
@@ -302,8 +301,8 @@ func TestAdmitPilot(t *testing.T) {
 				t.Fatal("should be allowed.")
 			}
 
-			if got.Warnings[0] != deprecatedWarning {
-				t.Fatalf("should get the deprecation warning \"%s\".", deprecatedWarning)
+			if got.Warnings[0] != smeDeprecationMessage {
+				t.Fatalf("should get the deprecation warning \"%s\".", smeDeprecationMessage)
 			}
 		})
 	}
--- a/pkg/webhooks/validation/server/server_test.go
+++ b/pkg/webhooks/validation/server/server_test.go
@@ -139,6 +139,34 @@ func makePilotConfig(t *testing.T, i int, validConfig bool, includeBogusKey bool
 	return raw
 }
 
+func makeServiceMeshExtensionConfig(t *testing.T, apiVersion string) []byte {
+	t.Helper()
+
+	var un unstructured.Unstructured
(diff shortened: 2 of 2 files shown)


jewertow:
What about tests? I took a look at [server_test.go](https://github.com/maistra/istio/blob/maistra-2.2/pkg/webhooks/validation/server/server_test.go#L220) and I guess that it should be quite simple to add a test case.

dgn:
> What about tests? I took a look at [server_test.go](https://github.com/maistra/istio/blob/maistra-2.2/pkg/webhooks/validation/server/server_test.go#L220) and I guess that it should be quite simple to add a test case.

Agreed. That test doesn't (yet) support testing for warnings. But we could extend it to do that

yannuil:
I will give a shot on extending it to build a test case for the deprecation warnings.

pkg/webhooks/validation/server/server_test.go
```diff
@@ -223,6 +251,52 @@ func TestAdmitPilot(t *testing.T) {
...
+			admit: wh.validate,
+			in: &kube.AdmissionRequest{
+				Kind:      kubeApisMeta.GroupVersionKind{Kind: "ServiceMeshExtension"},
+				Object:    runtime.RawExtension{Raw: smeV1alpha1},
+				Operation: kube.Create,
+			},
+			allowed: true,
+			warning: deprecatedWarning,
+		},
+	}
```
jewertow, line 294: 1. Operation `kube.Update` is not covered.
2. Both test cases have the same `admit`, `allowed` and `warning` fields, so the test case struct might be simplified to:
```
name       string
smeVersion []byte
operation  string
```
Then you would call in the test:
```
got := wh.validate(&kube.AdmissionRequest{
  Kind:      kubeApisMeta.GroupVersionKind{Kind: "ServiceMeshExtension"},
  Object:    runtime.RawExtension{Raw: c.smeVersion},
  Operation: operation,
})
```
It's not very important, but I just don't see a need to parameterize constant values.

pkg/webhooks/validation/server/server_test.go
```diff
@@ -223,6 +251,52 @@ func TestAdmitPilot(t *testing.T) {
...
+
+	for i, c := range maistraCases {
+		t.Run(fmt.Sprintf("[MAISTRA][%d] %s", i, c.name), func(t *testing.T) {
+			got := c.admit(c.in)
+			if got.Allowed != c.allowed {
+				t.Fatalf("got %v want %v", got.Allowed, c.allowed)
+			} else if got.Warnings[0] != c.warning {
+				t.Fatalf("did not get warning \"%s\"", c.warning)
+			}
```
jewertow, line 306: This `if-else` statement is absolutely correct, but since `else` is not important for the correctness of the test, couldn't it be simplified to two separate `if` statements? At the first glance it was confusing for me and I had to reflect if it matters or not. It's not important, so it's up to you.

yannuil: I have tried to improve it. The correct behavior is `allowed == true` as well as getting a warning message.

pkg/webhooks/validation/server/server_test.go
yannuil: Done

pkg/webhooks/validation/server/server_test.go
```diff
@@ -223,6 +251,57 @@ func TestAdmitPilot(t *testing.T) {
...
+				Object:    runtime.RawExtension{Raw: smeV1alpha1},
+				Operation: kube.Create,
+			},
+		},
+		{
+			name: "Update v1alpha1 ServiceMeshExtension",
+			in: &kube.AdmissionRequest{
+				Kind:      kubeApisMeta.GroupVersionKind{Kind: "ServiceMeshExtension"},
+				Object:    runtime.RawExtension{Raw: smeV1alpha1},
+				Operation: kube.Create,
```
jewertow: Operation should be `kube.Update` in this case.

pkg/webhooks/validation/server/server_test.go
jewertow: I meant two subsequent `if` statements:
```
if got.Allowed != true {
  t.Fatal(...)
}
if got.Warnings[0] != deprecatedWarning {
  t.Fatal(...)
}
```
Having a single assertion requires manual checking what's wrong once the test fails.

pkg/webhooks/validation/server/server_test.go
yannuil: Done

pkg/webhooks/validation/server/server_test.go
yannuil: Done

jewertow (approved):

pkg/webhooks/validation/server/server_test.go
```diff
@@ -223,6 +251,62 @@ func TestAdmitPilot(t *testing.T) {
...
+				Operation: kube.Update,
+			},
+		},
+	}
+
+	for i, c := range maistraCases {
+		t.Run(fmt.Sprintf("[MAISTRA][%d] %s", i, c.name), func(t *testing.T) {
+			got := wh.validate(c.in)
+
+			if got.Allowed != true {
```
jwendell, line 300: nit: `if !got.Allowed`

pkg/webhooks/validation/server/server_test.go
```diff
@@ -223,6 +251,62 @@ func TestAdmitPilot(t *testing.T) {
 			}
 		})
 	}
+
+	smeV1 := makeServiceMeshExtensionConfig(t, "v1")
+	smeV1alpha1 := makeServiceMeshExtensionConfig(t, "v1alpha1")
+	deprecatedWarning := "ServiceMeshExtension is deprecated. Please use WasmPlugin instead."
```
jwendell: nit: reuse this text (see my previous comment)

pkg/webhooks/validation/server/server.go
```diff
@@ -210,6 +210,10 @@ func (wh *Webhook) validate(request *kube.AdmissionRequest) *kube.AdmissionRespo
 		return toAdmissionResponse(fmt.Errorf("cannot decode configuration: %v", err))
 	}
 
+	if request.Kind.Kind == "ServiceMeshExtension" {
+		return &kube.AdmissionResponse{Allowed: true, Warnings: []string{"ServiceMeshExtension is deprecated. Please use WasmPlugin instead."}}
```
jwendell: nit: Make this text a variable so it can be reused in tests.

Also, what about something like:

`ServiceMeshExtension is deprecated. It works now but will be removed in a future version. Please use WasmPlugin instead.`

yannuil: Done

pkg/webhooks/validation/server/server_test.go
yannuil: Done

pkg/webhooks/validation/server/server_test.go
jwendell: 👀

PR #528, opened 2022-05-31
[misc] Skip Stackdriver tests

dgn:


dgn:
Turns out this is not required anymore. Upstream implemented a [fake GCP metadata server](https://github.com/istio/istio/blob/bd0649fc1869a1d5bffdb32e1799a29c1996b552/tests/integration/telemetry/stackdriver/common.go#L288-L292) so that the test can be run anywhere.

PR #315, opened 2021-04-13
MAISTRA-2271: Update xns-informers to fix delete events issue

bison:
This updates xns-informers to include a fix for sending delete events
for all objects when a namespace is no longer watched.

See: https://github.com/maistra/xns-informer/pull/12

PR #533, opened 2022-06-07
[no-cluster-rbac] Introduce xns-informer

dgn:
MAISTRA-2051: Add xns-informer for Kubernetes types (#223)

* vendor: Add github.com/maistra/xns-informer

* Update mirrored licenses to fix gen-check

* MAISTRA-2051: Add xns-informer for Kubernetes types

This integrates the xns-informer library for multi-namespace support,
but only for core Kubernetes types.  The informers are hard-coded to
watch all namespaces here.  The integration with MemberRoll will come
in a follow-up.

MAISTRA-2051 Add xns-informer for Istio types (#221)

MAISTRA-2051: Add MemberRollController (#227)

* MAISTRA-2051: Add MemberRollController

Includes the following:

* 75742b0cb4 - Add Makefile for Maistra client generation

This adds a Makefile with tasks for generating the Maistra specific
clients. It is adapted from the one in the Istio client-go repository.

* 5f8562cde4 - Add MemberRollController

Squashed commit, consisting of:

  * MAISTRA-417: Add MemberRollController

  * MAISTRA-450: Switch across to ConfiguredMembers and check existing
    config for changes

  * MAISTRA-1005 Do not expect namespaces to have a stable ordering

The clients have also been regenerated.

* 1b7a89dac6 - Add missing licenses

* fe2f8cda8f - Fix Copyright holder

* 06ea2db574 - MAISTRA-1666: Add `go mod vendor` as part of `make gen`

This is to make sure all PR's will have an updated
vendor directory. The `gencheck` prow job will ensure that.

* MAISTRA-2051: Update MemberRollController for xns-informers

* MAISTRA-2051: Add AddMemberRoll to kube client

* MAISTRA-2051: Configure MemberRoll controller in istiod

MAISTRA-2051: Add xns-informer for Service APIs (#229)

* MAISTRA-2051: Update xns-informers

* MAISTRA-2051: Add xns-informer for Service APIs

MAISTRA-2051: Use shared Kubernetes client in galley (#241)

This moves galley to the shared Kubernetes client, which will let it
use xns-informers for multi-namespace support like everything else.

MAISTRA-2051: Update xns-informers to latest (#252)

* MAISTRA-2051: Reset Kubernetes client to release-1.8 state

* MAISTRA-2051: Use xns-informers in Kubernetes client

* MAISTRA-2051: Un-skip tests with xns-informers issues

These tests were skipped because of issues with xns-informers, which
have now been solved.  They should no longer be skipped.

* MAISTRA-2051: Update xns-informers to latest

* MAISTRA-2051: Skip problematic integration test: TestDashboard

MAISTRA-2051: Use MultiNamespaceInformer in galley (#254)

* Revert "MAISTRA-2051: Use shared Kubernetes client in galley (#241)"

This reverts commit 0b1567a34d9c4e824ae694a0411997ca06345a2d.

* MAISTRA-2051: Use MultiNamespaceInformer in galley

This is a rework of the previous change that used the central
Kubernetes client in Galley.  The watchers in Galley manage stopping
and starting individual informers, which doesn't work well with the
factory returning cached informers.  This creates multi-namespace
informers individually instead of using the central factory.  It's
also a much smaller change overall and doesn't affect the tests.

* MAISTRA-2051: Add GetMemberRoll method to Kubernetes client

* MAISTRA-2051: Integrate MemberRoll with Galley

This configures the mulit-namespace aware infomers in Galley to
respond to changes in the set of namespaces via the MemberRoll.

This includes parts of the following:

  - 4b70b8dce74 MAISTRA-1895 Add option to enable/disable CRD scan
  - 6e8019eed8a MAISTRA-1968 allow analyzer to work with multi list watcher

MAISTRA-1724: Don't watch namespaces if MemberRoll is used (#257)

Rewrite of the following for Maistra 2.1 / Istio 1.8 rebase:

a161a53541 - MAISTRA-1724 Don't watch namespaces if MRC is used (#161)

MAISTRA-1755: invoke UpdateNamespaces() as part of MemberRollController.Register()

Cherry-pick of f0eed15f7a for Maistra 2.1 / Istio 1.8 rebase.

MAISTRA-1724: Better hanlde deletion in NamespaceController

MAISTRA-2153: Disable namespace informer if MemberRoll is used (#283)

This causes istiod to skip creating the namespace informer in the
Kubernetes service registry controller if a MemberRoll is specified,
because Maistra cannot read namespace objects.  This will affect the
behavior of multi-network deployments that do not use mesh networks
for configuration, i.e. a namespace cannot be labeled with a default
network now.

MAISTRA-2197: Push first update when cache is warm (#294)

Change SMMR Controller behavior to only send the first update to
listeners when it has a list of members ready. In other words, when
cache is synced.

Before this, the first update was sent with only one hardcoded member:
The control plane namespace. Now it is sent with the full list of SMMR
members.

While on that, switch to our own log scope, to facilitate debugging.

Manual cherry pick of https://github.com/maistra/istio/pull/291

MAISTRA-2233: Start xns-informers with empty namespace set (#303)

The multi-namespace informers created by xns-informers default to
watching all namespaces to match upstream behaviour.  This causes
permissions issues in Maistra because they will do this until the
first update from the MemberRoll controller sets the correct set of
namespaces on each informer.  We should instead explicitly configure
them with an empty set of namespaces at startup.

MAISTRA-2234: Seed MemberRoll listeners with system namespace (#302)

namespace if no SMMR resource is found.

MAISTRA-2271: Update xns-informers to fix delete events issue (#315)

This updates xns-informers to include a fix for sending delete events
for all objects when a namespace is no longer watched.

See: https://github.com/maistra/xns-informer/pull/12


jewertow:
**Important to note:**
This PR removes from the original commit handling the `WatchedNamespaces` option. Previously we had the following implementation:
```
memberRollName := args.RegistryOptions.KubeOptions.MemberRollName
if memberRollName != "" {
    err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName)
    if err != nil {
    }
} else {
    // No MemberRoll configured, set namespaces based on args.
    var namespaces []string
    if args.RegistryOptions.KubeOptions.WatchedNamespaces != "" {
        namespaces = strings.Split(args.RegistryOptions.KubeOptions.WatchedNamespaces, ",")
    } else {
        namespaces = []string{metav1.NamespaceAll}
    }
    s.kubeClient.SetNamespaces(namespaces...)
}
```
but now the `else` block was removed, so we can only either use a `ServiceMeshMemberRoll` or watch all namespaces.

jewertow (approved):

PR #543, opened 2022-06-17
Do not download envoy-centos image #234

jewertow:
Cherry-pick of a15c583541

### DIFF
--- a/prow/lib.sh
+++ b/prow/lib.sh
@@ -113,8 +113,8 @@ function build_images() {
   # Build just the images needed for tests
   targets="docker.pilot docker.proxyv2 "
 
-  # use ubuntu:jammy to test vms by default
-  nonDistrolessTargets="docker.app docker.app_sidecar_ubuntu_jammy docker.ext-authz "
+  # use centos:8 to test vms by default
+  nonDistrolessTargets="docker.app docker.app_sidecar_rockylinux_8 docker.ext-authz "
   if [[ "${JOB_TYPE:-presubmit}" == "postsubmit" ]]; then
     # We run tests across all VM types only in postsubmit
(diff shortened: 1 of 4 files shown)


dgn (approved):

prow/lib.sh
```diff
@@ -113,8 +113,8 @@ function build_images() {
   # Build just the images needed for tests
   targets="docker.pilot docker.proxyv2 "
 
-  # use ubuntu:jammy to test vms by default
-  nonDistrolessTargets="docker.app docker.app_sidecar_ubuntu_jammy docker.ext-authz "
+  # use centos:8 to test vms by default
+  nonDistrolessTargets="docker.app docker.app_sidecar_rockylinux_8 docker.ext-authz "
```
jewertow, line 117: I am not sure what to do with this change, because the original commit was changing Ubuntu Jammy to CentOS 8, but now upstream uses Rocky Linux instead of CentOS 8. What's more, tests on Rocky Linux (which is supposed to be an equivalent of RHEL/CentOS) don't work and were disabled. So since we don't support VMs yet, I suggest to skip this change for now and investigate this problem once we are ready to support VMs.
WDYT @dgn @jwendell?

jewertow:
@jwendell I can see in your [pull request](https://github.com/maistra/istio/pull/234/commits/bb004bd823640ea15e747e21bbb7c0d706a6f9e3#diff-06d8299a1749bda644043a6a0a694489ef15e09fba9402dbf3e8a717b47053e9) that the `envoy-centos` binary was removed. But this commit was probably squashed with others and it no longer contains this change, so I am confused and I am not sure if I should remove it or not? Could you help me?

PR #546, opened 2022-06-20
[no-cluster-rbac] MAISTRA-1895: Add option to enable/disable CRD scan (#260)

jewertow:
Cherry-pick of 4b70b8dce7 for Maistra 2.1 / Istio 1.8 rebase.

### DIFF
--- a/pkg/config/analysis/local/istiod_analyze.go
+++ b/pkg/config/analysis/local/istiod_analyze.go
@@ -269,7 +269,7 @@ func (sa *IstiodAnalyzer) AddRunningKubeSource(c kubelib.Client) {
 func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, revision string) {
 	// TODO: are either of these string constants intended to vary?
 	// This gets us only istio/ ones
-	store, err := crdclient.NewForSchemas(c, revision, "cluster.local", sa.kubeResources)
+	store, err := crdclient.NewForSchemas(c, revision, "cluster.local", sa.kubeResources, true)
 	// RunAndWait must be called after NewForSchema so that the informers are all created and started.
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/config/analysis/local/istiod_analyze.go
```diff
@@ -269,7 +269,7 @@ func (sa *IstiodAnalyzer) AddRunningKubeSource(c kubelib.Client) {
 func (sa *IstiodAnalyzer) AddRunningKubeSourceWithRevision(c kubelib.Client, revision string) {
 	// TODO: are either of these string constants intended to vary?
 	// This gets us only istio/ ones
-	store, err := crdclient.NewForSchemas(c, revision, "cluster.local", sa.kubeResources)
+	store, err := crdclient.NewForSchemas(c, revision, "cluster.local", sa.kubeResources, true)
```
jewertow, line 272: To be honest, I am not sure what value should be passed to `enableCRDScand` argument, because I don't understand well what are the consequences of disabling this flag. The documentation of this field says:
```
If this is set to false, all CRDs defined in the schema must be present for istiod to function.
```
so `true` seems to be safe option.

Anyway, I have to investigate it in more detail. I just leave the comment to remember about this issue.

jewertow: I think that `true` is the right value in this case, because `true` is the default value of `--enableCRDScan` argument and when it's true, `NewForSchemas` works exactly the same as in upstream Istio. `AddRunningKubeSourceWithRevision` is used by istioctl and operator, so I can't see any reason to change behavior of this function.

dgn: as discussed on Slack: this setting is required to allow istiod to run without CRD read privileges. I think passing `true` here is fine as long as we don't use these codepaths in production - AFAIK this is only called from within `istioctl` so we should be fine

dgn (approved):

PR #548, opened 2022-06-21
([no-cluster-rbac] OSSM-1420: Ensure istiod can run with no cluster-scoped privileges (#513)

jewertow:


pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
```diff
@@ -252,3 +253,55 @@ func (d *discoveryNamespacesFilter) isSelected(labels labels.Set) bool {
...
+	return d.namespaces.Has(obj.(metav1.Object).GetNamespace())
+}
+
+func (d *maistraDiscoveryNamespacesFilter) SelectorsChanged(
+	discoverySelectors []*metav1.LabelSelector) (selectedNamespaces []string, deselectedNamespaces []string) {
+	panic("Not implemented")
+}
+
+func (d *maistraDiscoveryNamespacesFilter) SyncNamespaces() error {
+	panic("Not implemented")
```
jewertow, line 288: This function must be implemented, I'm working on it.

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
```diff
@@ -252,3 +253,64 @@ func (d *discoveryNamespacesFilter) isSelected(labels labels.Set) bool {
...
+// this function should never be called if memberRollName is specified, so it's safe to call panic.
+func (d *maistraDiscoveryNamespacesFilter) SelectorsChanged(_ []*metav1.LabelSelector) (selectedNamespaces []string, deselectedNamespaces []string) {
+	panic("Not implemented")
+}
+
+// SyncNamespaces does not need to do anything, because namespaces are set by the serviceMeshMemberRollListener
+// that reacts on ServiceMeshMemberRoll changes.
+func (d *maistraDiscoveryNamespacesFilter) SyncNamespaces() error {
+	return nil
+}
```
jewertow, line 289: @luksa I had to implement this function, because it was added to the interface in the meantime. Could you take a look a this? I added some comments, because I had to investigate why some functions call `panic` and why are not needed.

dgn: Strange- this shouldn't be necessary. The function should only be called when `nsLister != nil`: https://github.com/maistra/istio/blob/5becddf2178d59e541a5012a7b9533aad88cc9fc/pilot/pkg/serviceregistry/kube/controller/controller.go#L799-L801

dgn:
I think you should squash in https://github.com/maistra/istio/commit/609d79e7eb and https://github.com/maistra/istio/commit/92ca3f2ce8 here

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
jewertow: Ah right... I forgot that when `maistraDiscoveryNamespacesFilter`, then `nsLister` is not used.

dgn:
Also, I think your integration test from https://github.com/maistra/istio/commit/f8423ff442 could be added here to make it self-contained and actually testable. Be sure to include the extensions I added in https://github.com/maistra/istio/commit/b5de0be03c (but you can keep the gateway-api parts of that separate)

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
jewertow: But returning nil is also correct and I think it's safer, because I can imagine that the upstream removes this condition, we merge it without conflict and then our Istio is broken. WDYT?

dgn: return `nil` works for me, too, but I guess it'd be more consistent to panic

jewertow:
@dgn SMMR still doesn't work, because of panicing CRD client. First I have to cherry-pick dce56c3daa, but there is a complex conflict, so I will push these commits in another PRs.

pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go
jewertow: Done

dgn (approved):

PR #549, opened 2022-06-27
[misc] SMMR integration test

jewertow:


prow/config/default.yaml
```diff
@@ -29,3 +29,5 @@ containerdConfigPatches:
   - |-
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
       endpoint = ["http://kind-registry:5000"]
+networking:
```
dgn: I don't think this should go in here, I believe it is incompatible with some changes Marko had made that add this serviceSubnet

jewertow: I'm going to remove it. I still work on this PR.

jewertow: Done

tests/integration/servicemesh/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestMain(m *testing.M) {
+	// do not change order of setup functions
+	framework.
+		NewSuite(m).
+		Setup(maistra.ApplyServiceMeshCRDs).
+		Setup(istio.Setup(&i, nil)).
+		Setup(maistra.Install).
+		Run()
+}
```
jewertow, line 39: I extracted this suite from smmr_test to reuse it later for federation_test and gateway_test. It will reduce execution time.
I know that federation will need multicluster setup, but `RequireMinClusters(2)` can be called after `Run()`.

tests/integration/servicemesh/maistra/maistra.go
```diff
@@ -0,0 +1,216 @@
...
+	{
+		"op": "add",
+		"path": "/spec/template/spec/containers/0/args/1",
+		"value": "--memberRollName=default"
+	},
+	{
+		"op": "add",
+		"path": "/spec/template/spec/containers/0/env/1",
+		"value": {
+			"name": "PRIORITIZED_LEADER_ELECTION",
```
jewertow: This is a new env variable.

jewertow:
It seems that some roles are missing:
```
022-06-28T20:20:43.223078Z	error	klog	error retrieving resource lock istio-system/istio-webhook-patcher-election: configmaps "istio-webhook-patcher-election" is forbidden: User "system:serviceaccount:istio-system:istio-reader-service-account" cannot get resource "configmaps" in API group "" in the namespace "istio-system"
2022-06-28T20:20:43.224300Z	error	klog	error retrieving resource lock istio-system/istio-namespace-controller-election: configmaps "istio-namespace-controller-election" is forbidden: User "system:serviceaccount:istio-system:istio-reader-service-account" cannot get resource "configmaps" in API group "" in the namespace "istio-system"
```

jewertow:
It turned out that remote cluster is failing:
```
2022-06-28T20:21:17.094799Z	info	kube	Initializing Kubernetes service registry "remote"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2771634]

goroutine 298 [running]:
istio.io/istio/pilot/pkg/serviceregistry/kube/controller/filter.NewMaistraDiscoveryNamespacesFilter(...)
	istio.io/istio/pilot/pkg/serviceregistry/kube/controller/filter/discoverynamespaces.go:264
istio.io/istio/pilot/pkg/serviceregistry/kube/controller.NewController({0x3808380, 0xc0019ff4a0}, {{0xc0000740ae, 0xc}, 0xc000f040b0, {0x7ffdd810b9ed, 0xd}, {0x7ffdd810b9a4, 0x7}, {0xc000f43020, ...}, ...})
	istio.io/istio/pilot/pkg/serviceregistry/kube/controller/controller.go:376 +0x7d4
istio.io/istio/pilot/pkg/serviceregistry/kube/controller.(*Multicluster).ClusterAdded(0xc001474780, 0xc0012fd980, 0xc001b3a600)
	istio.io/istio/pilot/pkg/serviceregistry/kube/controller/multicluster.go:166 +0x30d
istio.io/istio/pkg/kube/multicluster.(*Controller).handleAdd(0x268ff40?, 0xc000d8e090?, 0xc001ba59a0?)
	istio.io/istio/pkg/kube/multicluster/secretcontroller.go:400 +0xae
istio.io/istio/pkg/kube/multicluster.(*Controller).addSecret(0xc0009f4640, {{0xc000f42ff0?, 0x2?}, {0xc000d64320?, 0xc00145c558?}}, 0xc000dda500)
	istio.io/istio/pkg/kube/multicluster/secretcontroller.go:350 +0x64f
istio.io/istio/pkg/kube/multicluster.(*Controller).processItem(0xc0009f4640, {{0xc000f42ff0?, 0x2b895a0?}, {0xc000d64320?, 0xc00145c638?}})
	istio.io/istio/pkg/kube/multicluster/secretcontroller.go:199 +0x258
istio.io/istio/pkg/kube/controllers.WithReconciler.func1.1({0x2ef74a0?, 0xc001486fe0?})
	istio.io/istio/pkg/kube/controllers/queue.go:61 +0x45
istio.io/istio/pkg/kube/controllers.Queue.processNextItem({{0x3803638, 0xc00002d420}, 0xc000766978, {0x333074a, 0x13}, 0x0, 0xc0011767b0, 0xc000b9e630})
	istio.io/istio/pkg/kube/controllers/queue.go:158 +0x184
istio.io/istio/pkg/kube/controllers.Queue.Run.func1()
	istio.io/istio/pkg/kube/controllers/queue.go:112 +0x4e
created by istio.io/istio/pkg/kube/controllers.Queue.Run
	istio.io/istio/pkg/kube/controllers/queue.go:110 +0x1cb
```
I investigated why nil pointer dereference occurs in `NewMaistraDiscoveryNamespacesFilter` and it's very strange.
```
func NewMaistraDiscoveryNamespacesFilter(mrc memberroll.MemberRollController) DiscoveryNamespacesFilter {
	d := &maistraDiscoveryNamespacesFilter{}
	// here nil pointer dereference occurrs, so mrc is nil
	mrc.Register(d, "MaistraDiscoveryNamespacesFilter")
	return d
}
```
but we call `NewMaistraDiscoveryNamespacesFilter` only if `memberRoll` is specified:
```
        // Don't start the namespace informer if Maistra's MemberRoll is in use.
	if options.MemberRollName == "" {
	        ...
		if c.opts.DiscoveryNamespacesFilter == nil {
			c.opts.DiscoveryNamespacesFilter = filter.NewDiscoveryNamespacesFilter(c.nsLister, options.MeshWatcher.Mesh().DiscoverySelectors)
		}
		...
	} else if c.opts.DiscoveryNamespacesFilter == nil {
	        // MemberRollName is specified, but GetMemberRoll returns nil
		c.opts.DiscoveryNamespacesFilter = filter.NewMaistraDiscoveryNamespacesFilter(kubeClient.GetMemberRoll())
	}
```
so something goes wrong, because MemberRollController was not created:
```
func (s *Server) initKubeClient(args *PilotArgs) error {
	...
	if hasK8SConfigStore || hasKubeRegistry(args.RegistryOptions.Registries) {
                ...
		memberRollName := args.RegistryOptions.KubeOptions.MemberRollName
		if memberRollName != "" {
			err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName)
			if err != nil {
			}
		}
	}

	return nil
}
```
It seems that the condition `hasK8SConfigStore || hasKubeRegistry(args.RegistryOptions.Registries)` returns false for some reason in multicluster environment.


Anyway, I forgot that we don't test maistra in multicluster setup, so I will skip it for now and we can come back to this issue once we want to support multicluster.

dgn:
I think this should also include the changes I made in https://github.com/maistra/istio/commit/b5de0be03c -- not the Gateway API related changes, but the ones where I actually apply the limited ClusterRole

tests/integration/servicemesh/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+var i istio.Instance
+
+func TestMain(m *testing.M) {
+	// do not change order of setup functions
+	framework.
+		NewSuite(m).
+		RequireMaxClusters(1).
+		Setup(maistra.ApplyGatewayAPICRDs).
```
jewertow: It was necessary to apply Gateway API CRD, because otherwise CRD client is failing when `--enableCRDScan=false` is set. Btw. does OCP comes with Gateway API installed?

pilot/pkg/config/kube/ingressv1/controller.go
```diff
@@ -254,8 +254,10 @@ func (c *controller) SetWatchErrorHandler(handler func(r *cache.Reflector, err e
 	if err := c.ingressInformer.SetWatchErrorHandler(handler); err != nil {
 		errs = multierror.Append(err, errs)
 	}
-	if err := c.classes.Informer().SetWatchErrorHandler(handler); err != nil {
-		errs = multierror.Append(err, errs)
```
jewertow, line 258: It was causing nil pointer dereference when `--enableIngressClassName=false` was passed.

tests/integration/servicemesh/main_test.go
dgn: I think we shoujld rather disable Gateway API. That's  what we do in the operator

dgn: and no, I don't think OCP comes with Gateway API CRDs installed - which is why we should disable this by default. Reminds me.. we should really fix the CRD scan. There's a way to do it without privileges. But that's for another time

jewertow: Ok, I didn't know about flags `PILOT_ENABLE_GATEWAY_API*`. I disabled them and removed Gateway API CRD.

dgn (approved):

PR #554, opened 2022-06-29
OSSM-1325 lock service before json serialize in registryz

rcernich:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/xds/debug.go
+++ b/pilot/pkg/xds/debug.go
@@ -232,7 +232,11 @@ func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 	}
 	_, _ = fmt.Fprintln(w, "[")
 	for _, svc := range all {
-		b, err := json.MarshalIndent(svc, "", "  ")
+		b, err := func() ([]byte, error) {
+			svc.Mutex.RLock()
+			defer svc.Mutex.RUnlock()
+			return json.MarshalIndent(svc, "", "  ")
+		}()
 		if err != nil {
 			return
 		}
@@ -293,6 +297,8 @@ func (s *DiscoveryServer) endpointz(w http.ResponseWriter, req *http.Request) {
 			all := s.Env.ServiceDiscovery.InstancesByPort(ss, p.Port, nil)
 			_, _ = fmt.Fprintf(w, "\n{\"svc\": \"%s:%s\", \"ep\": [\n", ss.Hostname, p.Name)
 			for _, svc := range all {
+				// prevent crash during serialization
+				svc = svc.DeepCopy()
 				b, err := json.MarshalIndent(svc, "  ", "  ")
 				if err != nil {
 					return


pilot/pkg/xds/debug.go
```diff
@@ -293,6 +297,8 @@ func (s *DiscoveryServer) endpointz(w http.ResponseWriter, req *http.Request) {
 			all := s.Env.ServiceDiscovery.InstancesByPort(ss, p.Port, nil)
 			_, _ = fmt.Fprintf(w, "\n{\"svc\": \"%s:%s\", \"ep\": [\n", ss.Hostname, p.Name)
 			for _, svc := range all {
+				// prevent crash during serialization
+				svc = svc.DeepCopy()
```
dgn, line 301: why not use the mutex like above?

dgn:
We'll need a similar fix for 2.2- in the upstream master branch this mutex doesn't even exist anymore, though.

pilot/pkg/xds/debug.go
rcernich: This `svc` is actually a `ServiceInstance`, which embeds a `Service` object, so we'd be doing something like `svc.Service.RLock()`, and because I don't know what other nested fields in `ServiceInstance` might cause problems, it seemed best to just copy it before serializing it.

rcernich:
Yep, already done: #555.

pilot/pkg/xds/debug.go
```diff
@@ -293,6 +297,8 @@ func (s *DiscoveryServer) endpointz(w http.ResponseWriter, req *http.Request) {
 			all := s.Env.ServiceDiscovery.InstancesByPort(ss, p.Port, nil)
 			_, _ = fmt.Fprintf(w, "\n{\"svc\": \"%s:%s\", \"ep\": [\n", ss.Hostname, p.Name)
 			for _, svc := range all {
+				// prevent crash during serialization
```
oschaaf, line 300: Maybe provide a hint as to why it would crash for future readers/maintainers? (Race on shared data?)
Are we sure `DeepCopy()` is thread safe?

pilot/pkg/xds/debug.go
```diff
@@ -232,7 +232,11 @@ func (s *DiscoveryServer) registryz(w http.ResponseWriter, req *http.Request) {
 	}
 	_, _ = fmt.Fprintln(w, "[")
 	for _, svc := range all {
-		b, err := json.MarshalIndent(svc, "", "  ")
+		b, err := func() ([]byte, error) {
+			svc.Mutex.RLock()
```
oschaaf, line 236: In Envoy/c++ we tend to try and stick to RAII, so this would probably translate into a ScopedMutex or some such. Is there a similar construct we can use here?

oschaaf (approved):
Added some naive comments, but otherwise LGTM

jwendell (approved):
Interesting. I think if this code were covered by tests, it would be caught, because all tests run with `-race` flag to detect such issues.

PR #286, opened 2021-03-04
MAISTRA-1813: Add unit tests for IOR

jwendell:

PR #269, opened 2021-02-16
MAISTRA-2131: ior: honor Gateway's httpsRedirect

jwendell:
If Gateway's `httpsRedirect` is set to true, create the OpenShift Route
with Insecure Policy set to `Redirect`.

PR #556, opened 2022-06-30
[ior] MAISTRA-1400 Add IOR to Pilot

jewertow:
* [MAISTRA-1089][MAISTRA-1400][MAISTRA-1744][MAISTRA-1811]: Add IOR to Pilot (#135) (#240)

* MAISTRA-1400: Add IOR to Pilot (#135)

* MAISTRA-1400: Add IOR to Pilot

* [MAISTRA-1744] Add route annotation propagation (#158)

* MAISTRA-1811 Store resourceVersion of reconciled Gateway resource (#190)

* MAISTRA-1089 Add support for IOR routes in all namespaces (#193)

* MAISTRA-2131: ior: honor Gateway's httpsRedirect (#276)

If Gateway's httpsRedirect is set to true, create the OpenShift Route
with Insecure Policy set to `Redirect`.

Manual cherrypick from https://github.com/maistra/istio/pull/269.

* MAISTRA-2149: Make IOR robust in multiple replicas (#282)

In scenarios where multiple replicas of istiod are running,
only one IOR should be in charge of keeping routes in sync
with Istio Gateways. We achieve this by making sure IOR only
runs in the leader replica.

Also, because leader election is not 100% acurate, meaning
that for a small window of time there might be two instances
being the leader - which could lead to duplicated routes
being created if a new gateway is created in that time frame -
we also change the way the Route name is created: Instead of
having a generateName field, we now explicitly pass a name to
the Route object to be created. Being deterministic, it allows
the Route creation to fail when there's already a Route object
with the same name (created by the other leader in that time frame).

Use an exclusive leader ID for IOR

* Manual cherrypick of https://github.com/maistra/istio/pull/275

* MAISTRA-1813: Add unit tests for IOR (#286)

* MAISTRA-2051 fixes for maistra install

* MAISTRA-2164: Refactor IOR internals (#295)

Instead of doing lots of API calls on every event - this
does not scale well with lots of namespaces - keep the state
in memory, by doing an initial synchronization on start up and
updating it when receiving events.

The initial synchronization is more complex, as we have to deal with
asynchronous events (e.g., we have to wait for the Gateway store to
be warmed up). Once it's initialized, handling events as they arrive
becomes trivial.

Tests that make sure we do not make more calls to the API server than
the necessary were added, to avoid regressions.

* MAISTRA-2205: Add an option to opt-out for automatic route creation

If the Istio Gateway contains the annotation `maistra.io/manageRoute: false`
then IOR ignores it and doesn't attempt to create or manage route(s) for
this Gateway.

Also, ignore Gateways with the annotation `istio: egressgateway` as
these are not meant to have routes.


tests/integration/servicemesh/smmr_test.go
```diff
@@ -41,20 +48,29 @@ func TestSMMR(t *testing.T) {
...
-			}
-			verifyThatIngressHasVirtualHostForMember(ctx, "b")
+				if err := maistra.ApplyServiceMeshMemberRoll(ctx, namespaceGateway, namespaceB); err != nil {
+					t.Fatalf("failed to create ServiceMeshMemberRoll: %s", err)
+				}
+				verifyThatIngressHasVirtualHostForMember(t, "b")
+			})
+
+			ctx.NewSubTest("RouteCreation").Run(func(t framework.TestContext) {
+				if err := maistra.EnableIOR(t); err != nil {
```
jewertow, line 69: I don't enable IOR for all tests in the suite to increase coverage and be sure that Maistra works fine when IOR is enabled as well as disabled.

dgn (approved):

jwendell (approved):

PR #564, opened 2022-07-08
Rebase batch of misc commits

jewertow:


dgn:
TLS changes are in here as well?

jewertow:
> TLS changes are in here as well?

I rebased branch, so those changes are no longer in this PR.

bin/update_maistra_deps.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux
+
+UPDATE_BRANCH=${UPDATE_BRANCH:-"maistra-2.1"}
```
jwendell: This should probably change to something else

jwendell (approved):

bin/update_maistra_deps.sh
jewertow: Thanks for finding this! I changed this variable to maistra-2.2.

PR #567, opened 2022-07-12
OSSM-1650: cherry-pick to maistra

jewertow:


dgn (approved):

jewertow:
It works well on my local machine, so I'm investigating why integration job fails.

dgn:
Looks to me like a timing issue

PR #568, opened 2022-07-14
WIP: rebase federation

jewertow:


pilot/pkg/bootstrap/server.go
```diff
@@ -246,25 +249,83 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
+	if err := s.initIstiodCerts(args, string(istiodHost)); err != nil {
 		return nil, err
 	}
 
-	// Parse and validate Istiod Address.
-	istiodHost, _, err := e.GetDiscoveryAddress()
-	if err != nil {
+	// federation support must be initialized before config and service controllers
+	if features.EnableFederation {
+		if s.kubeClient == nil {
```
jewertow: This condition will never be true, because if kube client cannot be initialized, the initialization of server will be interrupted with an error. So I think we can just remove this condition.

pilot/pkg/bootstrap/server.go
```diff
@@ -246,25 +249,83 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
+			log.Errorf("could not initialize federation discovery server: kubeClient is nil")
+		} else {
```
jewertow: I think we could encapsulate this initialization to an unexported function to hide all its details and be consistent with other init functions.

pilot/pkg/bootstrap/server.go
```diff
@@ -246,25 +249,83 @@ func NewServer(args *PilotArgs, initFuncs ...func(*Server)) (*Server, error) {
...
+			peerCertVerifier, err := s.createPeerCertVerifier(args.ServerOptions.TLSOptions)
+			if err != nil {
+				return nil, err
+			}
+			if peerCertVerifier == nil {
+				panic("No peerCertVerifier")
```
jewertow: Why panic and not an error? Both should stop starting istiod server.

pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -131,6 +131,14 @@ func (s *Server) initConfigController(args *PilotArgs) error {
 		})
 	}
 
+	if features.EnableFederation {
+		if s.federation == nil {
+			log.Errorf("Federation support disabled: federation not initialized")
```
jewertow, line 136: This log does not make sense, because it says "Federation support disabled" when `features.EnableFederation` is true.

The second thing is that we don't have to check if both conditions `features.EnabledFederation` and `s.federation == nil` are true, because they cannot happen together. The reason of that is the initialization of istiod server is interrupted when federation is enabled, but not succeeded.

I would like to simplify this entire if-else statement to just:
```
if s.federation != nil {
    s.ConfigStores = append(s.ConfigStores, s.federation.ConfigStore())
}
```

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -0,0 +1,1116 @@
...
+		if importConfig == nil {
+			return false
+		}
+		c.storeLock.RLock()
+		defer c.storeLock.RUnlock()
+		return c.importConfigGeneration == importConfig.Generation
+	}() {
+		return
+	}
+	func() {
```
jewertow, line 252: Does it have to be a function?

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -0,0 +1,1116 @@
...
+	url := c.discoveryURL + "/v1/services/"
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: '%s': %s", url, err)
+	}
+	req.Header.Add("discovery-service", c.discoveryServiceName)
+	req.Header.Add("remote", fmt.Sprint(common.RemoteChecksum(c.remote)))
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to GET URL: '%s': %s", url, err)
```
jewertow, line 284: WDYT about adding more context to this error message, like: `"failed to poll services [URL=%s]: %s", err`?

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -0,0 +1,1116 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("status code is not OK: %v (%s)", resp.StatusCode, resp.Status)
```
jewertow, line 287: The same as above: `"failed to poll services: unexpected status code %d", statusCode`?

PR #571, opened 2022-07-26
OSSM-1105: Support namespace portion in gateway hostnames

jwendell:


pilot/pkg/config/kube/ior/ior_test.go
```diff
@@ -194,6 +194,46 @@ func TestCreate(t *testing.T) {
...
+			map[string]string{"istio": "ingressgateway"},
+			1,
+			"",
+			false,
+			nil,
+		},
+		{
+			"Host with a namespace and wildcard",
+			"istio-system",
+			[]string{"*/*.one.org"},
```
oschaaf, line 230: Would it make sense to test the behaviour of leading and trailing slashes, as well as consecutive slashes in addition to this?

pilot/pkg/config/kube/ior/fake.go
```diff
@@ -103,6 +104,10 @@ func (fk *FakeRouter) Create(ctx context.Context, route *v1.Route, opts metav1.C
 	fk.routesLock.Lock()
 	defer fk.routesLock.Unlock()
 
+	if strings.Contains(route.Spec.Host, "/") {
+		return nil, fmt.Errorf("invalid hostname")
```
oschaaf, line 108: Maybe append the value of `route.Spec.Host` in the error message too? (Unless that ends up in the logs and it could be considered sensitive).

pilot/pkg/config/kube/ior/route.go
```diff
@@ -579,6 +577,12 @@ func getRouteName(namespace, name, actualHost string) string {
 // It also returns the route's WildcardPolicy based on the hostname
 func getActualHost(originalHost string, emitWarning bool) (string, v1.WildcardPolicyType) {
 	wildcard := v1.WildcardPolicyNone
+
+	if strings.Contains(originalHost, "/") {
```
oschaaf, line 581: Instead of checking if `/` occurs, would it be safer to count occurrences of `/` .. and only proceed to treat it as a containing a namespace part when that count equals `1`? (not sure what to do when the count would > 1 though)

pilot/pkg/config/kube/ior/route.go
```diff
@@ -365,10 +365,8 @@ func (r *route) verifyResourceVersions(cfg config.Config) error {
 		return fmt.Errorf("could not find an internal reference to gateway %s/%s", cfg.Namespace, cfg.Name)
 	}
 
-	for _, route := range syncRoute.routes {
```
oschaaf, line 368: I have a hard time grokking how this change related to the rest of this pr .. could you perhaps explain? Probably due to me being naive with respect to this code base .. but if you anticipate future readers may benefit, perhaps add it as a code-level comment.

oschaaf (approved):
leaving some comments

pilot/pkg/config/kube/ior/ior_test.go
jwendell: Invalid entries (e.g. `www.host*.com`) are already handled by Istio. The validation webhook will prevent such entries to be applied. So, we're safe here.

pilot/pkg/config/kube/ior/fake.go
jwendell: ditto

pilot/pkg/config/kube/ior/route.go
jwendell: ditto

pilot/pkg/config/kube/ior/route.go
jwendell: See the commit message. It's a separate commit.

oschaaf: Got it, sorry about the noise.

jewertow (approved):
Reading test case structures without keys is painful. What do you think about adding then in a subsequent commit or PR?

jewertow:
Please cherry-pick these commits to maistra branch after merging this PR.

PR #576, opened 2022-08-04
Remove bug report type

jwendell:


oschaaf (approved):
not sure how this change could cause a CI failure?
therefore LGTM regardless

jwendell:
@oschaaf it's very very likely a flaky test. Will retry once the other jobs finish.

PR #581, opened 2022-08-10
OSSM-1871 Fix API group in ownerReference

luksa:


jwendell (approved):
If you are going to fix this upstream, consider backporting it to 1.14 and 1.15 as well (just add the labels there), then we don't need to cherrypick it here for 2.3

jewertow:
Could you explain this change?

jewertow (approved):

PR #584, opened 2022-08-16
Revert setting ALPNDownstreamWithMxc on TCP gateways in ISTIO_MUTUAL mode

jewertow:
This is cherry-pick of b4c427158e47d6c3cfe026b8b96a5209d60e589c from
upstream Istio. This commit must be reverted, because it was not
included in Istio 1.14 and it breaks traffic in some cases, e.g. in
federation deployment.

Original pull request: https://github.com/istio/istio/pull/38990


rcernich (approved):

jwendell:
@jewertow this PR targets main. Is that correct? If not, I think it's better to close it and create a new one, otherwise prow jobs get confused.

jewertow:
@jwendell thanks! That was a mistake.

PR #585, opened 2022-08-17
Federation deployment

jewertow:

PR #586, opened 2022-08-17
Revert setting ALPNDownstreamWithMxc on TCP gateways in ISTIO_MUTUAL mode

jewertow:
This is cherry-pick of b4c427158e47d6c3cfe026b8b96a5209d60e589c from
upstream Istio. This commit must be reverted for now, because it was not
included in Istio 1.14 and it breaks traffic in some cases, e.g. in
federation deployment.

Original pull request: https://github.com/istio/istio/pull/38990


jwendell:
@luksa could you please take a look at this? It's clear there's something wrong with the test or the environment. It's not just a flaky, since it's failing consecutively a lot.

jewertow:
@jwendell I'm also looking at this problem, but it would be nice if Marko checked that out as well. I just would like to note that this test works locally, passed in [this PR](https://github.com/maistra/istio/pull/585) and always fail at a different stage, so there is no problem with the test itself.

jwendell:
Perhaps the change in `common/scripts/kind_provisioner.sh` is not necessary anymore?
https://github.com/maistra/istio/pull/530/files

jwendell (approved):

PR #609, opened 2022-09-09
Use Maistra proxy based on CentOS Stream 8 in integration tests

jewertow:
Istio has been using its own tool to build container images for some time. This tool is [docker-builder](https://github.com/maistra/istio/tree/maistra-2.3/tools/docker-builder).
Unfortunately docker-builder does not support `RUN` command, so we cannot install `iptables` and `openssl` packages in proxyv2 with `RUN dnf install ...`.

To work around this limitation, I built an image with necessary packages, pushed to my registry and used it as the base image for proxyv2.

Image `quay.io/jewertow/base-maistra-proxyv2:2.3` was built from the following Dockerfile:
```
FROM quay.io/centos/centos:stream8

RUN dnf -y upgrade --refresh --nobest && \
    dnf -y install iptables iproute openssl && \
    dnf -y clean all
```

I think that it does not make sense to extend docker-builder to support `RUN`, because it's too complicated (if possible)
and we just have to build our official base image and use it as I did with my image.

**Update:**

I figured out that we can disable building container images with crane by setting `ISTIO_DOCKER_BUILDER=docker`, so I updated images to use CentOS Stream 8 as a base image and added image `app_sidecar_centos_stream_8` to test VMs.

### DIFF
--- a/pilot/docker/Dockerfile.proxyv2
+++ b/pilot/docker/Dockerfile.proxyv2
@@ -1,28 +1,14 @@
-# BASE_DISTRIBUTION is used to switch between the old base distribution and distroless base images
-ARG BASE_DISTRIBUTION=debug
-
-# Version is the base image version from the TLD Makefile
-ARG BASE_VERSION=latest
-
-# The following section is used as base image if BASE_DISTRIBUTION=debug
-FROM gcr.io/istio-release/base:${BASE_VERSION} as debug
-
-# The following section is used as base image if BASE_DISTRIBUTION=distroless
-# This image is a custom built debian11 distroless image with multiarchitecture support.
-# It is built on the base distroless image, with iptables binary and libraries added
-# The source can be found at https://github.com/istio/distroless/tree/iptables
-# This version is from commit 6faf2f9348d382c6a95bbc3171b89406103cb0ad.
-FROM gcr.io/istio-release/iptables@sha256:64ba191166d2c3f87815148cc1a9f530b770e0dfe81acb164ab79cbb31da582c as distroless
-
-# This will build the final image based on either debug or distroless from above
-# hadolint ignore=DL3006
-FROM ${BASE_DISTRIBUTION:-debug}
+FROM quay.io/centos/centos:stream8
 
 WORKDIR /
 
 ARG proxy_version
 ARG istio_version
 ARG SIDECAR=envoy
+
+RUN dnf -y upgrade --refresh --nobest && \
+    dnf -y install iptables iproute openssl && \
+    dnf -y clean all
 
 # Copy Envoy bootstrap templates used by pilot-agent
 COPY envoy_bootstrap.json /var/lib/istio/envoy/envoy_bootstrap_tmpl.json
--- a/pilot/docker/Dockerfile.proxyv2
+++ b/pilot/docker/Dockerfile.proxyv2
@@ -6,6 +6,7 @@ ARG proxy_version
 ARG istio_version
 ARG SIDECAR=envoy
 
+# hadolint ignore=DL3041
 RUN dnf -y upgrade --refresh --nobest && \
     dnf -y install iptables iproute openssl && \
     dnf -y clean all
--- a/pkg/test/framework/components/echo/config.go
+++ b/pkg/test/framework/components/echo/config.go
@@ -53,13 +53,14 @@ type Configurable interface {
 type VMDistro = string
 
 const (
-	UbuntuXenial VMDistro = "UbuntuXenial"
-	UbuntuJammy  VMDistro = "UbuntuJammy"
-	Debian11     VMDistro = "Debian9"
-	Centos7      VMDistro = "Centos7"
-	Rockylinux8  VMDistro = "Centos8"
+	UbuntuXenial  VMDistro = "UbuntuXenial"
+	UbuntuJammy   VMDistro = "UbuntuJammy"
+	Debian11      VMDistro = "Debian9"
+	Centos7       VMDistro = "Centos7"
+	Rockylinux8   VMDistro = "Centos8"
+	CentosStream8 VMDistro = "CentosStream8"
 
-	DefaultVMDistro = UbuntuJammy
+	DefaultVMDistro = CentosStream8
 )
 
 // Config defines the options for creating an Echo component.
--- a/prow/integ-suite-kind.sh
+++ b/prow/integ-suite-kind.sh
@@ -42,7 +42,7 @@ KIND_CONFIG=""
 CLUSTER_TOPOLOGY_CONFIG_FILE="${ROOT}/prow/config/topology/multicluster.json"
 
 export FAST_VM_BUILDS=true
-export ISTIO_DOCKER_BUILDER=crane
+export ISTIO_DOCKER_BUILDER=docker
 
 PARAMS=()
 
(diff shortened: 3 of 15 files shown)


jwendell (commented):
Not sure about this.

I have experimented with this a week ago: https://github.com/maistra/istio/pull/604 and I think it worked...

Also, please align with @bmangoen and hisr PR: https://github.com/maistra/istio/pull/606

jewertow:
Your PR didn't work. Look at the [log](https://storage.googleapis.com/maistra-prow-testing/pr-logs/pull/maistra_istio/604/istio-integration-pilot-2.3/1565692492618141696/build-log.txt) - everything failed. You can find there the following warnings:
```
2022-09-02T13:29:02.519537Z	warn	dockerfile	Skipping RUN: [dnf -y upgrade --refresh --nobest &&     dnf -y install iptables iproute openssl &&     dnf -y clean all]
2022-09-02T13:29:02.521035Z	warn	dockerfile	Skipping RUN: [dpkg -i /tmp/istio-sidecar.deb && rm /tmp/istio-sidecar.deb]
```
And as you can see in [one of istio-init logs](https://storage.googleapis.com/maistra-prow-testing/pr-logs/pull/maistra_istio/604/istio-integration-pilot-2.3/1565692492618141696/artifacts/pilot-094f83e4065e41bb81d08b899/_suite_context/echo-2-96428-state-1646371241/primary-0/a-v1-57f5b84fbd-5zl7c_istio-init.log), iptables-restore did not exist in the proxy container.

jwendell:
hmm, I see. Is it possible to install the `libssl1` in the [upstream] base image? If yes, then we don't need a special Dockerfile based on centos. We can use the same upstream mechanism without any changes.

In the past we needed this image based on centos because the envoy binary built on centos was not compatible with the libc present on Ubuntu. It looks like it's not the case anymore. I was able to run our Envoy on this base image (which is derived from Ubuntu). The problem now is that our Envoy relies on the presence of the `/lib64/libssl.so.1.1` library, which is provided by the Ubuntu package `libssl1`, which is not installed by default.

If there's no way to have this package installed in the base image (and since it's not possible to have `RUN apt install` anymore), I suggest you add a new target as a dependency for the `make docker` target .i.e, that will run first, that will create a new base image, let's say, `maistra-base` which is merely the Dockerfile you created this custom base image. Then, for every run, we first create this custom base image, and proxy is based on that. In other words, we would use your solution, but baking it at build-time instead of relying in a pre-built image hosted somewhere.

jewertow:
@jwendell thanks for your advice. I was trying to run tests on Ubuntu, but currently it provides libssl.so.3.0 by default. I was trying to find a workaround to install OpenSSL 1.1.1, e.g. using repositories from older releases, but I had some other issues, so I gave up in favor of bringing an image based on CentOS Stream 8.

pkg/test/framework/components/echo/config.go
```diff
@@ -53,13 +53,14 @@ type Configurable interface {
...
-
-	DefaultVMDistro = UbuntuJammy
+	UbuntuXenial  VMDistro = "UbuntuXenial"
+	UbuntuJammy   VMDistro = "UbuntuJammy"
+	Debian11      VMDistro = "Debian9"
+	Centos7       VMDistro = "Centos7"
+	Rockylinux8   VMDistro = "Centos8"
+	CentosStream8 VMDistro = "CentosStream8"
+
+	DefaultVMDistro = CentosStream8
```
jwendell, line 63: Is there any reason to have a centos8 image?

prow/integ-suite-kind.sh
```diff
@@ -42,7 +42,7 @@ KIND_CONFIG=""
 CLUSTER_TOPOLOGY_CONFIG_FILE="${ROOT}/prow/config/topology/multicluster.json"
 
 export FAST_VM_BUILDS=true
-export ISTIO_DOCKER_BUILDER=crane
+export ISTIO_DOCKER_BUILDER=docker
```
jwendell, line 45: I guess this is the change that allows you have have `RUN` commands in the proxy dockerfile?

tests/integration/security/egress_gateway_origination_test.go
```diff
@@ -45,6 +45,8 @@ import (
 // TestSimpleTlsOrigination test SIMPLE TLS mode with TLS origination happening at Gateway proxy
 // It uses CredentialName set in DestinationRule API to fetch secrets from k8s API server
 func TestSimpleTlsOrigination(t *testing.T) {
+	// https://issues.redhat.com/browse/OSSM-755
```
jwendell: nit: Put a `FIXME:` prefix to those comments? Perhaps it makes it easier for us in the future to locate them?

tests/integration/security/egress_gateway_origination_test.go
```diff
@@ -45,6 +45,8 @@ import (
...
+	t.Skip("https://github.com/istio/istio/0")
```
jwendell: I think the format is `https://github.com/istio/istio/issues/0`

jewertow: Thanks, I fixed it.

tests/integration/security/egress_gateway_origination_test.go
jewertow: Done

prow/integ-suite-kind.sh
jewertow: Yes, you're right. As you can see, I updated the PR description and mentioned about this variable.

pkg/test/framework/components/echo/config.go
jewertow: Nope, it is not used, but we don't use other images as well. I didn't remove them to not increase the scope of this PR.

tests/integration/telemetry/stats/prometheus/wasm/stats_wasm_filter_test.go
```diff
@@ -39,6 +39,8 @@ func TestWasmStatsFilter(t *testing.T) {
 
 func TestMain(m *testing.M) {
 	framework.NewSuite(m).
+		// FIXME: https://issues.redhat.com/browse/OSSM-2000
+		Skip("https://github.com/istio/istio/issues/0").
```
jewertow, line 43: It turned out that we have to disable whole suite. It works on my local machine, but does not on prow. I was trying to find out what's wrong there, but it seems to be a more complicated issue. I will solve it later.

pilot/docker/Dockerfile.proxyv2
```diff
@@ -1,29 +1,16 @@
...
 
 WORKDIR /
 
 ARG proxy_version
 ARG istio_version
 ARG SIDECAR=envoy
 
+# hadolint ignore=DL3041
+RUN dnf -y upgrade --refresh --nobest && \
+    dnf -y install iptables iproute openssl && \
```
yxun, line 11: Do we need to install a specific versoin of openssl or just the latest one ?

jwendell (approved):

pilot/docker/Dockerfile.proxyv2
jewertow: We need OpenSSL 1.1.1 and our istio-proxy expects to find libssl.so.1.1. In CentOS Stream 8 the package `openssl` is exactly this one that we expect. I don't think that this package will be replaced by another version, because the next one is 3.0.0 which provides breaking API and ABI and upgrading from 1.1.1 to 3.0.0 is a big challenge, so it will probably not happen.

PR #614, opened 2022-09-19
OSSM-1962: Use EndpointSlices instead of Endpoints in federation controller

jewertow:
OSSM 2.3 will be supported on OpenShift 4.8 and higher, which is compatible with Kubernetes 1.21, so we don't have to worry about whether to use Endpoints or EndpointSlices. 

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -445,7 +445,7 @@ func (c *Controller) gatewayForNetworkAddress() []model.NetworkGateway {
 		remotePort = common.DefaultFederationPort
 	}
 	for _, address := range c.remote.Addresses {
-		addrs, err := c.getIPAddrsForHostOrIP(address)
+		addrs, err := c.getIPOrResolveHost(address)
 		if err != nil {
 			c.logger.Errorf("error resolving IP addr for federation network %s: %v", address, err)
 		} else {
@@ -462,12 +462,12 @@ func (c *Controller) gatewayForNetworkAddress() []model.NetworkGateway {
 	return gateways
 }
 
-func (c *Controller) getIPAddrsForHostOrIP(host string) ([]string, error) {
-	gwIP := net.ParseIP(host)
+func (c *Controller) getIPOrResolveHost(ipOrHost string) ([]string, error) {
+	gwIP := net.ParseIP(ipOrHost)
 	if gwIP != nil {
 		return []string{gwIP.String()}, nil
 	}
-	return net.LookupHost(host)
+	return net.LookupHost(ipOrHost)
 }
 
 // store has to be Lock()ed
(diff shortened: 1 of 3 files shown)


pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -491,27 +491,31 @@ func (c *Controller) updateGateways(serviceList *federationmodel.ServiceListMess
...
-			c.logger.Warnf("unable to retrieve ServiceAccount information for imported services accessed through %s: %s",
+			c.logger.Warnf("failed to retrieve ServiceAccount information for imported services accessed through %s: %s",
 				c.egressName, err)
 		}
 	}
 	var addrs []model.NetworkGateway
 	var sas []string
-	for _, subset := range endpoints.Subsets {
-		for index, address := range subset.Addresses {
-			if subset.Ports[index].Port == common.DefaultFederationPort {
```
jewertow, line 511: There is no need to look for a port under an index, because the [API reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoint-slice-v1/) of EndpoinSlices says:
```
ports specifies the list of network ports exposed by each endpoint in this slice
```
so it means that all addresses in all endpoints from a given EndpointSlice object expose ports specified in the `Ports` array.

jewertow: I will also fix it in the previous versions.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -491,27 +491,31 @@ func (c *Controller) updateGateways(serviceList *federationmodel.ServiceListMess
 }
 
 func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string) {
-	endpoints, err := common.EndpointsForService(c.kubeClient, c.egressName, c.namespace)
+	endpointSlices, err := common.EndpointSlicesForService(c.kubeClient, c.egressName, c.namespace)
 	if err != nil {
-		c.logger.Errorf("unabled to retrieve endpoints for federation egress gateway %s: %s", c.egressName, err)
+		c.logger.Errorf("failed to retrieve EndpointSlices for federation egress gateway %s: %s", c.egressName, err)
```
jewertow, line 496: I thought that "failed" is more common phrase than "unable" when it comes to search in logs. WDYT?

rcernich (approved):

pkg/servicemesh/federation/common/util.go
```diff
@@ -38,9 +38,23 @@ func DefaultFederationCARootResourceName(instance *v1.ServiceMeshPeer) string {
 	return fmt.Sprintf("%s-ca-root-cert", instance.Name)
 }
 
-// EndpointsForService returns the Endpoints for the named service.
-func EndpointsForService(client kube.Client, name, namespace string) (*corev1.Endpoints, error) {
-	return client.KubeInformer().Core().V1().Endpoints().Lister().Endpoints(namespace).Get(name)
+// EndpointSlicesForService returns the Endpoints for the named service.
+func EndpointSlicesForService(client kube.Client, name, namespace string) ([]*discoveryv1.EndpointSlice, error) {
+	endpointSlices, err := client.KubeInformer().Discovery().V1().EndpointSlices().Lister().EndpointSlices(namespace).
+		List(kubelabels.Set(map[string]string{discoveryv1.LabelServiceName: name}).AsSelector())
```
luksa: This can be simplified a bit.

```suggestion
		List(kubelabels.Set{discoveryv1.LabelServiceName: name}.AsSelector())
```

luksa (approved):

pkg/servicemesh/federation/common/util.go
jewertow: Thanks, done.

PR #616, opened 2022-09-20
OSSM-2006 Fix multiNamespaceInformer.HasSynced()

luksa:


dgn (approved):
Thank you! I think the decision to default to `NamespaceAll` was flawed, I remember having to go through the istio code base and making 100% sure that we initialize with the right namespace set in order not to accidentally watch all namespaces (this was a problem in unit tests as they used different init methods IIRC)

dgn:
The problem with the failing tests probably is because you need to call `SetNamespaces(v1.NamespaceAll)` somewhere- otherwise we're watching nothing, which obviously doesn't work for the tests. I helped myself with this solution in the 2.3 rebase:
https://github.com/maistra/istio/blob/6e442e18f6b47e66dbdfd14e697b34de2d986d75/pkg/kube/client.go#L556-L560

That might not be very smart though, as it is always called (I don't remember why I don't check for mrc existence). In any case, it was so hard to figure out all the places/tests that needed to init the NamespaceSet that I did it in this central location. It might make more sense to do it in the bootstrap file

rcernich:
Yep.  I started going through this last night.  There are about 66 instances. :(

luksa:
I'm trying to find a good solution for this. Currently, I think we should initialize informer factories with `NamespaceAll`, but then clear the namespace list (and mark the informers as uninitialized) when the informer factory is added as a listener to the member roll controller. This should ensure that the informers work correctly in both scenarios (in unit tests, which have no member roll controller, in the normal istiod execution where the member roll controller is the only thing that has control of the namespace list).

luksa:
Replaced by #643

PR #617, opened 2022-09-20
OSSM-2049: Fix handling Endpoints in federation controller

jewertow:
While working on #614 I noticed that we incorrectly handle Endpoint ports.

### DIFF
--- a/pilot/pkg/serviceregistry/federation/controller.go
+++ b/pilot/pkg/serviceregistry/federation/controller.go
@@ -35,6 +35,7 @@ import (
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pilot/pkg/serviceregistry"
 	"istio.io/istio/pilot/pkg/serviceregistry/provider"
+	"istio.io/istio/pilot/pkg/util/sets"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/constants"
 	"istio.io/istio/pkg/config/host"
@@ -502,7 +503,7 @@ func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string) 
 		}
 	}
 	var addrs []model.NetworkGateway
-	var sas []string
+	uniqueSAs := sets.Set{}
 	for _, subset := range endpoints.Subsets {
 		if !common.ContainsPort(subset.Ports, common.DefaultFederationPort) {
 			continue
@@ -512,10 +513,12 @@ func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string) 
 				Addr: address.IP,
 				Port: uint32(common.DefaultFederationPort),
 			})
-			sas = append(sas, serviceAccountByIP[address.IP])
+			if sa, exists := serviceAccountByIP[address.IP]; exists {
+				uniqueSAs.Insert(sa)
+			}
 		}
 	}
-	return addrs, sas
+	return addrs, uniqueSAs.UnsortedList()
 }
 
 func (c *Controller) convertServices(serviceList *federationmodel.ServiceListMessage) {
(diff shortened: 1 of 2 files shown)


pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -504,21 +504,15 @@ func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string)
 	var addrs []model.NetworkGateway
 	var sas []string
 	for _, subset := range endpoints.Subsets {
-		for index, address := range subset.Addresses {
-			if subset.Ports[index].Port == common.DefaultFederationPort {
```
jewertow, line 508: As the documentation of [Endpoint API](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/) says, a subset of endpoints contains:
```
IP addresses which offer the related ports that are marked as ready.
```
so there is no need to look for a port at index of an address.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -504,21 +504,15 @@ func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string)
...
-				if err != nil {
-					c.logger.Errorf("error converting to IP address from %s: %s", address.IP, err)
-					continue
-				}
-				for _, ip := range ips {
-					addrs = append(addrs, model.NetworkGateway{
-						Addr: ip,
-						Port: uint32(common.DefaultFederationPort),
-					})
-					sas = append(sas, serviceAccountByIP[ip])
```
jewertow, line 519: Shouldn't we also check whether `serviceAccountByIP` has a value at given ip? I guess that we cannot be 100% sure that ServiceAccounts and Endpoints are in sync. Then `sas` might contain empty strings. Do you agree with me?

luksa: Yup, looks like `sas` can contain multiple empty strings and the same service account name repeated several times. While neither empty strings nor repeated SA names break anything, we should clean this up and make `sas` a set.

pilot/pkg/serviceregistry/federation/controller.go
luksa: Correct. Ports are a property of the subset, not of the address.

luksa (approved):
LGTM. Good work!

pilot/pkg/serviceregistry/federation/controller.go
jewertow: Thanks. I fixed it, look at the last commit.

pilot/pkg/serviceregistry/federation/controller.go
```diff
@@ -512,10 +513,12 @@ func (c *Controller) getEgressServiceAddrs() ([]model.NetworkGateway, []string)
...
 				Port: uint32(common.DefaultFederationPort),
 			})
-			sas = append(sas, serviceAccountByIP[address.IP])
+			if sa, exists := serviceAccountByIP[address.IP]; exists {
+				uniqueSAs.Insert(sa)
+			}
 		}
 	}
-	return addrs, sas
+	return addrs, uniqueSAs.UnsortedList()
```
luksa: Given that `UnsortedList()` returns elements in random order (i.e. different order every time you invoke it), can this change now break any processing that happens afterwards? For example, when you have an array stored in a custom resource, and a controller calculates this array during every reconcile, if the order of elements changes, the controller constantly updates the resource, even though semantically there is no real change in the array. 

The previous version of this function always returned the same order, since it was based on the order of the addresses and subsets in the resource, but that's no longer the case. Would returning a sorted list be more appropriate?

jewertow: I didn't take it into account, but it definitely makes sense. Thanks!

jewertow: I fixed it.

jewertow: Fixed

dgn (approved):

PR #623, opened 2022-09-22
OSSM-2006 Fix multiNamespaceInformer.HasSynced()

luksa:


luksa:
I noticed the maistra/api dependency didn't point to the correct commit, so I fixed it in a separate commit within this PR, just so that the changes introduced by the actual commit (the 2nd commit in this PR) are more clearly visible.

pkg/kube/client.go
```diff
@@ -412,7 +412,6 @@ func newClientInternal(clientFactory util.Factory, revision string) (*client, er
 		return nil, err
 	}
 	c.dynamicInformer = xnsinformers.NewDynamicSharedInformerFactory(c.dynamic, resyncInterval)
-	c.dynamicInformer.SetNamespaces() // Maistra needs to start with an empty namespace set.
```
luksa, line 415: There's no longer any need for this, since it happens automatically when you register the informer as a listener in the member roll controller.

pkg/kube/client.go
```diff
@@ -555,9 +534,6 @@ func (c *client) GetMemberRoll() memberroll.MemberRollController {
 // RunAndWait starts all informers and waits for their caches to sync.
 // Warning: this must be called AFTER .Informer() is called, which will register the informer.
 func (c *client) RunAndWait(stop <-chan struct{}) {
-	// make sure to watch all namespaces if we're not using a MemberRollController
-	c.SetNamespaces(metav1.NamespaceAll)
-
```
luksa, line 560: This is no longer needed, because the client (and underlying informer factories) are already configured with `NamespaceAll`. 

Previously, this was necessary because, while the factory was initially created with `NamespaceAll`, the list of namespaces was then immediately set to empty (via the `istioinformer.WithNamespaces()` call (see above)). Now, the latter is no longer the case, so there's no need to set the client back to `NamespaceAll` here.

jwendell (approved):

dgn (approved):

PR #624, opened 2022-09-22
[WIP] OSSM-755 Re-enable security integration tests

dgn:


jwendell:
Safe to close this in favor of #657?

jewertow:
Yes, it was an experimental PR and the root cause was solved, so it's no longer needed.

PR #625, opened 2022-09-23
OSSM-2041 Add --logKubernetesApiRequests option to istiod

luksa:
This option logs each individual request sent to the Kubernetes API. In addition to enabling this option, you need to set the log level to INFO.

### DIFF
--- a/pilot/cmd/pilot-discovery/app/cmd.go
+++ b/pilot/cmd/pilot-discovery/app/cmd.go
@@ -197,6 +197,9 @@ func addFlags(c *cobra.Command) {
 	c.PersistentFlags().IntVar(&serverArgs.RegistryOptions.KubeOptions.KubernetesAPIBurst, "kubernetesApiBurst", 160,
 		"Maximum burst for throttle when communicating with the kubernetes API")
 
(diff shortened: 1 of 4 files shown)


jwendell (commented):
Any chance to get this merged in upstream, and cherry-picked to 1.14 branch so we can pick it on next sync?

luksa:
> Any chance to get this merged in upstream, and cherry-picked to 1.14 branch so we can pick it on next sync?

I don't think upstream really needs this. We do because we send many more requests to the API server.

jwendell:
> > Any chance to get this merged in upstream, and cherry-picked to 1.14 branch so we can pick it on next sync?
> 
> I don't think upstream really needs this. We do because we send many more requests to the API server.

Well, that's what community means, no? It's an open source project, used by many companies. It's useful to us, might be to someone else. It's no an intrusive change, and it's opt-in, so, why not?

luksa:
> Well, that's what community means, no? It's an open source project, used by many companies. It's useful to us, might be to someone else. It's no an intrusive change, and it's opt-in, so, why not?

True, but we need this now so that QE can verify the cluster-wide install, so there's no time to add this feature through upstream.

jwendell (approved):
Fair enough. This is just me trying to cope with the "upstream first" policy. I'm happy if it's "upstream later". Better than "no upstream" at all :)

PR #626, opened 2022-09-23
[wip] OSSM-2042 Ensure mesh config map is read when revision is "default"

luksa:


pilot/pkg/bootstrap/mesh.go
```diff
@@ -119,7 +119,7 @@ func (s *Server) initMeshNetworks(args *PilotArgs, fileWatcher filewatcher.FileW
 
 func getMeshConfigMapName(revision string) string {
 	name := defaultMeshConfigMapName
-	if revision == "" || revision == "default" {
+	if revision == "" {
```
jwendell, line 122: Is this specific to us? Should we submit this upstream (and cherry pick to 1.14)?

luksa: I think this issue might also appear in upstream, but I haven't tested it. In any case, we don't have time to wait for this to be merged upstream and then pulled in here, because the cluster-wide option requires this to be fixed immediately.

luksa: Yup, upstream also has this issue. I've filed https://github.com/istio/istio/issues/41139.

luksa: Note: in Maistra, we always set the revision (it's never ""). Upstream, this isn't the case, so the fix might need to be different. That's also why I don't want to waste time fixing this upstream.

jwendell (approved):

luksa:
The integration test failures are legit. It looks like the revision is "default", but the configmap is named "istio" instead of "istio-default". I need to find a different approach to this issue.

luksa:
Closing. This approach doesn't work and would require a careful review of the istio code to find all places where revision `""` or `"default"` is referenced.

PR #627, opened 2022-09-26
OSSM-2049: Fix handling ServiceAccounts in federation controller

jewertow:


luksa (approved):

dgn (approved):
lgtm. any way we can add a test? I know the issue says it's hard to reproduce but I would be interested in what kind of scenarios this occurs?

PR #631, opened 2022-09-28
OSSM-1301 Wait for Route resource type to become available on ior startup

luksa:


pilot/pkg/config/kube/ior/route.go
```diff
@@ -105,8 +105,9 @@ func newRoute(
 	mrc controller.MemberRollController,
 	stop <-chan struct{},
 ) (*route, error) {
-	if !kubeClient.IsRouteSupported() {
-		return nil, fmt.Errorf("routes are not supported in this cluster")
+	for !kubeClient.IsRouteSupported() {
+		IORLog.Infof("routes are not supported in this cluster; waiting for Route resource to become available...")
+		time.Sleep(10 * time.Second)
```
jwendell, line 110: Is 10s too much maybe?

jwendell (approved):

jwendell:
btw, the linter failure seems legit.

pilot/pkg/config/kube/ior/route.go
luksa: Yeah, maybe. Note that the code within `IsRouteSupported()` performs full API Discovery, which takes several seconds, so even if I reduce this sleep, it's going to take a lot of time between attempts.

In normal cluster operation, the code in the for loop won't run at all, so it's not a big deal if it takes x + 10 seconds to make the next attempt when the cluster is cold-started. WDYT?

jwendell: SGTM. If somehow Route is not available this loop will stuck forever. It's an user error anyway, since he enabled IOR in a plain k8s cluster, for example. Perhaps repeat this 10 times and give up with a message like "Disabling IOR because..."? Not a strong opinion though.

luksa: Actually, I think giving up is dangerous. If ior is enabled, then it should work. If it can't work, istiod should either retry indefinitely (as in this PR), or terminate (either immediately or after a few tries), so that the cluster admin can immediately see that something is wrong with istiod.

PR #632, opened 2022-09-28
ignore me

jwendell:


yxun:
@jwendell 
For comparison, when I change the tag from 2.3 to 2.2 , the api 2.3 job passed.
https://github.com/openshift/release/pull/32705

nvm, I see the istio 2.3 job passed here as well. must be something else caused the git issue before.

jwendell:
> nvm, I see the istio 2.3 job passed here as well. must be something else caused the git issue before.

It did not pass. I'm changing the test to debug the issue.

PR #637, opened 2022-10-04
OSSM-2000 Re-enable Wasm tests

dgn:
I fixed this by uploading the upstream builds of the Wasm filters to our bucket. That means we're not testing our own builds, but at least we're testing our proxy-wasm implementation with a large Wasm filter.

### DIFF
--- a/tests/integration/telemetry/stats/prometheus/wasm/stats_wasm_filter_test.go
+++ b/tests/integration/telemetry/stats/prometheus/wasm/stats_wasm_filter_test.go
@@ -39,8 +39,6 @@ func TestWasmStatsFilter(t *testing.T) {
 
 func TestMain(m *testing.M) {
 	framework.NewSuite(m).
-		// FIXME: https://issues.redhat.com/browse/OSSM-2000
-		Skip("https://github.com/istio/istio/issues/0").
 		Label(label.CustomSetup).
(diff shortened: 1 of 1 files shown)


jewertow:
Now I understand why it worked locally and was failing on prow.

jwendell:
If I understand it correctly, This workaround will only work until we change istio.deps with a new proxy sha, unless you perform the copy again. And again. And again.

dgn:
> If I understand it correctly, This workaround will only work until we change istio.deps with a new proxy sha, unless you perform the copy again. And again. And again.

Are we not planning on pushing the Wasm modules again in the future? In that case I can change the PR to download a fixed set of Wasm modules (the ones I use here). But I still prefer having the test run over skipping it, just in case we break Wasm support in envoy

jwendell:
Yes, Dario is close to have them working again. But until they're back we will have to maintain this by hand if we change istio.deps.

jewertow (approved):

jewertow:
@dgn did you see that the related [PR](https://github.com/maistra/proxy/pull/206) in maistra/proxy was merged?

dgn:
> @dgn did you see that the related [PR](https://github.com/maistra/proxy/pull/206) in maistra/proxy was merged?

No! Thanks for the heads-up. I verified that wasm extensions are being pushed to GCS again. Will rebase this and add `okay-to-merge`

PR #638, opened 2022-10-05
OSSM-2082 CNI installer now creates the net.d directory if necessary

luksa:


cni/pkg/install/kubeconfig.go
```diff
@@ -120,6 +120,11 @@ func createKubeconfigFile(cfg *config.InstallConfig, saToken string) (kubeconfig
 		return "", err
 	}
 
+	// When using Multus, the net.d dir might not exist yet, so we must create it
+	if err := os.MkdirAll(cfg.MountedCNINetDir, os.FileMode(0o755)); err != nil {
```
jewertow, line 124: Shouldn't we rather create this directory in the container image?

luksa: No, this is a directory on the host node. It's mounted in the container so that install-cni can install the cni plugin onto the node.

jewertow: Ok, I understand.

jewertow (approved):

yxun (approved):

luksa:
/test unit

Argh, that flaky StatusManager test.

jwendell:
@luksa Should we submit this upstream as well?

luksa:
This isn't needed in upstream, since they always install the plugin into the cluster's main net.d dir, which always exists.

jwendell:
Well, the `MountedCNINetDir` parameter is present on upstream and is configurable so that vendors like us can use it, and this PR makes it more robust, right? So, IMO we should do that as this will also benefit us in the future rebases. Upstream first policy :)

luksa:
Earlier versions did create the directory, so this was actually removed in upstream.

PR #642, opened 2022-10-07
OSSM-2074 Fix issues in the analysis controller

luksa:


pkg/config/analysis/incluster/controller.go
```diff
@@ -51,8 +51,20 @@ func NewController(stop <-chan struct{}, rwConfigStore model.ConfigStoreControll
 		"", resource.Namespace(namespace), func(name collection.Name) {}, true)
 	ia.AddSource(rwConfigStore)
 	// Filter out configs watched by rwConfigStore so we don't watch multiple times
+	schemas := collections.All.
+		Remove(rwConfigStore.Schemas().All()...)
+
+	// remove cluster-scoped resources if working in multi-tenant mode (with member roll controller)
+	if kubeClient.GetMemberRoll() != nil {
```
jwendell, line 58: When working on cluster-scoped mode, will this return nil?

luksa: Yes. We don't set the member roll in cluster-scoped mode.

jwendell (approved):

PR #645, opened 2022-10-10
OSSM-2109 Fix flaky IOR unit test

luksa:
The sleep in ensureNamespaceExists was hardcoded to 100ms, regardless of r.handleEventTimeout. This timeout during unit tests is only 1ms, so the 100ms sleep caused the for loop to only run once.

Here we change the duration of the sleep to be 1/100 of r.handleEventTimeout. This change preserves the production sleep time of 100ms, but reduces the sleep time in unit tests to 10μs. This makes ensureNamespaceExists() run the for loop multiple times before giving up, fixing the test's flakiness.

### DIFF
--- a/pilot/pkg/config/kube/ior/route.go
+++ b/pilot/pkg/config/kube/ior/route.go
@@ -277,7 +277,7 @@ func (r *route) ensureNamespaceExists(cfg config.Config) error {
 		default:
 			IORLog.Debugf("Namespace %s not found in SMMR, trying again", cfg.Namespace)
 		}
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(r.handleEventTimeout / 100)
 	}
 }
 


luksa:
/cherrypick-2.2
/cherrypick-2.3

yxun (approved):

oschaaf (approved):
LGTM - but I think that with these changes, now the pr description is a little out of sync.

luksa:
I've decided to only keep my original change, as the other changes considerably slowed down the test. This test will probably be rewritten anyway. 

Also, although my original change reduces the sleep to just 1/100.000s, this only applies to the test itself, not when the code runs in production (there, the sleep is .1s long). In the test, this leads to the for loop running for around 5-15 loops, so it's not a big deal. It's still better than having a tight loop with no sleep (in that case, it does around 400-500 loops on my machine).

oschaaf (approved):

luksa:
/test istio-unit-2-1

Flaky `TestStatusManager()` strikes again.

PR #649, opened 2022-10-11
OSSM-2123 gw-api deploymentcontroller should not set securityContext

dgn:
I'm just removing the whole `securityContext` block here because I think adding that is the job of gateway injection. I tested this on OCP 4.8 (k8s 1.21) and was able to create a Gateway resource with a listener on port 80- it's correctly created and running without privileges in the `restricted` SCC.

I believe ideally this would work similarly to the injectors, by using a ConfigMap that stores the template, instead of keeping it compiled into the binary. Might submit a PR upstream

### DIFF
--- a/pilot/pkg/config/kube/gateway/templates/deployment.yaml
+++ b/pilot/pkg/config/kube/gateway/templates/deployment.yaml
@@ -42,30 +42,6 @@ spec:
       containers:
       - image: auto
         name: istio-proxy
-        securityContext:
-        {{- if .KubeVersion122 }}
-          # Safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326
-          capabilities:
-            drop:
-            - ALL
-          allowPrivilegeEscalation: false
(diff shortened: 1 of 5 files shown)


jwendell (approved):
You also need to update the golden files for tests to pass

rcernich:
I think all of the gateway injection issues have been fixed in 2.2.  Is it possible that some cherry picks were missed?

rcernich:
- maistra/istio-operator#1020

dgn:
> I think all of the gateway injection issues have been fixed in 2.2. Is it possible that some cherry picks were missed?

Yes, those are in. This is specifically about the Gateway API Deployment Controller (which uses Gateway Injection in the background)

jewertow:
Will it work on OCP 4.11 (k8s 1.25) with new Pod Security Standards and Pod Security Admission? I think we should test it.

jewertow (approved):

PR #652, opened 2022-10-14
OSSM-1689 Simplify IOR

yannuil:
Signed-off-by: Yann Liu <yannliu@redhat.com>

The IOR will use `crdclient.Client`, which implements a `model.ConfigStoreCache` interface. It will use the underlying `Gateway` kind informer to trigger events on `Gateway`. IOR will only response to these events. When IOR attaches the handler, the informer will trigger `ADD` to let IOR catch up the state. When setting a new namespace, the informer will also trigger `DELETE` on the removed namespace and `ADD` on the new namespace.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/ior/route.go
+++ b/pilot/pkg/config/kube/ior/route.go
@@ -366,7 +366,7 @@ func (r *route) createRoute(metadata config.Meta, originalHost string, tls *netw
 
 	nr, err := r.routerClient.Routes(serviceNamespace).Create(context.TODO(), &v1.Route{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        getRouteName(metadata.Namespace, metadata.Name, actualHost),
+			Name:        getRouteName(metadata.Namespace, metadata.Name, originalHost),
 			Namespace:   serviceNamespace,
 			Labels:      labels,
 			Annotations: annotations,
@@ -437,8 +437,8 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
 		gwSelector.String(), namespaces)
 }
 
-func getRouteName(namespace, name, actualHost string) string {
-	return fmt.Sprintf("%s-%s-%s", namespace, name, hostHash(actualHost))
+func getRouteName(namespace, name, host string) string {
+	return fmt.Sprintf("%s-%s-%s", namespace, name, hostHash(host))
 }
 
 // getActualHost returns the actual hostname to be used in the route
--- a/pilot/pkg/config/kube/ior/route.go
+++ b/pilot/pkg/config/kube/ior/route.go
@@ -26,6 +26,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	v1 "github.com/openshift/api/route/v1"
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -56,7 +57,7 @@ const (
 type syncRoutes struct {
 	metadata config.Meta
 	gateway  *networking.Gateway
-	routes   []*v1.Route
+	routes   map[string]*v1.Route
 }
 
 // route manages the integration between Istio Gateways and OpenShift Routes
@@ -136,6 +137,7 @@ func (r *route) addNewSyncRoute(cfg config.Config) *syncRoutes {
 	syncRoute := &syncRoutes{
 		metadata: cfg.Meta,
 		gateway:  gw,
+		routes:   make(map[string]*v1.Route),
 	}
 
 	r.gatewayMap[gatewayMapKey(cfg.Namespace, cfg.Name)] = syncRoute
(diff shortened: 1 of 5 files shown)


pilot/pkg/config/kube/ior/route.go
```diff
@@ -625,3 +553,52 @@ func hostHash(name string) string {
...
+		IORLog.Errorf("could not decode object as Gateway. Object = %v", curr)
+		return
+	}
+
+	debugMessage := fmt.Sprintf("Event %v arrived:", event)
+	if event == model.EventUpdate {
+		debugMessage += fmt.Sprintf("\tOld object: %v", old)
+	}
+	debugMessage += fmt.Sprintf("\tNew object: %v", curr)
+	IORLog.Debug(debugMessage)
```
luksa:
```suggestion
	if IORLog.GetOutputLevel() >= log.DebugLevel {
		debugMessage := fmt.Sprintf("Event %v arrived:", event)
		if event == model.EventUpdate {
			debugMessage += fmt.Sprintf("\tOld object: %v", old)
		}
		debugMessage += fmt.Sprintf("\tNew object: %v", curr)
		IORLog.Debug(debugMessage)
	}
```

luksa:
@yannuil Please rebase this to 2.3.

luksa:
Note: I've tested this a bit, and I don't think it works as it should.

I saw the following error in the istiod log:
```
$ istiod-log | grep ior                                                                                                                                   
2022-10-17T10:46:40.782619Z     info    klog    attempting to acquire leader lease istio-system/ior-leader...
2022-10-17T10:46:50.191436Z     info    klog    successfully acquired lease istio-system/ior-leader
2022-10-17T10:46:50.191628Z     info    leader election lock obtained: ior-leader
2022-10-17T10:46:50.191660Z     info    ior     setting up IOR
2022-10-17T10:46:50.242077Z     info    smmr    Cache synced for listener "ior"
2022-10-17T10:47:29.096767Z     error   ior     gateway bookinfo/bookinfo-gateway does not exists, IOR failed to update  <-- THIS
2022-10-17T10:49:38.524773Z     error   ior     gateway bookinfo/bookinfo-gateway does not exists, IOR failed to update
2022-10-17T10:57:21.103023Z     info    ior     This pod is no longer a leader. IOR stopped responding
2022-10-17T10:57:21.117225Z     info    leader election lock lost: ior-leader
```

The gateway did exist. I edited the gateway object and saw the same error again (second error in the log). I then restarted istiod several times and didn't see the error again. 

I'll do more testing tomorrow.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -387,64 +323,24 @@ func (r *route) handleEvent(event model.Event, cfg config.Config) error {
...
-		var result *multierror.Error
-		result = multierror.Append(result, r.handleDel(cfg))
-		result = multierror.Append(result, r.handleAdd(cfg))
-		return result.ErrorOrNil()
+		return r.onGatewayUpdated(cfg)
 
 	case model.EventDelete:
-		return r.handleDel(cfg)
+		return r.onGatewayRemoved(cfg)
 	}
```
luksa: Are we sure that we always get the correct event here? For example, is it possible to get an ADD event when the Gateway already exists? 

I'm concerned that after these changes, IOR is still edge-triggered, whereas controllers in Kubernetes should be level-triggered. IMO all three events should invoke the same function ("reconcileGateway"). This function should be idempotent and only consider the current state of the Gateway object (not the change in state as it does now). The controller would be much more robust.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -290,27 +182,91 @@ func (r *route) handleAdd(cfg config.Config) error {
 	r.gatewaysLock.Lock()
 	defer r.gatewaysLock.Unlock()
 
-	if _, ok := r.gatewaysMap[gatewaysMapKey(cfg.Namespace, cfg.Name)]; ok {
+	if _, ok := r.gatewayMap[gatewayMapKey(cfg.Namespace, cfg.Name)]; ok {
 		IORLog.Infof("gateway %s/%s already exists, not creating route(s) for it", cfg.Namespace, cfg.Name)
 		return nil
 	}
```
luksa: In this function, we're handling the addition of a Gateway. Here, we check if the Gateway already exists. If this is indeed possible, this means that whatever code called this function assumes that an ADD event can only be received when the Gateway object is created. This code here indicates that that's not true. 

Also, if `onGatewayAdded` gets called and find that the Route for the Gateway already exists, it simply does nothing. What if the Route's configuration doesn't match the Gateway? IOR should update the Route if necessary. 

The same applies to the other `onGateway*` functions. These three functions should be a single function that updates the Route regardless of what caused the function to be invoked (either an ADD, UPDATE, or DELETE event).

pilot/pkg/config/kube/ior/route.go
yannuil: This is the [queue](https://github.com/maistra/istio/blob/maistra-2.2/pkg/queue/instance.go#L100) implementation IOR relying on. The queue is populated via an [informer](https://github.com/maistra/istio/blob/maistra-2.2/pilot/pkg/config/kube/crdclient/cache_handler.go#L86) of `Gateway` kind. The queue will guarantee the order of the event for us.

For an ADD event, this is triggered after the Gateway gets recreated. The error message showed in the log is related to that  house keeping map in the IOR for us to handle deletion. The error tells that the CREATE event is not there yet but it sees an UPDATE.

I agree on the idempotent. I am making some changes, so it will discard event if it is outdated.

yannuil:
Close in favor of #673

PR #653, opened 2022-10-14
OSSM-1093 Shorten exported resource name

bmangoen:

This will fix the creation of exported resource with services/namespaces long names

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


pkg/servicemesh/federation/server/routing.go
```diff
@@ -47,8 +51,25 @@ var (
...
+	return strings.Replace(name, "-", "", -1)
+}
+
 func createResourceName(mesh string, source federationmodel.ServiceKey) string {
-	return fmt.Sprintf("federation-exports-%s-%s-%s", mesh, source.Name, source.Namespace)
+	resourceName := fmt.Sprintf("federation-exports-%s-%s-%s", mesh, source.Name, source.Namespace)
+	if !labels.IsDNS1123Label(resourceName) {
+		resourceName = fmt.Sprintf("federation-exports-%s-%s-%s", shortenName(mesh), shortenName(source.Name), shortenName(source.Namespace))
+		if !labels.IsDNS1123Label(resourceName) {
+			resourceName = fmt.Sprintf("%s-%s", resourceName[:52], hashResourceName(resourceName, 5))
```
yxun: what's the 52 definition there ? Is that a hard-coded slice length ? It's better to add a comment with an example of that slice.
If it's a hard-coded length, we may need to define a constant variable.

luksa: I'd change this code:
- if too long, truncate to 55 characters + add 8 characters of hash
- I don't see a good reason to remove dashes from the mesh, name, or namespace
- shorten `"federation-exports-"` to `"fed-exp-"`
- change order from: `mesh-name-namespace` to `name-namespace-mesh` so that the name is truncated last
- use constants, as @yxun suggested (e.g. `maxNameLength`, `hashLength`, and use `[:maxNameLength-hashLength]` to truncate)

Please also add a unit test for the `createResourceName` function showing examples of how the function truncates the name. 

You must also add an equivalent change to https://github.com/Maistra/istio/blob/403a6b63c620b8b7c2376f3a24b17022acbc0509/pilot/pkg/serviceregistry/federation/routing.go#L36-L38, otherwise you didn't fix the issue, as it's just going to break on the import side.

In fact, there should be no duplication between the two functions (make the existing two functions call a new function that takes the prefix (`fed-exp` or `fed-imp`) and generates the correct formatted name.

pkg/servicemesh/federation/common/util.go
```diff
@@ -15,17 +15,25 @@
...
 	kubelabels "k8s.io/apimachinery/pkg/labels"
 	v1 "maistra.io/api/federation/v1"
 
+	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/spiffe"
 )
 
+var (
+	maxNameLength = 63
```
luksa: You can remove this constant and use `labels.DNS1123LabelMaxLength` instead.

bmangoen: There is this [comment](https://github.com/maistra/istio/blob/maistra-2.3/pkg/config/labels/instance.go#L27) for the usage of `DNS1123LabelMaxLength` variable. Does it mean for testing in the condition blocks or for unit tests ?

pkg/servicemesh/federation/common/util.go
```diff
@@ -94,3 +102,20 @@ func RemoteChecksum(remote v1.ServiceMeshPeerRemote) uint64 {
...
+	return (hex.EncodeToString(hash[:]))[:n]
+}
+
+// ProcessResourceName returns the imported/exported resource name
+// checking if the length of the formatted name < maxNameLength
+// If not returns the truncated formatted name suffixed by its hash part
+func ProcessResourceName(prefix, meshName, namespace, serviceName string) string {
+	resourceName := fmt.Sprintf("%s-%s-%s-%s", prefix, serviceName, namespace, meshName)
+	if !labels.IsDNS1123Label(resourceName) {
+		return fmt.Sprintf("%s%s", resourceName[:maxNameLength-hashLength], hashResourceName(resourceName, hashLength))
```
luksa: Why use `Sprintf`, when you can just concatenate two strings with `+`?

pkg/servicemesh/federation/server/routing_test.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, tc := range testCases {
+		resourceName := createResourceName(tc.mesh.String(), tc.source)
+		if len(resourceName) > labels.DNS1123LabelMaxLength {
+			t.Fatalf("%s\n%s length: %d, should be < 63", tc.name,
+				resourceName, len(resourceName))
+		}
+		if resourceName != tc.expectedResourceName {
+			t.Fatalf("%s\n%s not equals to the expected resource name %s", tc.name,
+				resourceName, tc.expectedResourceName)
+		}
```
luksa: Wrap this in
```		
t.Run(tc.name, func(t *testing.T) {
   ...
}
```
This way, you don't have to pass `tc.name` to `Fatalf` and it's easier to see which cases succeeded and which failed.

luksa: You might also want to use `labels.IsDNS1123Label()` instead of only checking the length.

pkg/servicemesh/federation/common/util.go
luksa: I think they just meant that the const isn't used outside the file (except in tests). But we can definitely use it instead of creating our own copy.

pkg/servicemesh/federation/server/routing_test.go
```diff
@@ -0,0 +1,89 @@
...
+			expectedResourceName: "fed-exp-productpage-rest-private-bookinfo-mesh",
+		},
+		{
+			name: "long mesh name + long namespace + long service name => truncated + hashed",
+			mesh: "test-remote",
+			source: federationmodel.ServiceKey{
+				Namespace: "bookinfo-production",
+				Name:      "productpage-rest-private",
+			},
+			expectedResourceName: "fed-exp-productpage-rest-private-bookinfo-production-teefc2b162",
```
yxun: Could we call a hash function in this test case instead of hard coding the hash value "teefc2b162" ?

luksa: Wouldn't that just make the test less readable for no benefit? I mean, you'd have to concatenate the full name in the exact same way as the code does it.

yxun: well, if we assume this value is fixed and always generated by using hash length 8, then that's fine here. 
If we need to update a hash length value or the implementation of the hashResourceName in the future, this test case is not flexible to test a hash length 9 case.

pkg/servicemesh/federation/server/routing_test.go
```diff
@@ -0,0 +1,89 @@
...
+			expectedResourceName: "fed-exp-productpage-bookinfo-mesh",
+		},
+		{
+			name: "long mesh name",
+			mesh: "myveryveryverylongmesh",
+			source: federationmodel.ServiceKey{
+				Namespace: "bookinfo",
+				Name:      "productpage",
+			},
+			expectedResourceName: "fed-exp-productpage-bookinfo-myveryveryverylongmesh",
```
yxun: Funny names : )
Should we test random string of a fixed length name cases ?

pkg/servicemesh/federation/common/util.go
```diff
@@ -94,3 +102,20 @@ func RemoteChecksum(remote v1.ServiceMeshPeerRemote) uint64 {
...
+// hashResourceName applies a sha256 on the host and truncates it to the first n char
+func hashResourceName(name string, n int) string {
+	hash := sha256.Sum256([]byte(name))
+	return (hex.EncodeToString(hash[:]))[:n]
+}
+
+// ProcessResourceName returns the imported/exported resource name
+// checking if the length of the formatted name < maxNameLength
+// If not returns the truncated formatted name suffixed by its hash part
+func ProcessResourceName(prefix, meshName, namespace, serviceName string) string {
```
luksa: Can you rename this to `formatResourceName` or something similar?

pkg/servicemesh/federation/server/routing_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"testing"
+
+	"istio.io/istio/pkg/cluster"
+	"istio.io/istio/pkg/config/labels"
+	federationmodel "istio.io/istio/pkg/servicemesh/federation/model"
+)
+
+func TestCreateResourceName(t *testing.T) {
```
luksa: It's probably better to have just one test that tests the `ProcessResourceName` function instead of having duplicate tests for import and export.

pkg/servicemesh/federation/server/routing_test.go
bmangoen: I guess the main purpose of these tests is to have clear examples of how the function formats the resource name. Wouldn't adding and using a generated random string be to format the expected result `expectedResourceName` with it as the function would do ?

pkg/servicemesh/federation/server/routing_test.go
luksa: I've just thought of an alternative approach. What if `expectedResourceName` was a RegExp. In the example above, the regexp would be: `"fed-exp-productpage-rest-private-bookinfo-production-te[a-f0-9]{8}"` indicating that we only care about the part that was truncated and that the part at the end is an 8-character hex value, but we don't care about the actual value.

But then again, this would just make the test more complicated. Someone seeing the test for the first time would have a harder time understanding the test compared to what the test looks like now. That's why I think the current form is perfectly fine.

pkg/servicemesh/federation/server/routing_test.go
yxun: ok, I see now. It's fine with those cases now.

yxun: Could we log the length of a resource name if a test case fails ?
For example, name: "long mesh name", mesh: "myveryveryverylongmesh", if there is  anything failed, log the length of an input string name + mesh + ...

pkg/servicemesh/federation/common/util.go
```diff
@@ -94,3 +101,20 @@ func RemoteChecksum(remote v1.ServiceMeshPeerRemote) uint64 {
 	}
 	return checksum
 }
+
+// hashResourceName applies a sha256 on the host and truncates it to the first n char
+func hashResourceName(name string, n int) string {
+	hash := sha256.Sum256([]byte(name))
+	return (hex.EncodeToString(hash[:]))[:n]
```
luksa, line 106: Is the `[:]` necessary?

pkg/servicemesh/federation/common/util.go
```diff
@@ -15,17 +15,24 @@
...
 	discoveryv1 "k8s.io/api/discovery/v1"
 	kubelabels "k8s.io/apimachinery/pkg/labels"
 	v1 "maistra.io/api/federation/v1"
 
+	"istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/spiffe"
 )
 
+var (
```
luksa: const?

pkg/servicemesh/federation/common/util_test.go
```diff
@@ -0,0 +1,88 @@
...
+			},
+			expectedResourceName: "fed-imp-productpage-rest-private-bookinfo-production-te160043b8",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resourceName := FormatResourceName(tc.prefix, tc.mesh.String(),
+				tc.source.Namespace, tc.source.Name)
+			if len(resourceName) > labels.DNS1123LabelMaxLength {
+				t.Fatalf("%s length: %d, should be < 63",
```
luksa: <=

luksa: Actually, you should use IsDNS11234Label() instead of checking just the length.

pkg/servicemesh/federation/common/util.go
bmangoen: It is the way for converting the `[32]byte` result from the `Sum256()` to the `[]byte` required by the `EncodeToString()`

pkg/servicemesh/federation/common/util_test.go
```diff
@@ -0,0 +1,88 @@
...
+			expectedResourceName: "fed-imp-productpage-rest-private-bookinfo-production-te160043b8",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resourceName := FormatResourceName(tc.prefix, tc.mesh.String(),
+				tc.source.Namespace, tc.source.Name)
+			if !labels.IsDNS1123Label(resourceName) {
+				t.Fatalf("%s length: %d, should be <= 63",
+					resourceName, len(resourceName))
```
luksa: Change the error message. Should be something like `Not a valid RFC 1123 label`, but you can also append the length and max length to the message, since that will most likely be the cause of the error.

luksa (approved):

PR #657, opened 2022-10-19
OSSM-755: Fix TLS certificates used in security tests

jewertow:
I figured out that certificates used in `security.egress.tls.*` and `security.peer.file-mounted-certs` tests don't pass verification when OpenSSL 1.1.1k and higher is used. For example, those certificates pass verification on Ubuntu 20.04 which delivers `OpenSSL 1.1.1f  31 Mar 2020`, but do not pass on CentOS Stream 8 or RHEL 8 which deliver `OpenSSL 1.1.1k  FIPS 25 Mar 2021`.
I noticed that root-cert.pem and cert-chain.pem have specified the same common name, so as a workaround I added SAN as a prefix to CN in the cert-chain.pem. Without this change, a TLS handshake fail with an alert message "UNKNOWN CA (48)".


luksa (approved):

jwendell:
Good catch!

jewertow:
I regenerated all certificates for the sake of consistency, because all of them use the same root-cert.pem.

jwendell:
Does it make sens to submit this upstream as well, so that we reduce our differences on next rebase?

I'd say it's useful as it fixes an issue when running in a different environment (newer openssl) and doesn't break any existent behavior.

jewertow:
Yes, I submitted a [PR](https://github.com/istio/istio/pull/41566).

PR #660, opened 2022-10-20
OSSM-2083 Update Istio to 1.14.5

dgn:
Depends on https://github.com/maistra/xns-informer/pull/28


manifests/charts/gateways/istio-egress/templates/_affinity.tpl
```diff
@@ -11,6 +11,7 @@ nodeAffinity:
 {{- define "nodeAffinityRequiredDuringScheduling" }}
       nodeSelectorTerms:
       - matchExpressions:
+        {{- if .global.arch }}
```
jewertow, line 14: This change comes from upstream [bug fix](https://github.com/maistra/istio/commit/361ab420c06ea833b05d50367375cc16d7ef98bc) which overlaps with [our change](https://github.com/maistra/istio/commit/f418472a1c3f5aa8efc06859df18da8530458995) of this template, so we could revert our commit before merging this PR to avoid confusion during the next rebase.

jewertow: After a short reflection, I think we can ignore it in this release. I will just prepare a note for the next rebase, because it's not the only one commit to skip.

dgn: yeah, tbh I had to make a bunch of changes to make things work again especially in the gateway API controller. It's all in the merge commit, so we shouldn't miss that one during the rebase

jwendell (approved):

jewertow (approved):

PR #663, opened 2022-10-21
OSSM-2147 Initialize analysis client with the actual revision

luksa:


jewertow (approved):

jewertow:
@luksa are you going to fix it in istio/istio as well?

luksa:
Are you able to reproduce this upstream?

pkg/config/analysis/incluster/controller.go
```diff
@@ -63,7 +63,7 @@ func NewController(stop <-chan struct{}, rwConfigStore model.ConfigStoreControll
 			Remove(collections.K8SApiextensionsK8SIoV1Customresourcedefinitions)
 	}
 
-	store, err := crdclient.NewForSchemas(kubeClient, "default",
```
dgn, line 66: this hard-coded revision definitely looks like upstream is affected, too. except if they only ever want this to work on the default revision?

dgn (approved):
good catch!

PR #664, opened 2022-10-25
Add delete configmap privilege to fix distribution report cleanup

luksa:
This fixes the following error during istiod shutdown when PILOT_ENABLE_STATUS=true:

error   status  failed to properly clean up distribution report: configmaps "istiod-649887579d-rck66-distribution" is forbidden: User "system:serviceaccount:istio-system:istiod" cannot delete resource "configmaps" in API group "" in the namespace "istio-system"


jewertow:
Since distribution reports are created in the same namespace as istiod is deployed, couldn't this permission be added to istiod role instead of cluster role?

dgn:
Strange that this permission doesn't exist upstream- it likely affects them, too

luksa:
@jewertow Good catch. I'll update the PR. I've also submitted this upstream: https://github.com/istio/istio/pull/41621

jewertow (approved):

PR #665, opened 2022-10-25
OSSM-2179: changes the restriction about the number of layers in a WASM plugin.

jwendell:
Manual cherry-pick of upstream commit 9a2359d8f08be06ee5f854b30e44da3523992e41

The attempt to backport it to upstream 1.14 branch was not successful: https://github.com/istio/istio/pull/41550

### DIFF
--- a/pkg/wasm/imagefetcher.go
+++ b/pkg/wasm/imagefetcher.go
@@ -20,6 +20,7 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


dgn:
Oh wow, this is super confusing. This doesn't actually properly support multiple layers- `plugin.wasm` must be in the *last layer* otherwise it won't be found. This is really just a tiny fix to allow for container labels in the first layer. The release note is wrong imo (but that doesn't mean we should change it, we should behave the same way upstream behaves)

dgn (approved):

PR #670, opened 2022-11-01
OSSM-728 Configuration scripts for Federation on Z and P, and bare metal

dgn:
Let's finally merge the config examples created by @cfilleke. This also moves our federation examples to the `./samples` directory where they belong.


jwendell:
Worth to wait for 2.4?

dgn:
> Worth to wait for 2.4?

Why wait? It's just example config, no user impact

jwendell:
> > Worth to wait for 2.4?
> 
> Why wait? It's just example config, no user impact

One less commit to cherry-pick to 2.4 :) But I don't have a strong opinion. I'm fine either way.

jewertow (approved):

dgn:
btw I double-checked with QE to make sure nobody's referencing these scripts. I don't want to break anyone by moving them around

PR #673, opened 2022-11-03
OSSM-1689 Simplify IOR

yannuil:
Signed-off-by: Yann Liu <yannliu@redhat.com>**Please provide a description of this PR:**

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/ior/fake.go
+++ b/pilot/pkg/config/kube/ior/fake.go
@@ -24,6 +24,7 @@ import (
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"golang.org/x/net/context"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/rest"
@@ -34,20 +35,33 @@ import (
 
 // FakeRouter implements routev1.RouteInterface
 type FakeRouter struct {
+	namespace      string
 	routes         map[string]*v1.Route
 	routesLock     sync.Mutex
-	callCounts     map[string]int
-	callCountsLock sync.Mutex
+	callCounts     *map[string]int
+	callCountsLock *sync.Mutex
+}
+
+type FakeAggregateRouter struct {
+	FakeRouter
+	routersByNamespace map[string]*FakeRouter
 }
 
 // FakeRouterClient implements routev1.RouteV1Interface
 type FakeRouterClient struct {
-	routesByNamespace     map[string]routev1.RouteInterface
-	routesByNamespaceLock sync.Mutex
+	routersByNamespace     map[string]*FakeRouter
+	routersByNamespaceLock sync.Mutex
+	callCounts             *map[string]int
+	callCountsLock         sync.Mutex
 }
 
 type fakeKubeClient struct {
 	client kube.Client
+}
+
+type FakeRouterInterface interface {
+	routev1.RouteInterface
+	getRoutes()
 }
 
 // NewFakeKubeClient creates a new FakeKubeClient
@@ -69,17 +83,63 @@ func (c *fakeKubeClient) GetHandleEventTimeout() time.Duration {
 
 // NewFakeRouterClient creates a new FakeRouterClient
 func NewFakeRouterClient() routev1.RouteV1Interface {
+	cc := make(map[string]int)
+
 	return &FakeRouterClient{
-		routesByNamespace: make(map[string]routev1.RouteInterface),
+		routersByNamespace: make(map[string]*FakeRouter),
+		callCounts:         &cc,
 	}
 }
 
 // NewFakeRouter creates a new FakeRouter
-func NewFakeRouter() routev1.RouteInterface {
+func NewFakeRouter(namespace string, callCounts *map[string]int, callCountsLock *sync.Mutex) *FakeRouter {
 	return &FakeRouter{
-		routes:     make(map[string]*v1.Route),
-		callCounts: make(map[string]int),
+		namespace:      namespace,
+		routes:         make(map[string]*v1.Route),
+		callCounts:     callCounts,
+		callCountsLock: callCountsLock,
 	}
+}
+
+func NewFakeAggregateRouter(routersByNamespace map[string]*FakeRouter, callCounts *map[string]int, callCountsLock *sync.Mutex) *FakeAggregateRouter {
+	fk := &FakeAggregateRouter{
+		FakeRouter{
+			namespace:      "",
+			routes:         nil,
+			callCounts:     callCounts,
+			callCountsLock: callCountsLock,
+		},
+		routersByNamespace,
+	}
+
+	return fk
+}
+
+func (fk *FakeRouter) getRoutes() map[string]*v1.Route {
+	fk.routesLock.Lock()
+	defer fk.routesLock.Unlock()
+
+	res := make(map[string]*v1.Route)
+
+	for n, r := range fk.routes {
+		res[fmt.Sprintf("%s/%s", fk.namespace, n)] = r
+	}
+
+	return res
+}
+
+func (fk *FakeAggregateRouter) getRoutes() map[string]*v1.Route {
+	res := make(map[string]*v1.Route)
+
+	for _, r := range fk.routersByNamespace {
+		routes := r.getRoutes()
+
+		for k, v := range routes {
+			res[k] = v
+		}
+	}
+
+	return res
 }
 
 // RESTClient implements routev1.RouteV1Interface
@@ -89,27 +149,37 @@ func (rc *FakeRouterClient) RESTClient() rest.Interface {
 
 // Routes implements routev1.RouteV1Interface
 func (rc *FakeRouterClient) Routes(namespace string) routev1.RouteInterface {
-	rc.routesByNamespaceLock.Lock()
-	defer rc.routesByNamespaceLock.Unlock()
+	rc.routersByNamespaceLock.Lock()
+	defer rc.routersByNamespaceLock.Unlock()
 
-	if _, ok := rc.routesByNamespace[namespace]; !ok {
-		rc.routesByNamespace[namespace] = NewFakeRouter()
+	if namespace == metav1.NamespaceAll {
+		return NewFakeAggregateRouter(rc.routersByNamespace, rc.callCounts, &rc.callCountsLock)
 	}
 
-	return rc.routesByNamespace[namespace]
+	router, ok := rc.routersByNamespace[namespace]
+
+	if !ok {
+		router = NewFakeRouter(namespace, rc.callCounts, &rc.callCountsLock)
+		rc.routersByNamespace[namespace] = router
+	}
+
+	return router
 }
 
 func (fk *FakeRouter) GetCallCount(functionName string) int {
 	fk.callCountsLock.Lock()
 	defer fk.callCountsLock.Unlock()
 
-	return fk.callCounts[functionName]
+	cc := *fk.callCounts
+
+	return cc[functionName]
 }
 
 func (fk *FakeRouter) incrementCallCount(functionName string) int {
 	fk.callCountsLock.Lock()
-	fk.callCounts[functionName]++
-	num := fk.callCounts[functionName]
+	cc := *fk.callCounts
+	cc[functionName]++
+	num := cc[functionName]
 	fk.callCountsLock.Unlock()
 
 	return num
@@ -121,11 +191,27 @@ func (fk *FakeRouter) generateHost() string {
 	return fmt.Sprintf("generated-host%d.com", num)
 }
 
-// Create implements routev1.RouteInterface
-func (fk *FakeRouter) Create(ctx context.Context, route *v1.Route, opts metav1.CreateOptions) (*v1.Route, error) {
+func (fk *FakeRouter) getRoute(name string) *v1.Route {
 	fk.routesLock.Lock()
 	defer fk.routesLock.Unlock()
 
+	return fk.routes[name]
+}
+
+func (fk *FakeRouter) setRoute(name string, route *v1.Route) error {
+	fk.routesLock.Lock()
+	if route == nil {
+		delete(fk.routes, name)
+	} else {
+		fk.routes[name] = route
+	}
+	fk.routesLock.Unlock()
+
+	return nil
+}
+
+// Create implements routev1.RouteInterface
+func (fk *FakeRouter) Create(ctx context.Context, route *v1.Route, opts metav1.CreateOptions) (*v1.Route, error) {
 	if strings.Contains(route.Spec.Host, "/") {
 		return nil, fmt.Errorf("invalid hostname")
 	}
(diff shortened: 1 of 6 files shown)


luksa:
@yannuil there are one or two legit lint failures (see https://storage.googleapis.com/origin-ci-test/pr-logs/pull/maistra_istio/673/pull-ci-maistra-istio-maistra-2.3-istio-lint-2-3/1588454856429932544/build-log.txt)

pilot/pkg/config/kube/ior/fake.go
```diff
@@ -121,7 +121,26 @@ func (fk *FakeRouter) Create(ctx context.Context, route *v1.Route, opts metav1.C
 
 // Update implements routev1.RouteInterface
 func (fk *FakeRouter) Update(ctx context.Context, route *v1.Route, opts metav1.UpdateOptions) (*v1.Route, error) {
-	panic("not implemented")
+	fk.routesLock.Lock()
+	defer fk.routesLock.Unlock()
+
+	if _, ok := fk.routes[route.Name]; !ok {
+		return nil, fmt.Errorf("exisiting not found")
```
luksa:
```suggestion
		return nil, fmt.Errorf("existing route not found")
```

pilot/pkg/config/kube/ior/route.go
```diff
@@ -104,10 +89,9 @@ func newRoute(
...
 	mrc controller.MemberRollController,
 	stop <-chan struct{},
-) *route {
-	for !kubeClient.IsRouteSupported() {
-		IORLog.Infof("routes are not supported in this cluster; waiting for Route resource to become available...")
-		time.Sleep(10 * time.Second)
+) (*route, error) {
+	if !kubeClient.IsRouteSupported() {
+		return nil, fmt.Errorf("routes are not supported in this cluster")
 	}
```
luksa: Are you handling this elsewhere now? The purpose of the for loop was for the function to wait until the Route resource appears in the cluster. Previously, the function would return an error instead of waiting, but this was problematic during cold-starts of the cluster (see OSSM-1301).

pilot/pkg/config/kube/ior/route.go
```diff
@@ -555,36 +290,41 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
 
 	for _, ns := range namespaces {
```
luksa: I think you can get rid of this for loop and the `namespaces` list altogether. Instead, list the pods via the informer/lister (xns-informer will make sure you're listing pods only in the member namespaces). 

If you do this, IOR no longer needs to be a `MemberRollListener`.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -555,36 +290,41 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
...
+				svcSelector := labels.SelectorFromSet(service.Spec.Selector)
 
-			for _, svc := range svcList.Items {
-				svcSelector := labels.SelectorFromSet(svc.Spec.Selector)
+				IORLog.Debugf("matching service selector %s against %s")
 				if svcSelector.Matches(podLabels) {
-					return ns, svc.Name, nil
+					return pod.Namespace, service.Name, nil
 				}
+
```
luksa: unnecessary empty line

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +367,155 @@ func hostHash(name string) string {
...
+		IORLog.Debug(debugMessage)
+	}
+
+	var (
+		err       error
+		isManaged bool
+	)
+
+	isManaged, err = isManagedByIOR(*curr)
```
luksa:
```suggestion
	isManaged, err := isManagedByIOR(*curr)
```

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +367,155 @@ func hostHash(name string) string {
...
+	config := r.store.Get(collections.IstioNetworkingV1Alpha3Gateways.Resource().GroupVersionKind(), curr.Name, curr.Namespace)
+
+	// Stop early
+	if config != nil && curr.ResourceVersion < config.ResourceVersion {
+		return nil
+	}
+
+	var routes *v1.RouteList
+
+	routes, err = r.findRoutes(curr.Meta)
```
luksa, line 428:
```suggestion
	routes, err := r.findRoutes(curr.Meta)
```

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +367,155 @@ func hostHash(name string) string {
...
+
+	if !isManaged {
+		IORLog.Debugf("skipped processing routes for gateway %s/%s, as it is annotated by user", curr.Name, curr.Namespace)
+		return nil
+	}
+
+	config := r.store.Get(collections.IstioNetworkingV1Alpha3Gateways.Resource().GroupVersionKind(), curr.Name, curr.Namespace)
+
+	// Stop early
+	if config != nil && curr.ResourceVersion < config.ResourceVersion {
```
luksa: ResourceVersion is "opaque", meaning that you shouldn't depend on the (current) format of its contents. You're only allowed to check if two resourceVersions are equal or not (in other words: you can do `resourceVersion1 == resourceVersion2` and nothing else).

Yes, currently, this is an integer, but this is an implementation detail of etcd AFAIK. It might change in the future. Also, since resourceVersion is a string, comparing them with `<` doesn't even work correctly (e.g. `"12" < "2"`).

luksa:
@yannuil I'm testing this and it looks like Gateway deletion isn't handled. When I delete a Gateway, the associated Route is not deleted. If I then create the same Gateway again, I get an "already exists" error when IOR tries to create the Route.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -531,19 +214,71 @@ func (r *route) createRoute(metadata config.Meta, gateway *networking.Gateway, o
...
 
 	return nr, nil
 }
 
+func (r *route) findRoutes(metadata config.Meta) (*v1.RouteList, error) {
+	defaultLabelSet := getDefaultRouteLabelMap(metadata.Name, metadata.Namespace)
+
+	labels := labels.SelectorFromSet(defaultLabelSet)
+
+	return r.routerClient.Routes(metadata.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.String()})
```
luksa, line 260: This is looking for Routes only in the namespace of the Gateway, but in my case, the Route is in `istio-system`.

luksa:
Another problem: when I restart istiod, it doesn't reconcile any Gateways. Any changes performed while istiod was offline are thus ignored until the Gateway is touched.

EDIT: actually, sometimes they are reconciled, but not always. Looks like a race condition.

EDIT2: this might be related to IOR being a MemberRollListener. In one of my comments, I suggested making IOR no longer a MemberRollListener, so maybe this issue will fix itself if you do that.

pilot/pkg/config/kube/ior/route.go
luksa: Actually, I think you shouldn't fetch the latest config at all. Just use `curr`.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -555,36 +290,41 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
 
 	for _, ns := range namespaces {
 		// Get the list of pods that match the gateway selector
-		podList, err := r.kubeClient.CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: gwSelector.String()})
+		pods, err := r.kubeClient.CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: gwSelector.String()})
```
luksa: This performs an actual GET on the API server instead of reading from the cache.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +367,155 @@ func hostHash(name string) string {
...
+				result = multierror.Append(result, err)
+			} else {
+				delete(routeMap, route.Name)
+			}
+		}
+	}
+
+	for k, v := range routeMap {
+		IORLog.Debugf("clean up route %s for host %s", k, getHost(v))
+		if err := r.deleteRoute(&v); err != nil {
```
luksa, line 395: This line deletes the route if the update of the route fails (in line 406).

yannuil:
> Another problem: when I restart istiod, it doesn't reconcile any Gateways. Any changes performed while istiod was offline are thus ignored until the Gateway is touched.
> 
> EDIT: actually, sometimes they are reconciled, but not always. Looks like a race condition.
> 
> EDIT2: this might be related to IOR being a MemberRollListener. In one of my comments, I suggested making IOR no longer a MemberRollListener, so maybe this issue will fix itself if you do that.

@luksa Yes I agree. It is because `SetNamespace`. Using it will create this synchronization issue. IOR starts processing the `Gateway` before `SetNamespace` gets called with the current state. Initially I removed the use of the `namespaces`, but I failed to use the `Lister`. During the test runs, the `Lister` always returns empty result. Here is how I create the `Lister`. 
```go
	r.podLister = kubeClient.GetActualClient().KubeInformer().Core().V1().Pods().Lister()
	r.serviceLister = kubeClient.GetActualClient().KubeInformer().Core().V1().Services().Lister()
```
I could not figure out why the result is always empty in the test. It seems that the cache thinks its filled. I will try it in a real cluster to see if it is the test environment setup problem or not.

On a slightly different topic, do you think the MemberRollContoller should call the newly registered listener with the current state if the MMR is initialized, instead of [calling it with `nil`](https://github.com/maistra/istio/blob/cccdbb9ed1089992171c81fb490320d8387f7704/pkg/servicemesh/controller/controller.go#L101)?

pilot/pkg/config/kube/ior/route.go
yannuil: Make sure using `for` loop.

pilot/pkg/config/kube/ior/route.go
yannuil: Removed usage of `MemberRoll`.

pilot/pkg/config/kube/ior/route.go
yannuil: Still need to fetch current to determinate if the `Gateway` still exists. For reconciliation, `curr` is used.

pilot/pkg/config/kube/ior/route.go
yannuil: The `Route` is created at the namespace of where the `IngressGateway` service is. It is normally `istio-system`.

pilot/pkg/config/kube/ior/route.go
yannuil: Updated to use `Lister`.

pilot/pkg/config/kube/ior/route.go
yannuil: Updated to remove the `Route` from the current `Route` map even if the update fails. I agree that the deletion should only happen when the user tells us to.

pilot/pkg/config/kube/ior/route.go
yannuil: The reason I used this syntax is the `err` variable will be re-used. For example the code below, the `item, err := processList(list)` will not be happy about the use of  `err`.

```go
	list, err := findX()
	if err != nil {
		return err
	}

	items, err := processList(list)
	
	if err != nil {
		return err
	}

	return items
```

luksa: Well, yes, but the variable isn't being shadowed in this part of the code, but further down. Whether you declare the variable using `var` or `:=` here makes no difference in the code further down - the `err` variable is shadowed in both cases.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -50,38 +50,18 @@ const (
...
 )
 
-type syncRoutes struct {
-	metadata config.Meta
-	gateway  *networking.Gateway
-	routes   []*v1.Route
-}
-
 // route manages the integration between Istio Gateways and OpenShift Routes
 type route struct {
```
luksa: Can you rename this to `routeController`? The current name is misleading and confuses me every time I read through this file. Since this represent a controller and not a route, it should be named accordingly. This change will also remove the shadowing in https://github.com/maistra/istio/pull/673/files#diff-d92eadb815953403f29e9eb8198b9b337121ecf2e8571c8f5d3bf0127f3a4353R446.

pilot/pkg/config/kube/ior/route.go
luksa: Aha, this is where the `err` variable would be shadowed. OK, I agree you shouldn't make this change here. But you can definitely make the change in lines 414-419.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +341,149 @@ func hostHash(name string) string {
...
+	}
+
+	config := r.store.Get(collections.IstioNetworkingV1Alpha3Gateways.Resource().GroupVersionKind(), curr.Name, curr.Namespace)
+
+	var routes *v1.RouteList
+
+	routes, err = r.findRoutes(curr.Meta)
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to find routes matching gateway %s/%s", curr.Name, curr.Namespace)
```
luksa: Can we change this error to "error finding routes ..."

The "unable to find routes" error might be misunderstood as "no routes found".

pilot/pkg/config/kube/ior/route.go
luksa: Yes, but consider this scenario: I create the `Gateway` in `bookinfo`, while the `IngressGateway` service is in `istio-system`. The `Route` is therefore in `istio-system`.  But this code tries to find the `Route` in `bookinfo`, since the code uses the namespace of the `Gateway` not the `IngressGateway` service. 

Also be aware that the `IngressGateway` service might be in a different namespace (neither `bookinfo` nor `istio-system`).

luksa: Some more info on this: I create a `Gateway` called `bookinfo-gateway` in `bookinfo`. The ingress gateway is in `istio-system`. If I then do:
```
kubectl -n bookinfo annotate gateway bookinfo-gateway foo=bar
```
the istiod log shows the following error:
```
error   ior     failed to process gateway bookinfo-gateway/bookinfo event update: 1 error occurred:
        * error creating a route for the host * from gateway: bookinfo/bookinfo-gateway: routes.route.openshift.io "bookinfo-bookinfo-gateway-684888c0ebb17f37" already exists
```

As I explained above, IOR thinks that the `Route` doesn't exist (because it's looking for it in `bookinfo`) and thus tries to create it (in `istio-system`). This fails, because the Route *does* exist there.

yannuil: It will look for all namespaces now. It uses the labels to narrow the matching results. The labels will include the name of the Gateway resource and the Namespace the Gateway is in.

yannuil: @luksa when I changed [line 260](https://github.com/maistra/istio/pull/673/files?diff=split&w=0#diff-d92eadb815953403f29e9eb8198b9b337121ecf2e8571c8f5d3bf0127f3a4353R260) to be 
```go
return r.routerClient.Routes(metav1.NamespaceAll).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.String()})
```

using `NamespaceAll` to select all routes and expect the label selector to narrow the result set. The service mesh integration failed. I will test it more.

luksa:
/istio-push-images-2.3

pilot/pkg/config/kube/ior/route.go
luksa: Why do you need to search in all namespaces? The `buildRoute()` function creates the Route in the `serviceNamespace` (the namespace of the ingress gateway pod). Can't you look for the Route only in that namespace?

luksa:
Closing this PR. These changes are going into 2.4 (see #747).

PR #674, opened 2022-11-07
OSSM-487 Shorten generated route name

bmangoen:

This will fix the creation of generated route name by IOR relying on the long gateway namespace

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/ior/route.go
+++ b/pilot/pkg/config/kube/ior/route.go
@@ -36,6 +36,7 @@ import (
 	networking "istio.io/api/networking/v1alpha3"
 	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config"
+	istiolabels "istio.io/istio/pkg/config/labels"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/istio/pkg/kube"
 	"istio.io/istio/pkg/servicemesh/controller"
@@ -52,6 +53,7 @@ const (
 	ShouldManageRouteAnnotation = maistraPrefix + "manageRoute"
 
 	eventDuplicatedMessage = "event UPDATE arrived but resourceVersions are the same - ignoring"
+	hashLength             = 8
 )
 
 type syncRoutes struct {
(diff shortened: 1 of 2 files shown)


pilot/pkg/config/kube/ior/route.go
```diff
@@ -583,8 +585,13 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
 		gwSelector.String(), namespaces)
 }
 
-func getRouteName(namespace, name, actualHost string) string {
-	return fmt.Sprintf("%s-%s-%s", namespace, name, hostHash(actualHost))
+func getRouteName(namespace, name, actualHost, serviceNamespace string) string {
```
luksa, line 588: Why add `serviceNamespace` here?

pilot/pkg/config/kube/ior/route.go
```diff
@@ -583,8 +585,13 @@ func (r *route) findService(gateway *networking.Gateway) (string, string, error)
...
+	routeName := fmt.Sprintf("%s-%s-%s", namespace, name, hostHash(actualHost))
+	routeHost := fmt.Sprintf("%s-%s", routeName, serviceNamespace)
+	if len(routeHost) > istiolabels.DNS1123LabelMaxLength {
+		routeName = routeName[:istiolabels.DNS1123LabelMaxLength-(len(serviceNamespace)+1)]
```
luksa, line 592: This doesn't look right. It checks the length of `routeHost`, but then shortens `routeName`?

pilot/pkg/config/kube/ior/route.go
bmangoen: I need to know what is the ingress controller namespace and its length because when the hostname is set to "*" we let [OpenShift creates the route](https://github.com/maistra/istio/blob/maistra-2.3/pilot/pkg/config/kube/ior/route.go#L603-L607). 
In this case, the Route `spec.host` is set following the format `gateway_namespace-gateway_name-ingressgateway_namespace` (`Gateway.metadata.name` + ingress controller namespace) and it shouldn't exceed 63 characters.

PR #675, opened 2022-11-07
OSSM-1698 Restrict istio-cni-node DaemonSet

InnocentK:

istio-cni by default is installed on all nodes. This PR adds the ability to exclude istio-cni from being installed on specified nodes by adding the `maistra.io/exclude-cni` label to that node and setting it to `true`.

- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/manifests/charts/istio-cni/templates/daemonset.yaml
+++ b/manifests/charts/istio-cni/templates/daemonset.yaml
@@ -47,7 +47,7 @@ spec:
               - key: maistra.io/exclude-cni
                 operator: NotIn
                 values:
-                - true
+                - "true"
       tolerations:
         # Make sure istio-cni-node gets scheduled on all nodes.
         - effect: NoSchedule
--- a/manifests/charts/istio-cni/templates/daemonset.yaml
+++ b/manifests/charts/istio-cni/templates/daemonset.yaml
@@ -38,7 +38,7 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
-      # Allows for exluding instio-cni from being scheduled on specified nodes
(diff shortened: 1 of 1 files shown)


manifests/charts/istio-cni/templates/daemonset.yaml
```diff
@@ -38,6 +38,16 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
+      # Allows for exluding instio-cni from being scheduled on specified nodes
```
luksa:
```suggestion
      # Allows for excluding instio-cni from being scheduled on specified nodes
```

manifests/charts/istio-cni/templates/daemonset.yaml
```diff
@@ -38,6 +38,16 @@ spec:
...
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: maistra.io/exclude-cni
+                operator: NotIn
+                values:
+                - true
```
luksa: I think this manifest will fail to apply because of this boolean value. It should be a string.
```suggestion
                - "true"
```

luksa (commented):
Just a quick first review. I'll take another look tomorrow, but I think this change might need to go into `maistra/istio-operator` instead of here, since we completely override this file there.

InnocentK:
Closed in favor of https://github.com/maistra/istio-operator/pull/1057 (see @luksa's comment about istio-operator overriding this file)

PR #679, opened 2022-11-09
[no-cluster-rbac] OSSM-2255: Preliminary support for Gateway API

bartoszmajsak:
Several things had to be adjusted in order to make it working with `1.16` stream.

1. As new CRDs have been introduced (such as `ReferenceGrant`) both cluster-role and role manifest have been updated.
2. Removed `v1alpha2` gateway tests, as I assume that's gone now?
3. `Kind: Mesh` is not a valid `parentRef` anymore. It's required to refer to the service now (see `gateway_test.go`). Without this "mesh" test does not work anymore. One could wonder why it's called `"mesh"` at this point, but it's the same in `ingress_test.go` from which our tests are derived I assume.
4. In order to have access to the `test-gateway-cert-cross` secret from the Gateway a `ReferenceGrant` needs to be defined - I added it to our test. The reason why it works in `ingress_test.go` is because there, the secret is created in the same ns as the gateway.


pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -23,6 +23,7 @@ import (
 	"text/template"
 	"time"
 
+	maistrav1alpha2 "github.com/maistra/xns-informer/pkg/generated/gatewayapi/apis/v1beta1"
```
bartoszmajsak: TODO: shouldn't this be `maistrav1beta1`

bartoszmajsak:
~~@dgn @jwendell in the [test data](https://github.com/istio/istio/blob/release-1.16/tests/integration/pilot/testdata/gateway-api-crd.yaml#L6365-L6377) there's no `v1beta1` version of `ReferenceGrant`. Should this be fixed or we just leave it as is for now?~~

Fixed in `master@upstream`.

dgn (approved):
> Several things had to be adjusted in order to make it working with `1.16` stream.
> 
>     1. As new CRDs have been introduced (such as `ReferenceGrant`) both cluster-role and role manifest have been updated.

:+1: 
> 
>     2. Removed `v1alpha2` gateway tests, as I assume that's gone now?

Hmm no, v1alpha2 is still a thing, as a lot of the GW API CRDs have not been promoted to beta yet. I think it's fine this way though, we have the same coverage as upstream.

>     3. `Kind: Mesh` is not a valid `parentRef` anymore. It's required to refer to the service now (see `gateway_test.go`). Without this "mesh" test does not work anymore. One could wonder why it's called `"mesh"` at this point, but it's the same in `ingress_test.go` from which our tests are derived I assume.

:+1:

>     4. In order to have access to the `test-gateway-cert-cross` secret from the Gateway a `ReferenceGrant` needs to be defined - I added it to our test. The reason why it works in `ingress_test.go` is because there, the secret is created in the same ns as the gateway.

:+1:

PR #680, opened 2022-11-10
OSSM-2256: Add IOR

jwendell:


go.mod
```diff
@@ -55,7 +55,8 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/moby/buildkit v0.10.4
 	github.com/onsi/gomega v1.20.2
-	github.com/openshift/api v0.0.0-20200713203337-b2494ecb17dd
+	github.com/openshift/api v0.0.0-20200929171550-c99a4deebbe5
+	github.com/openshift/client-go v0.0.0-20200929181438-91d71ef2122c
```
bartoszmajsak, line 59: Is there any particular reason for 2y old version?

pilot/pkg/config/kube/ior/client.go
```diff
@@ -0,0 +1,59 @@
...
+	"strings"
+	"time"
+
+	"istio.io/istio/pkg/kube"
+)
+
+// KubeClient is an extension of `kube.Client` with auxiliary functions for IOR
+type KubeClient interface {
+	IsRouteSupported() bool
+	GetActualClient() kube.Client
```
bartoszmajsak, line 27: If we can choose the name here what do you think about `UnwrapClient` instead?

pilot/pkg/config/kube/ior/client.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"strings"
+	"time"
+
+	"istio.io/istio/pkg/kube"
+)
+
+// KubeClient is an extension of `kube.Client` with auxiliary functions for IOR
+type KubeClient interface {
```
bartoszmajsak, line 25: Then why not simply call it `IORClient`?

pilot/pkg/config/kube/ior/ior.go
```diff
@@ -0,0 +1,97 @@
...
+	networking "istio.io/api/networking/v1alpha3"
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pkg/config"
+	"istio.io/istio/pkg/config/schema/collections"
+	"istio.io/istio/pkg/servicemesh/controller"
+	"istio.io/pkg/log"
+)
+
+// IORLog is IOR-scoped log
+var IORLog = log.RegisterScope("ior", "IOR logging", 0)
```
bartoszmajsak, line 32: Does it need to be exported?

pilot/pkg/config/kube/ior/ior.go
```diff
@@ -0,0 +1,97 @@
...
+		// (2) It might take a few seconds to this channel to be closed. So, both pods might be leader for a few seconds.
+		<-stop
+		IORLog.Info("This pod is no longer a leader. IOR stopped responding")
+		aliveLock.Lock()
+		alive = false
+		aliveLock.Unlock()
+	}(stop)
+
+	IORLog.Debugf("Registering IOR into Istio's Gateway broadcast")
+	kind := collections.IstioNetworkingV1Alpha3Gateways.Resource().GroupVersionKind()
```
bartoszmajsak, line 64: Shouldn't we switch to `v1beta1` here?

tests/integration/servicemesh/smmr_test.go
```diff
@@ -166,16 +211,16 @@ spec:
...
-  - %s/common-gateway
+  - %s/%s
   http:
   - route:
     - destination:
         host: localhost
         port:
           number: 8080
-`, virtualServiceName, virtualServiceName, gatewayNs)
+`, virtualServiceName, virtualServiceName, gatewayNs, gatewayName)
```
bartoszmajsak, line 221: Instead of duplicating arguments here you could use positional and refer to virtual service name using `%[1]s`, so that would become sth like

```
metadata:
  name: %[1]s
spec:
  hosts:
  - "%[1]s.maistra.io"
  gateways:
  - %[2]s/%[3]s
```

bartoszmajsak (commented):
Few minor things we could improve here.

jwendell:
@bartoszmajsak IOR is dying. In 2.4 it will be deprecated (OSSM-2254). In 2.5 (or 3.0) it will be removed. Yann is refactoring some parts of it (OSSM-1689). Ideally I wouldn't like to introduce new changes. At least not in this PR. IMO any improvements (if any at all) should be done in follow ups. This is all about cherry-picking stuff from 2.3.

bartoszmajsak:
@jwendell as you wish, but I think if there are places we can improve a little bit and are not interfering with coming PR that would be nice.

bartoszmajsak (approved):

PR #685, opened 2022-11-16
Fix test arg from previous merge

jwendell:


dgn (approved):

dgn:
good catch :sweat_smile:

bartoszmajsak:
@dgn the funny thing is though that it does not matter. Does it mean we don't have any tests which exercise Node watch?

bartoszmajsak (approved):

PR #688, opened 2022-11-18
[cni] MAISTRA-2291 Support deployment of multiple plugin versions in Istio CNI

bartoszmajsak:
Includes:

  * MAISTRA-2135 Add unit tests for our CNI binary-prefix work (#325)

  * MAISTRA-2137 Make network namespace setup executable name configurable (#273)

    To support the deployment of multiple CNI plugin versions, the name of the
    executable that is invoked to set up the network namespace must be configurable.

  * OSSM-1430: CNI: Watch for modified files with a prefix (#510)

    Because our CNI pod contains more than one container, and they write to
    the same directory, and they watch for changes on those directories,
    changes made by one container trigger the watch on the other, which will
    responde by copying the files to the directory, which will in turn
    trigger the watcher of the other container in an endless loop.

    This leads to high CPU usage on the node.

    This PR changes the logic to only monitor for files that have the
    desired prefix. Thus, for example, the 2.2 container will only react to
    changes to files whose names  start with "v2-2". This avoid this race
    condition and achieve the same end result.**Please provide a description of this PR:**
 
 * MAISTRA-2051 use correct UID/GID in istio-iptables
 * OSSM-2082 CNI installer now creates the net.d directory if necessary #638

### DIFF
--- a/cni/pkg/install/binaries_test.go
+++ b/cni/pkg/install/binaries_test.go
@@ -15,7 +15,6 @@
 package install
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
@@ -64,17 +63,9 @@ func TestCopyBinaries(t *testing.T) {
 		},
 	}
 
-	for i, c := range cases {
+	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			srcDir, err := os.MkdirTemp("", fmt.Sprintf("test-case-%d-src-", i))
-			if err != nil {
-				t.Fatal(err)
-			}
-			defer func() {
-				if err := os.RemoveAll(srcDir); err != nil {
-					t.Fatal(err)
-				}
-			}()
+			srcDir := t.TempDir()
 			for filename, contents := range c.srcFiles {
 				err := os.WriteFile(filepath.Join(srcDir, filename), []byte(contents), os.ModePerm)
 				if err != nil {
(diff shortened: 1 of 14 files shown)


jwendell (approved):

cni/pkg/install/binaries_test.go
```diff
@@ -54,27 +56,49 @@ func TestCopyBinaries(t *testing.T) {
...
-			srcDir := t.TempDir()
+			srcDir, err := os.MkdirTemp("", fmt.Sprintf("test-case-%d-src-", i))
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer func() {
+				if err := os.RemoveAll(srcDir); err != nil {
+					t.Fatal(err)
+				}
+			}()
```
luksa: Not sure why we don't use `t.TempDir()` here.

luksa (approved):
LGTM, but we can probably simplify that temp dir stuff.

cni/pkg/install/binaries_test.go
bartoszmajsak: I see it came from https://github.com/maistra/istio/blame/maistra-2.3/cni/pkg/install/binaries_test.go#L86 (as a result of the previous rebase).

I will align it with how upstream does that (which is still a mix, but... in favor of simpler `t.TempDir()` call)

Thanks for spotting it!

luksa (approved):

PR #689, opened 2022-11-18
[proxy-uid] OSSM-2292: Run sidecars with dynamically-determined user IDs

bartoszmajsak:
(#230) (#541)

  * MAISTRA-1751: Set securityContext.fsGroup to comply with restricted SCC (#256)

  This is a rewrite of 46f0f4a for the Maistra 2.1 / Istio 1.8 rebase.

  * MAISTRA-2452 Apply dynamically-determined user ID to the istio-validation init container

  * OSSM-1847 ensure ProxyUID and ProxyGID are initialized when using gateway injection template #618 
    * do not default ProxyUID and ProxyGID for gateways
    * initialize ProxyGID appropriately for gateway injection

Co-authored-by: rcernich <rcernich@redhat.com>
Co-authored-by: Marko Lukša <marko.luksa@gmail.com>
Co-authored-by: Brad Ison <bison@coreos.com>
Co-authored-by: Brian Avery <bavery@redhat.com>
Co-authored-by: Marko Lukša <marko.luksa@gmail.com>

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -275,6 +275,7 @@ refresh-goldens:
 		./pkg/kube/inject/... \
 		./pilot/pkg/security/authz/builder/... \
 		./cni/pkg/plugin/...
+	@REFRESH_GOLDEN=true go test ${GOBUILDFLAGS} ./istioctl/cmd/...
 
 update-golden: refresh-goldens
 
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -274,8 +274,8 @@ refresh-goldens:
 		./pkg/bootstrap/... \
 		./pkg/kube/inject/... \
 		./pilot/pkg/security/authz/builder/... \
-		./cni/pkg/plugin/...
-	@REFRESH_GOLDEN=true go test ${GOBUILDFLAGS} ./istioctl/cmd/...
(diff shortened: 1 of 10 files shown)


Makefile.core.mk
```diff
@@ -275,6 +275,7 @@ refresh-goldens:
 		./pkg/kube/inject/... \
 		./pilot/pkg/security/authz/builder/... \
 		./cni/pkg/plugin/...
+	@REFRESH_GOLDEN=true go test ${GOBUILDFLAGS} ./istioctl/cmd/...
```
jwendell: nit: this could be appended to the previous command

jwendell (approved):

luksa (approved):

Makefile.core.mk
bartoszmajsak: You really love to stress our Prow, don't you? :D Do we do it here or upstream?

jwendell: I don't think upstream test this directory. Perhaps it's our addition.
Anyway, this is just a nit, don't bother to change this just for the sake of "fixing" it.

bartoszmajsak: Yeah, I noticed it's not in upstream. And it's already pushed 792b022078 :P

PR #396, opened 2021-07-09
MAISTRA-2495 Remove runAsUser field from bookinfo examples

luksa:

PR #691, opened 2022-11-22
[samples] Update bookinfo images for 2.4

bartoszmajsak:
update/rebuild maistra bookinfo images for 2.3 #505. remove runasuser from deployments #396.
add sidecar injection annotations #402.

use maistra org images

### DIFF
--- a/samples/bookinfo/platform/kube/bookinfo-db.yaml
+++ b/samples/bookinfo/platform/kube/bookinfo-db.yaml
@@ -41,13 +41,15 @@ spec:
       version: v1
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "true"
       labels:
         app: mongodb
         version: v1
     spec:
       containers:
       - name: mongodb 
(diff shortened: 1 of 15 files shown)


bartoszmajsak:
@jwendell should we update images to `2.4.0`? If so - where should I send a PR to?

jwendell:
> @jwendell should we update images to `2.4.0`? If so - where should I send a PR to?

I think we are fine with 2.3. We don't have 2.4 images yet.

Plus, can we have a single PR for all stuff under `samples`? For instance, can we move the changes in this PR https://github.com/maistra/istio/pull/695  into a commit here?

And this:
62ba2fef5f | ([misc] MAISTRA-2355: Change httpbin port to 8000 (#340), 2021-05-13)

bartoszmajsak:
@jwendell should we squash them all to one under this PR or keep them as is?

jwendell (approved):
I'd prefer squash everything together

PR #692, opened 2022-11-22
[istiod] OSSM-2294 Add --logKubernetesApiRequests option to istiod 

bartoszmajsak:
This option logs each individual request sent to the Kubernetes API. In addition to enabling this option, you need to set the log level to INFO.

Based on #625

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/controller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/controller.go
@@ -142,6 +142,9 @@ type Options struct {
 	// Maximum burst for throttle when communicating with the kubernetes API
 	KubernetesAPIBurst int
 
+	// Whether to log requests sent to the kubernetes API
+	LogKubernetesAPIRequests bool
+
 	// SyncTimeout, if set, causes HasSynced to be returned when timeout.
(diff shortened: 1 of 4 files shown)


jwendell (approved):
We should have upstreamed this
cc @luksa

pilot/pkg/serviceregistry/kube/controller/controller.go
```diff
@@ -109,6 +109,7 @@ type Options struct {
 	// MeshServiceController is a mesh-wide service Controller.
 	MeshServiceController *aggregate.Controller
 
+	ResyncPeriod time.Duration
```
luksa: This looks unrelated to the logKubernetesApiRequests thing.

bartoszmajsak: Yeah, I couldn't really find this being used anywhere. I guess we can remove it?

bartoszmajsak: For whatever reason, this comes from https://github.com/maistra/istio/pull/585. Removing for now.

luksa (approved):

PR #697, opened 2022-11-25
OSSM-2074 Fix issues in the analysis controller (#642)

bartoszmajsak:


jewertow (approved):

jwendell (approved):
I suggest you to create a subtask under OSSM-2149 for this. As it requires changes in operator as well. You won't forget about it until it's completely closed.

bartoszmajsak:
> As it requires changes in operator as well.

@jwendell what kind of changes do we need there? Sth more than [clusterrole update](https://github.com/maistra/istio-operator/pull/1024/files#diff-a5562f01aef8a1d54166f684abbbafaf1f5168dd3f28445c16d51c21c8cc0264R121-R125)? Can you comment on [OSSM-2311](https://issues.redhat.com/browse/OSSM-2311)?

jwendell:
> > As it requires changes in operator as well.
> 
> @jwendell what kind of changes do we need there? Sth more than [clusterrole update](https://github.com/maistra/istio-operator/pull/1024/files#diff-a5562f01aef8a1d54166f684abbbafaf1f5168dd3f28445c16d51c21c8cc0264R121-R125)? Can you comment on [OSSM-2311](https://issues.redhat.com/browse/OSSM-2311)?

The original Jira ticket had links to commits on both repos. Since we are not yet dealing with operator, I guess this is not complete until the operator part is also ported, right? Hence my suggestion of creating a new Jira so that it keeps open until then.

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -517,7 +517,10 @@ func handleCRDAdd(cl *Client, name string, stop <-chan struct{}) {
 	resourceGVK := s.Resource().GroupVersionKind()
 	gvr := s.Resource().GroupVersionResource()
 
-	if cl.client.GetMemberRoll() != nil && resourceGVK == gvk.GatewayClass {
```
dgn, line 520: I don't understand why we need this change?

bartoszmajsak: @dgn So the current code will skip it if gateway API is not enabled - if it's always enabled for our distribution then it's not needed. Otherwise that actually make sense? It seems though that 

```
else if cl.client.GetMemberRoll() != nil && resourceGVK == gvk.GatewayClass
```
this part is a bit confusing. Should I entirely remove it? Are we using `features.EnableGatewayAPI` now?

bartoszmajsak: It's a simple/naive merge from https://github.com/maistra/istio/pull/642/files

bartoszmajsak: @luksa as there's a bit of confusion about this `if` clause - I was hoping you could help us and share some context about why this is needed?

luksa: I remember getting an error related to `GatewayClass` when I enabled analysis. That's why this change was neede. However, I think I may have found other issues afterwards that I found out were caused by some other issue. It's possible that after fixing this second issue, this change here is no longer needed. 

But I don't remember what that other issue/fix was.

bartoszmajsak: Ok, so I am going to merge as-is. But we should create an issue to follow up on this. Would it make sense @dgn @luksa?

dgn: works for me

dgn (approved):

pilot/pkg/config/kube/crdclient/client.go
bartoszmajsak: https://issues.redhat.com/browse/OSSM-2337

PR #696, opened 2022-11-25
[tls] MAISTRA-2051: Add extended TLS options

bartoszmajsak:
MAISTRA-2051: TLS Config: Add TLS related feature flags

Partial pick of 1fe34221d5 for Maistra 2.1 / Istio 1.8 rebase.

This registers the feature flags for TLS related configuration.  The settings have been updated for the v3 Envoy APIs now in use by Istio.

Related Maistra issues: MAISTRA-1787, MAISTRA-1456, MAISTRA-1681, MAISTRA-1697, MAISTRA-1698, MAISTRA-1699

MAISTRA-2051: TLS Config: Set TLS options for istiod and webhooks

Partial pick of 1fe34221d5 for Maistra 2.1 / Istio 1.8 rebase.

This configures the TLS options for the main istiod server and for webhooks based on the previously added TLS feature flags.

Related Maistra issues: MAISTRA-1787, MAISTRA-1456, MAISTRA-1681, MAISTRA-1697, MAISTRA-1698, MAISTRA-1699

MAISTRA-2051: TLS Config: Set TLS options in gateway listeners

Partial pick of 1fe34221d5 for Maistra 2.1 / Istio 1.8 rebase.

This sets TLS options when building gateway listeners based on previously added TLS feature flags.

Related Maistra issues: MAISTRA-1787, MAISTRA-1456, MAISTRA-1681, MAISTRA-1697, MAISTRA-1698, MAISTRA-1699

MAISTRA-2051: TLS Config: Set TLS options in BuildInboundFilterChain

Partial pick of 1fe34221d5 for Maistra 2.1 / Istio 1.8 rebase.

This sets TLS options in the authn code in `BuildInboundFilterChain` based on the previously added TLS feature flags.  This removes the upstream `EnableTLSv2OnInboundPath` flag to avoid conflicting TLS settings.  That was introduced in fa0d260345 -- upstream PR: #27500

Related Maistra issues: MAISTRA-1787, MAISTRA-1456, MAISTRA-1681, MAISTRA-1697, MAISTRA-1698, MAISTRA-1699

MAISTRA-2051: TLS Config: Update charts with TLS options

Partial pick of 1fe34221d5 for Maistra 2.1 / Istio 1.8 rebase.

This updates the charts to set the new TLS feature flags.  Validation tests in the operator code are skipped because we add TLS settings in the charts, but don't add them to the Go types.

[...]

PR #418, opened 2021-08-16
MAISTRA-2472 allow DefaultKubernetesAudience to be configurable

rcernich:
Signed-off-by: rcernich <rcernich@redhat.com>

Please provide a description for what this PR is for.

Istio assumes the default audience used by the kubernetes token issuer is `kubernetes.default.svc`.  However, in OpenShift, the default audience is `https://kubernetes.default.svc`.  Furthermore, the default audience is configurable and not guaranteed to be `https://kuberenetes.default.svc` (even for bare kuberenetes, as it's a command line option).

This PR adds support for explicitly setting the appropriate default audience in use by the token issuer.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request. 

[ ] Does not have any changes that may affect Istio users.

PR #698, opened 2022-11-29
[misc] umbrella PR with different fixes

bartoszmajsak:
see commit list


bartoszmajsak:
When #696 gets merged I will rebase this branch to make linter happy.

jewertow (approved):

bin/update_maistra_deps.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux
+
+UPDATE_BRANCH=${UPDATE_BRANCH:-"maistra-2.3"}
```
jwendell: 2.4

bin/update_maistra_deps.sh
```diff
@@ -0,0 +1,46 @@
...
+go mod vendor
+
+# Copy CRD's
+dir=$(mktemp -d)
+git clone "https://github.com/maistra/api.git" "${dir}"
+pushd "${dir}"
+git checkout "${UPDATE_BRANCH}"
+popd
+
+# FIXME: https://issues.redhat.com/browse/MAISTRA-2353
```
jwendell: You can remove this fixme block entirely, this issue is no longer valid.

pkg/security/security.go
```diff
@@ -131,6 +131,12 @@ var (
 	TokenAudiences = strings.Split(env.Register("TOKEN_AUDIENCES", "istio-ca",
 		"A list of comma separated audiences to check in the JWT token before issuing a certificate. "+
 			"The token is accepted if it matches with one of the audiences").Get(), ",")
+
+	// DefaultKubernetesAudience specifies the default audience used by the kubernetes token issuer.
```
jwendell: https://github.com/maistra/istio/pull/418#issuecomment-899677060
:(

jwendell (commented):
nit: change the PR title as it currently only repeats the first commit message

pkg/security/security.go
jewertow: It seems that we no longer need this commit, because it's already implemented in the upstream Istio. Look at this PR: https://github.com/istio/istio/pull/34294.

bartoszmajsak: We have a bunch identified so maybe it's time to do it all at once?

jewertow: What do you mean? This feature is already implemented in upstream.

jwendell: Awesome! One less commit to carry over :)

bartoszmajsak: > What do you mean? This feature is already implemented in upstream.

@jewertow I misinterpreted `:(` comment from @jwendell and I think I missed your comment when submitting mine (as UI didn't refresh)

bartoszmajsak: That's gone now. Thanks guys!

jwendell (approved):

PR #701, opened 2022-12-02
OSSM-2324 Ensure ProxyUID and ProxyGID are never nil

luksa:
If they are nil, Helm renders them as "nil" (with quotes), which the YAML parser can't parse as a float.

### DIFF
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -893,7 +893,10 @@ func (wh *Webhook) inject(ar *kube.AdmissionReview, path string) *kube.Admission
 	var err error
 	var proxyUID *int64
 	var proxyGID *int64
+	tproxyInterceptionMode := pod.Annotations != nil && pod.Annotations["sidecar.istio.io/interceptionMode"] == string(model.InterceptionTproxy)
 	if hasOnlyIstioProxyContainer(pod) {
+		// we're injecting a gateway pod
+		// we set proxyUID/GID to whatever UID/GID is specified in securityContext.RunAsUser/RunAsGroup
 		if pod.Spec.SecurityContext != nil {
 			if pod.Spec.SecurityContext.RunAsUser != nil {
 				proxyUID = pointer.Int64Ptr(*pod.Spec.SecurityContext.RunAsUser)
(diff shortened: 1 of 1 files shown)


pkg/kube/inject/webhook.go
```diff
@@ -894,6 +894,8 @@ func (wh *Webhook) inject(ar *kube.AdmissionReview, path string) *kube.Admission
 	var proxyUID *int64
 	var proxyGID *int64
 	if hasOnlyIstioProxyContainer(pod) {
+		// we're injecting a gateway pod
+		// we set proxyUID/GID to whatever UID/GID is specified in securityContext.RunAsUser/RunAsGroup
```
jwendell, line 899: does the loop below (`for _, c := range pod.Spec.Containers {`) make any sense since this pod contains only the proxy container?

jewertow: I think it does. Nothing blocks you to deploy a sidecar container - I don't mean istio-proxy in this case. A user might deploy a sidecar container for collecting logs or any other custom component.

luksa: This if statement was changed in https://github.com/maistra/istio/pull/615. Before that change, the for loop was necessary, but after that PR, that's no longer the case.

luksa: Should we clean this up in this PR, a different PR, or just in 2.4?

jwendell: I'd do in this PR, separate commit

luksa: Done.

jwendell (approved):

jewertow:
Ok, I tested this PR and I got not default `runAsUser` values in sidecars and a gateway.

jewertow:
I would like to note that this problem occurs only in the default namespace.

jewertow:
I tested this change incorrectly so please ignore my previous comment.

dgn:
I guess this will fix the `nil` issue, but.. why does it even happen? Is the adding of the securityContext with the correct UID & GID done in another webhook? Is this a problem of which mutatingwebhook is triggered first?

luksa:
The reason this happens is complex. It depends on whether the user creating the Pod has the `anyuid` SCC. A cluster-admin, for example, does have this SCC, while regular cluster users don't. 

When a user _without_ the `anyuid` SCC created a Pod, the SCC(?) mutating webhook added the SecurityContext to the Pod in which it set the `runAsUser` ID. In this case, the injection webhook worked fine.

However, when a user _with_ the `anyuid` SCC created a Pod without any SecurityContext, the injection webhook didn't set the `ProxyUID` or `ProxyGID` values for the charts, resulting in `ProxyGID` being rendered as `"nil"`, which caused the parsing error. 

So, to avoid the error, we must ensure that the two values are set. And since the injection webhook doesn't have the `get namespaces` permission, it can't get the UID range of the namespace. So, the only thing it can do is set ProxyUID/GID to the standard value `1337`, which should be fine, since that's what upstream sets.

dgn:
Ah okay, so this only affects users that have cluster-admin (or similar) privileges. That's a start. Thanks Marko

luksa:
Test failures were legit. It turns out we shouldn't set `ProxyUID` to `1337` when the proxy uses the TPROXY interception mode.

dgn (approved):
I don't think the cherry-pick labels work, I believe you'll need to explicitly comment with `/cherry-pick <branch>`

PR #705, opened 2022-12-07
OSSM-2338: Do not configure sni-dnat router for federation deployment

jewertow:
ISTIO_META_ROUTER_MODE is ignored since Istio 1.11.


dgn (approved):
Thanks Jacek! Good catch.

PR #717, opened 2022-12-23
OSSM-2376: Don't start federation controllers until informers have synced

jewertow:
Federation discovery controller fetches config map with a remote CA root cert, so if the controller started before the informer has synced, it would fail to fetch the config map.

### DIFF
--- a/pkg/servicemesh/federation/discovery/controller.go
+++ b/pkg/servicemesh/federation/discovery/controller.go
@@ -24,7 +24,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	corev1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	v1 "maistra.io/api/federation/v1"
 
@@ -61,7 +60,6 @@ type Controller struct {
 	localNetwork      string
 	localClusterID    string
 	rm                common.ResourceManager
-	configMapInformer corev1.ConfigMapInformer
 	env               *model.Environment
 	federationManager server.FederationManager
 	statusManager     status.Manager
@@ -90,7 +88,6 @@ func NewController(opt Options) (*Controller, error) {
 		localClusterID:        opt.LocalClusterID,
 		localNetwork:          opt.LocalNetwork,
 		rm:                    opt.ResourceManager,
-		configMapInformer:     opt.ResourceManager.KubeClient().KubeInformer().Core().V1().ConfigMaps(),
 		env:                   opt.Env,
 		sc:                    opt.ServiceController,
 		stopChannels:          make(map[cluster.ID]chan struct{}),
@@ -124,7 +121,7 @@ func (c *Controller) Start(stopChan <-chan struct{}) {
 }
 
 func (c *Controller) HasSynced() bool {
-	return c.Controller.HasSynced() && c.configMapInformer.Informer().HasSynced()
+	return c.Controller.HasSynced() && c.rm.HasSynced()
 }
 
 func (c *Controller) RunInformer(_ <-chan struct{}) {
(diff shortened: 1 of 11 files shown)


pkg/servicemesh/federation/discovery/controller.go
```diff
@@ -116,8 +116,12 @@ func (c *Controller) Run(stopChan <-chan struct{}) {
 	}
 }
 
+func (c *Controller) Start(stopChan <-chan struct{}) {
+	c.Controller.StartController(stopChan, c.HasSynced)
```
jewertow: Federation discovery controller is not able to call its own `HasSynced()` when it calls `Start()` of its embedded `controller.Controller`. Now it calls `Start()` of the embedded controller, but it passes its own `HasSync()`.

pkg/servicemesh/federation/discovery/controller.go
```diff
@@ -116,8 +116,12 @@ func (c *Controller) Run(stopChan <-chan struct{}) {
...
+}
+
 func (c *Controller) HasSynced() bool {
-	return c.Controller.HasSynced()
+	return c.Controller.HasSynced() && c.rm.KubeClient().KubeInformer().Core().V1().ConfigMaps().Informer().HasSynced()
```
luksa: Can you store the ConfigMap informer as a field in `Controller` and use that field here and in all other places where the federation controller uses this informer? Like it's done here: https://github.com/Maistra/istio/blob/ac3eb3bf6e49b683bdddcb36b7697723c2a96ad7/pilot/pkg/config/kube/ingress/controller.go#L150

This way, it makes it clearer why we're checking the ConfigMap informer here. I think we should use the same pattern for all informers that a particular controller uses.

jewertow: Sure, I will do it.

pkg/servicemesh/federation/discovery/controller.go
luksa: But where is this `Start()` function being called?

pkg/servicemesh/federation/discovery/controller.go
jewertow: I created a field for ConfigMapInformer, but I didn't refactor other informers in this controller, because they are used just once, and if I stored them in fields, then ResourceManager would no longer make sense, but I don't want to extend the scope of this pull request.

pkg/servicemesh/federation/discovery/controller.go
jewertow: Here: https://github.com/maistra/istio/blob/403a6b63c620b8b7c2376f3a24b17022acbc0509/pkg/servicemesh/federation/federation.go#L175.

pkg/servicemesh/federation/discovery/controller.go
luksa: I actually thing that `ResourceManager` shouldn't expose `kube.Client`, but each informer individually. Only then can it properly implement `HasSynced()`. 

As you said on Slack, the federation controller also uses Endpoints and ServiceAccounts, but it doesn't ensure that those informers are synced. By exposing each informer individually, it makes it less likely that we'll forget to include an informer in the `HasSynced` logic. But I agree this is a broader change, so it can be done in a separate PR.

luksa: The more I think about this, the clearer it becomes that we need to refactor the ResourceManager, as I said above. This is an important source of current and future bugs.

And since the bug you're fixing here is one of those bugs, I think it makes sense to refactor the code in this PR (and thereby fix this entire group of HasSynced-related bugs).

jewertow: Ok, so I will expose only those informers that are used instead of exposing kube client. I will also extend HasSynced to check all informers.

pkg/servicemesh/federation/exports/controller.go
```diff
@@ -70,6 +70,14 @@ func (c *Controller) RunInformer(stopChan <-chan struct{}) {
...
 }
 
+func (c *Controller) Start(stopChan <-chan struct{}) {
+	c.Controller.StartController(stopChan, c.HasSynced)
+}
+
+func (c *Controller) HasSynced() bool {
+	return c.Controller.HasSynced() && c.rm.HasSynced()
+}
+
```
luksa: Can we get rid of the need to override these functions in every controller? Maybe by passing the `HasSynced()` function in the kube controller? Or, if it makes sense semantically, simply move the `resourceManager` to kube controller?

I'm thinking the current design is both too complex (we call StartController to which we pass our own implementation of HasSynced) and fragile (someone may implement a new controller and forget to override these method).

jewertow: Very good point. I removed these functions from discovery, imports and exports controller. Now only `ResourceManager.HasSynced()` is used in the base controller.

dgn (approved):
lgtm as a fix. I wonder if we should merge (or at least co-locate) the ResourceManager and the Controller code-- they are only ever used together anyway

jewertow:
> I wonder if we should merge (or at least co-locate) the ResourceManager and the Controller code

I considered this approach, but I prefer the current composition. Having `ResourceManger` that keeps all informers simplifies creation of controllers and relieves them of the need to have a function `HasSynced`.

dgn:
Yes- what I meant was that this Controller struct is created by us (even though from the package name it might not seem that way) and is only used for federation, so we could move it to the federation/common pkg

dgn:
To be clear I'm talking about the boiler plate struct in `pkg/kube/controller/controller.go`

dgn (approved):

PR #735, opened 2023-01-24
[wip] OSSM-3238 Revert protobuf-related parts of "Prevent deadlock between multiNamespaceInformer.lock and namespaceSet.lock (#729)

luksa:


luksa:
Closing in favor of #737

PR #747, opened 2023-02-16
OSSM-1689 Simplify IOR

yannuil:


pkg/kube/client.go
```diff
@@ -32,7 +32,10 @@ import (
 	gatewayapiinformer "github.com/maistra/xns-informer/pkg/generated/gatewayapi"
 	istioinformer "github.com/maistra/xns-informer/pkg/generated/istio"
 	kubeinformer "github.com/maistra/xns-informer/pkg/generated/kube"
+	osrouteinformer "github.com/maistra/xns-informer/pkg/generated/openshift/route"
 	xnsinformers "github.com/maistra/xns-informer/pkg/informers"
+	osrouteclient "github.com/openshift/client-go/route/clientset/versioned"
+	osroutefake "github.com/openshift/client-go/route/clientset/versioned/fake"
```
luksa: Can you rename these to just `route`, `routeinformer`, and `routefake`?

pkg/kube/client.go
```diff
@@ -125,6 +128,8 @@ type Client interface {
 	// Istio returns the Istio kube client.
 	Istio() istioclient.Interface
 
+	OsRoute() osrouteclient.Interface
```
luksa: This should also be renamed to just `Route()`.

luksa:
@yannuil If I update the Gateway, an error occurs when updating the Route:
```
2023-02-22T12:05:23.093027Z	error	ior	failed to process gateway bookinfo-gateway/bookinfo event update: 1 error occurred:
	* error updating a route for the host * from gateway: bookinfo/bookinfo-gateway: routes.route.openshift.io "bookinfo-bookinfo-gateway-684888c0ebb17f37" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

pilot/pkg/config/kube/ior/route.go
```diff
@@ -627,3 +331,143 @@ func hostHash(name string) string {
...
+	for _, server := range gateway.Servers {
+		for _, host := range server.Hosts {
+			var err error
+
+			name := getRouteName(config.Namespace, config.Name, host)
+
+			_, found := routeMap[name]
+
+			if found {
+				_, err = r.updateRoute(config.Meta, host, server.Tls, serviceNamespace, serviceName)
```
luksa: You need to pass the route from `routeMap` to `updateRoute` so that you can properly update it. To update an object, a controller must preserve the metadata of the existing object (this is especially true for the `resourceVersion`, but most other fields should be preserved as well.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -531,60 +183,112 @@ func (r *route) createRoute(metadata config.Meta, gateway *networking.Gateway, o
...
 		nr.Spec.Host,
 		metadata.Namespace, metadata.Name)
 
 	return nr, nil
 }
 
+func (r *routeController) findRoutes(metadata config.Meta) ([]*v1.Route, error) {
+	defaultLabelSet := getDefaultRouteLabelMap(metadata.Name, metadata.Namespace)
+
+	labels := labels.SelectorFromSet(defaultLabelSet)
```
luksa: please rename this variable to `selector` or simply inline it.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -531,60 +183,112 @@ func (r *route) createRoute(metadata config.Meta, gateway *networking.Gateway, o
...
+}
+
 // findService tries to find a service that matches with the given gateway selector
 // Returns the namespace and service name that is a match, or an error
-func (r *route) findService(gateway *networking.Gateway) (string, string, error) {
-	r.namespaceLock.Lock()
-	namespaces := r.namespaces
-	r.namespaceLock.Unlock()
-
+func (r *routeController) findService(gateway *networking.Gateway) (string, string, error) {
```
luksa: please change this function to return `NamespacedName` instead of `(string, string)`.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -531,60 +183,117 @@ func (r *route) createRoute(metadata config.Meta, gateway *networking.Gateway, o
...
+func (r *routeController) updateRoute(
+	metadata config.Meta,
+	originalHost string,
+	tls *networking.ServerTLSSettings,
+	serviceNamespace string, serviceName string,
+	route *v1.Route,
+) (*v1.Route, error) {
+	IORLog.Debugf("updating route for hostname %s", originalHost)
+
+	buildRoute(metadata, originalHost, tls, serviceNamespace, serviceName).DeepCopyInto(route.DeepCopy())
```
luksa: Does this do anything? I don't think it does.

It builds a new route and copies it into a copy of `route`. But that copy isn't stored anywhere. In the update request that follows, the original route is used, making the update a no-op.

tests/integration/servicemesh/smmr_test.go
```diff
@@ -110,33 +151,55 @@ func verifyThatIngressHasVirtualHostForMember(ctx framework.TestContext, expecte
...
-		if len(routes.Items) == 0 {
-			return fmt.Errorf("no Routes found")
+
+		if route == nil {
+			return fmt.Errorf("no Route found")
 		}
+
 		return nil
-	}, retry.Timeout(10*time.Second))
+	}, retry.Timeout(300*time.Second))
```
luksa: I don't think such a high timeout is necessary, since I've seen the test fail even with these 300s. 

I think these failed tests show that some kind of race condition is still present (either in the production code or the tests themselves).

tests/integration/servicemesh/smmr_test.go
```diff
@@ -110,33 +151,55 @@ func verifyThatIngressHasVirtualHostForMember(ctx framework.TestContext, expecte
...
+		if err != nil {
+			ctx.Fatalf("failed to get Routes: %v", err)
 		}
+
+		if route != nil {
+			ctx.Fatal("found unexpected Route")
+		}
+
+		return nil
+	}, retry.Timeout(300*time.Second))
```
luksa: Same here.

luksa:
Can you fix the `TestSMMR` test. I know you didn't write this, but the test enables IOR and never disables it. This also prevents us from running the test multiple times (with `go test -count 100`), which would be really helpful in finding the race condition.

tests/integration/servicemesh/smmr_test.go
yannuil: > I don't think such a high timeout is necessary, since I've seen the test fail even with these 300s.

Yes that is what I have found out with experimenting the increase of this timeout. The tests will still fail even with a long timeout. I will change them all back.

yannuil:
I will add the disabling IOR as well as cleaning up namespace and virtual services after the test.

yannuil:
@luksa I just found out that the [`targetPort`](https://github.com/maistra/istio/pull/747/files/40d101a3549b1273fd9e5ccd40fd6e7fbbc5cba2#diff-d92eadb815953403f29e9eb8198b9b337121ecf2e8571c8f5d3bf0127f3a4353L476-L481) can only be `http2` or `https`. How should we handle `Gateway` with hosts that are using other ports?

tests/integration/servicemesh/smmr_test.go
```diff
@@ -88,15 +99,24 @@ func TestSMMR(t *testing.T) {
 				verifyThatRouteExistsOrFail(t, namespaceGateway, gatewayName, "a.maistra.io")
 				verifyThatRouteExistsOrFail(t, namespaceGateway, gatewayName, "b.maistra.io")
 				verifyThatRouteIsMissingOrFail(t, namespaceGateway, gatewayName, "c.maistra.io")
+
+				if err := maistra.DisableIOR(t); err != nil {
+					t.Fatalf("failed to disable IOR: %s", err)
+				}
```
luksa: Doesn't this need to be in a `defer` block so that it runs even if the `verifyThatRouteExistsOrFail()` function causes the test to terminate?

pkg/test/framework/components/istio/ingress.go
```diff
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	getAddressTimeout = retry.Timeout(3 * time.Minute)
+	getAddressTimeout = retry.Timeout(6 * time.Minute)
```
jewertow: @luksa you can revert this change. Tests will faill until we merge this: https://github.com/maistra/istio/pull/761/commits/aede273d44a22db2dc630bb3770dbacaf7df80c9.

jewertow:
@yannuil could you rebase your branch? I merged commits which fixes integration tests.

pkg/test/framework/components/istio/ingress.go
luksa: 3 minutes seems a bit low. I think I remember LB provisioning taking up to 5 minutes.

luksa:
> @yannuil could you rebase your branch? I merged commits which fixes integration tests.

No need to rebase, as the rebase is performed automatically before the tests run.

pkg/test/framework/components/istio/ingress.go
jewertow: LB was failing, because of missing image, so this change is no longer needed.

luksa: @yannuil can you remove this commit?

luksa: Nevermind. I removed it myself.

pilot/pkg/config/kube/ior/route.go
```diff
@@ -50,267 +50,39 @@ const (
...
-		return nil, err
-	}
-
-	return client, nil
+	routeClient routeclient.Interface
+	routeLister routeListerV1.RouteLister
 }
 
 // newRoute returns a new instance of Route object
 func newRoute(
```
dgn: nit: maybe this should now be called `newRouteController?` I would also rename the file to `controller.go` I think

dgn (commented):
lgtm, I see the existing tests are passing, that's great. are we confident this is enough? one test (TestPerf) seems to have been removed- do we need a replacement?

luksa (approved):

pilot/pkg/config/kube/ior/route.go
luksa: renamed both.

PR #774, opened 2023-03-30
Fix integration tests

jewertow:
This PR includes:
1. Manual cherry-pick of fixed MetalLB configuration.
2. Fixes in maistra gateway API tests.


common/.commonfiles.sha
```diff
@@ -1 +1 @@
-45c6eb44b33b5df885719d4d5239b2038214852d
+9da77cfa03da2f1abdb2c325b63b6e68952c785f
```
jewertow, line 1: @jwendell this SHA comes from release 1.16. Is it a problem? We can also cherry-pick the same fix for 1.15, but not for 1.14.

jwendell: AFAICT this should not be an issue. This SHA is merely for informational purposes. It's not used anywhere, differently than the proxy sha in istio.deps. The worse it can happen is a merge conflict when pulling in upstream changes into 2.3 in the future. But this is very unlikely to happen because upstream 1.14 is dead, right?

jewertow: > because upstream 1.14 is dead, right?

Yes, it is.

dgn (approved):
LGTM. I'm guessing we'll want to cherry-pick this to 2.4 as well?

luksa:
I believe this is already in 2.4 (at least the metalLB part)

luksa:
@jewertow does this change also need to go into maistra-2.2? The tests also fail there, but I haven't checked why. I'd assume it's due to the same underlying cause.

jewertow:
Yes @luksa, the first commit should be cherry-picked to maistra-2.2 as well. I will do it.

PR #775, opened 2023-03-31
OSSM-3599 Federation egress-gateway gets wrong network gateway endpoints

bmangoen:
It would remove the adding of extra endpoints (IPs from the local ingress) in the egress gateway

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pkg/servicemesh/federation/model/model.go
+++ b/pkg/servicemesh/federation/model/model.go
@@ -23,9 +23,8 @@ type ServiceKey struct {
 }
 
 type ServiceListMessage struct {
-	Checksum                uint64             `json:"checksum" hash:"ignore"`
-	NetworkGatewayEndpoints []*ServiceEndpoint `json:"networkGatewayEndpoints,omitempty" hash:"set"`
-	Services                []*ServiceMessage  `json:"services,omitempty" hash:"set"`
+	Checksum uint64            `json:"checksum" hash:"ignore"`
+	Services []*ServiceMessage `json:"services,omitempty" hash:"set"`
 }
 
 type ServiceMessage struct {
--- a/pkg/servicemesh/federation/server/server.go
+++ b/pkg/servicemesh/federation/server/server.go
@@ -27,7 +27,6 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
-	"github.com/mitchellh/hashstructure/v2"
 	"k8s.io/apimachinery/pkg/util/errors"
 	v1 "maistra.io/api/federation/v1"
 
@@ -78,8 +77,6 @@ type Server struct {
 	// us to know what the workload identifiers were so we could manage the
 	// routing config for each mesh.  This may or may not be possible.
 	network string
-
-	currentGatewayEndpoints []*federationmodel.ServiceEndpoint
 }
 
 var _ FederationManager = (*Server)(nil)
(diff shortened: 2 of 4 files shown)


luksa:
I can reproduce this issue on 2.3.2, but not on 2.2.x, even though the same code is present in both releases. Not sure why there is a difference. However, applying this change does fix the issue. @bmangoen are you sure we can remove this code? Are you sure it doesn't break something else?

bmangoen:
> Are you sure we can remove this code? Are you sure it doesn't break something else?

@luksa After discussing with @dgn , removing all the related GatewayEndpoints code on the server side doesn't break anything (and fix this issue). The `gatewayForNetworkAddress` function [here](https://github.com/maistra/istio/blob/dd80818da4e87d5c86394fe92acbaf9eab8a9f6a/pilot/pkg/serviceregistry/federation/controller.go#L481) already fills the `gatewayStore` with the right endpoints on client side.

pkg/servicemesh/federation/server/server_test.go
```diff
@@ -579,7 +579,6 @@ func TestServiceList(t *testing.T) {
 			stopCh := make(chan struct{})
 			go s.Run(stopCh)
 			defer close(stopCh)
-			s.resyncNetworkGateways()
```
dgn, line 582: this test has two test cases that will need to be touched, we'll at the very least need to remove the NetworkGatewayEndpoints from their expected message: "exported service + gateway" and "exported service + gateway, updated"

pkg/servicemesh/federation/server/server.go
```diff
@@ -234,68 +230,23 @@ func (s *Server) Run(stopCh <-chan struct{}) {
...
-	if oldGatewayChecksum != newGatewayChecksum {
-		s.currentGatewayEndpoints = gatewayEndpoints
-		return true, nil
-	}
-	return false, nil
-}
-
 func (s *Server) UpdateService(svc *model.Service, event model.Event) {
 	// this might be a NetworkGateway
 	if svc != nil {
```
dgn: this whole if-block needs to be removed. this was for sending an empty update event in case a gateway was updated - we're not sending gateways anymore, so this can go. That will then need a bunch of changes in the tests again, especially TestWatch

dgn (commented):
LGTM once we resolve the unit test failures. Those are expected though as we're changing behavior a bit.

@luksa we came to the conclusion that this was the cleanest way to fix the problem. The code is not exactly the same across 2.2 and 2.3 streams, the rebase required a few changes because there were refactorings regarding NetworkGateways in istiod. In the past, we were able to filter NetworkGateways appropriately and only send local endpoints over the wire, but this bug surfaced because we're now sending both local and remote NetworkGatewayEndpoints- so in the end, the other side will add its own ingress gateway IPs as endpoints for the remote mesh, which is bound to fail. The whole idea of sending NetworkGatewayEndpoints using the protocol was added by me before we later added proper DNS resolution to the ServiceMeshPeer resource; in short, we don't need it anymore. We know the gateway endpoints by resolving DNS on the ServiceMeshPeer's remoteAddress, so we can remove this from the federation protocol, hopefully removing a bit of complexity here.

In fact, the NetworkGateway usage is wrong anyway... Another historical detail. I added it while working on the prototype that used SNI-DNAT, which is what upstream uses to route requests. That would only have allowed for renaming services on export, but Kev wanted renaming both on export and import, so we had to add the VirtualService/Gateway indirection which rendered the whole NetworkGateway effort useless. In my initial prototype, I relied entirely on Istio's NetworkGateway feature to resolve remote service endpoints, but now we're adding the endpoints manually, so we don't really need it anymore.

To give even more context, the NetworkGateways in istio belong to multi-cluster support and are used for east-west gateways. They allow sending traffic to a Service in another network without having its actual endpoints: istiod will simply add the endpoints of all NetworkGateways for the respective network as endpoints, and just assumes that the gateway will take care of routing to the actual service. That works because requests go straight to the other side's ingress-gateway and Service FQDNs are unaltered. In our case though, we would be sending requests to a gateway asking to connect to a Service with a manipulated FQDN (the 'import name'), and the remote end wouldn't know how to route the request.

pkg/servicemesh/federation/model/model.go
```diff
@@ -23,9 +23,8 @@ type ServiceKey struct {
 }
 
 type ServiceListMessage struct {
-	Checksum                uint64             `json:"checksum" hash:"ignore"`
-	NetworkGatewayEndpoints []*ServiceEndpoint `json:"networkGatewayEndpoints,omitempty" hash:"set"`
```
luksa, line 27: Will removing this field cause an incompatibility between different releases? (Is this message passed from one cluster to another? If they run different versions of istiod, will this change cause issues?)

dgn: nope it won't cause issues. as I mentioned, we don't need the endpoints that come over the wire to establish connections, and this is all transmitted in JSON, so if a field is not there it'll just be `nil` on the other side. The only place we're using it is in the loop where we `range` over it, and that will work

luksa: ok, thanks

luksa (approved):

PR #778, opened 2023-04-05
OSSM-3703: Fix AUTO_RELOAD_PLUGIN_CERTS in multi-tenant mode

jewertow:
Namespace controller crashes on secret update, because uninitialized namespace lister is called. In cluster-scoped deployment this failure should not happen, because initializing of the namespace lister is skipped only when SMMR controller is enabled.

```
2023-04-05T17:13:53.643387Z	info	ads	XDS: Incremental Pushing:2023-04-05T17:11:50Z/2 ConnectedEndpoints:3 Version:2023-04-05T17:11:50Z/2
2023-04-05T17:13:53.820263Z	info	ads	Push debounce stable[4] 2 for config Secret/istio-system/cacerts and 1 more configs: 100.411771ms since last change, 159.288056ms since last push, full=false
2023-04-05T17:13:53.820393Z	info	ads	XDS: Incremental Pushing:2023-04-05T17:11:50Z/2 ConnectedEndpoints:3 Version:2023-04-05T17:11:50Z/2
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x2984f6e]

goroutine 1070 [running]:
istio.io/istio/pilot/pkg/serviceregistry/kube/controller.(*NamespaceController).startCaBundleWatcher(0xc000735080, 0xc000e8d500)
	istio.io/istio/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go:147 +0x10e
created by istio.io/istio/pilot/pkg/serviceregistry/kube/controller.(*NamespaceController).Run
	istio.io/istio/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go:136 +0x20a
```

Logs of fixed istiod:
```
2023-04-05T19:34:14.678724Z info Update Istiod cacerts
2023-04-05T19:34:14.678825Z info Using kubernetes.io/tls secret type for signing ca files
2023-04-05T19:34:14.813470Z info Istiod has detected the newly added intermediate CA and updated its key and certs accordingly
2023-04-05T19:34:14.813707Z info x509 cert - Issuer: "CN=istiod-basic.istio-system.svc", Subject: "", SN: 2e3c5a56bc5913b8b634e144c79e61c3, NotBefore: "2023-04-05T19:32:14Z", NotAfter: "2033-04-02T19:34:14Z"
2023-04-05T19:34:14.813770Z info x509 cert - Issuer: "CN=root-ca.my-company.net,O=my-company.net", Subject: "CN=istiod-basic.istio-system.svc", SN: adc72370c45bb0cb71c0f8d5da97244c, NotBefore: "2023-04-05T19:33:37Z", NotAfter: "2023-05-07T11:33:37Z"
2023-04-05T19:34:14.813803Z info x509 cert - Issuer: "CN=root-ca.my-company.net,O=my-company.net", Subject: "CN=root-ca.my-company.net,O=my-company.net", SN: 352d9e6c05c500d887b433e367f97273, NotBefore: "2023-04-05T14:35:01Z", NotAfter: "2025-09-21T14:35:01Z"
2023-04-05T19:34:14.813810Z info Istiod certificates are reloaded
```

### DIFF
--- a/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
@@ -144,9 +144,16 @@ func (nc *NamespaceController) startCaBundleWatcher(stop <-chan struct{}) {
 	for {
 		select {
 		case <-watchCh:
-			namespaceList, _ := nc.namespaceLister.List(labels.Everything())
-			for _, ns := range namespaceList {
-				nc.namespaceChange(ns)
+			var namespaceList []*v1.Namespace
+			if nc.usesMemberRollController {
+				for _, ns := range nc.namespaces.List() {
+					nc.syncNamespaceByName(ns)
+				}
+			} else {
+				namespaceList, _ = nc.namespaceLister.List(labels.Everything())
+				for _, ns := range namespaceList {
+					nc.namespaceChange(ns)
+				}
 			}
 		case <-stop:
 			return
@@ -188,12 +195,13 @@ func (nc *NamespaceController) syncNamespace(ns *v1.Namespace) {
 	if nc.DiscoveryNamespacesFilter != nil && !nc.DiscoveryNamespacesFilter.FilterNamespace(ns.ObjectMeta) {
 		return
 	}
+	nc.queue.Add(types.NamespacedName{Name: ns.Name})
+}
 
-	// If a MemberRoll controller is in use, and the set of
-	// namespaces still includes the one for this ConfigMap,
-	// then recreate the ConfigMap, otherwise do nothing.
-	if nc.usesMemberRollController && !nc.namespaces.Contains(ns.Name) {
+func (nc *NamespaceController) syncNamespaceByName(ns string) {
+	// skip special kubernetes system namespaces
+	if inject.IgnoredNamespaces.Contains(ns) {
 		return
 	}
-	nc.queue.Add(types.NamespacedName{Name: ns.Name})
+	nc.queue.Add(types.NamespacedName{Name: ns})
 }
--- a/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
+++ b/pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
@@ -144,13 +144,12 @@ func (nc *NamespaceController) startCaBundleWatcher(stop <-chan struct{}) {
 	for {
 		select {
 		case <-watchCh:
-			var namespaceList []*v1.Namespace
 			if nc.usesMemberRollController {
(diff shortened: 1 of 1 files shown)


luksa (approved):

luksa:
Didn't test, but looks reasonable.

pilot/pkg/serviceregistry/kube/controller/namespacecontroller.go
```diff
@@ -188,12 +194,13 @@ func (nc *NamespaceController) syncNamespace(ns *v1.Namespace) {
...
 		return
 	}
+	nc.queue.Add(types.NamespacedName{Name: ns.Name})
+}
 
-	// If a MemberRoll controller is in use, and the set of
-	// namespaces still includes the one for this ConfigMap,
-	// then recreate the ConfigMap, otherwise do nothing.
-	if nc.usesMemberRollController && !nc.namespaces.Contains(ns.Name) {
+func (nc *NamespaceController) syncNamespaceByName(ns string) {
```
dgn, line 200: by creating a new `func` here we're skipping the DiscoveryNamespacesFilter that is present in `syncNamespace()`. That would be a behavior change, so I'd suggest we either merge the two functions again or add the filtering logic here as well.

Actually now that I think about it, this filtering is new in 2.4 so we can decide how to handle it... Either we apply the filters on top of the SMMR-preselection or we don't. I'd say we apply them, to stay closer to upstream behavior. So SMMR is the maximum set of namespaces watched, and it can be filtered by the DiscoverySelectors. otherwise DiscoverySelectors would be broken in multi-tenant OSSM

jewertow: I did it intentionally, because we cannot implement analogous function based on NamespeceSet from xns-informer, which is a list of namespace names, not namespace objects, so we cannot perform filtering of label selectors.
I ignored this filtering, because our users select namespaces with SMMR, not discovery selectors. Of course they could use SMMR + discovery selectors, but it can only be configured with `meshConfig.discoverySelectors`, which is tech-preview or an unsupported API. So should we care about this filtering?

jewertow: I think that I didn't break anything, because namespace lister is not initialized when SMMR controller is enabled, so that function couldn't work in multi-tenant mode anyway. I just found an execution path that fails.

dgn: oh you're right, we dont have access to the labels anyway. ignore my comment

dgn (approved):

PR #795, opened 2023-04-27
Remove vendor execution from script

leoieggli:


maistra/scripts/run-test-infra-script.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
jwendell: remove this script from this pr. It's already in the tree

jwendell (changes requested):

jwendell (approved):

PR #803, opened 2023-05-22
OSSM-3980: Fix gateway deployment controller in multi-tenant mode

jewertow:
Gateway deployment controller was logging errors in multi-tenant mode, because it was not using xns-informers for watching deployments.
```
2023-05-17T22:52:32.649678Z	error	klog	github.com/maistra/xns-informer/pkg/generated/kube/factory.go:135: Failed to watch *v1.Deployment: failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:istio-system:istiod-basic" cannot list resource "deployments" in API group "apps" at the cluster scope
```
We didn't find this problem in integration tests, because existing cluster role bindings were not removed before applying role bindings for multi-tenant deployment.


dgn (approved):

jewertow:
I verified failed test case locally and it succeeds.

PR #804, opened 2023-05-24
Sync with Istio 1.16.5

jewertow:


luksa (approved):
LGTM

Do we need to disable the automatic sync so we don't get additional commits?

jwendell:
> Do we need to disable the automatic sync so we don't get additional commits?

Up to you. You can also have the automation around. The generated PR is not merged automatically anyway.

jwendell:
FYI: The push-image job failure seems to be legit. It was not showing up before because we were using a different image to run the jobs. Now that we are using the proper image, this error showed up.

We don't need t fix it for this PR though. We can force this test to pass by commenting `/override ci/prow/maistra-istio-push-image-2-4`

PR #814, opened 2023-06-28
[2.5] OSSM-4245: Rebase multi-tenancy

jewertow:
This pull request rebases all multi-tenancy and no-cluster-rbac changes. Additionally, I implemented multi-tenancy integration tests and improved* existing tests.

*these tests may seem more complicated, but now they are more similar to istiod installed by our operator.

I will squash some commits when I will have approvals and after applying potential suggestions from reviewers.

@jwendell @yxun do you know, which label do I need to not squash commits?


go.mod
```diff
@@ -8,6 +8,8 @@ exclude k8s.io/kubernetes v1.13.0
 // Client-go does not handle different versions of mergo due to some breaking changes - use the matching version
 replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 
+replace maistra.io/api v0.0.0-20221103173341-6ef6ed929778 => github.com/maistra/api v0.0.0-20230704084350-dfc96815fb16
```
jewertow: @jwendell @yxun do you know how to publish maistra.io/api to remove this hack?

pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -175,7 +175,11 @@ func (s *Server) initK8SConfigStore(args *PilotArgs) error {
 					AddRunFunction(func(leaderStop <-chan struct{}) {
 						// We can only run this if the Gateway CRD is created
 						if configController.WaitForCRD(gvk.KubernetesGateway, leaderStop) {
-							tagWatcher := revisions.NewTagWatcher(s.kubeClient, args.Revision)
+							var tagWatcher revisions.TagWatcher
+							// TagWatcher requires permission for MutatingWebhook, so it can't be used in multi-tenant mode
+							if s.kubeClient.GetMemberRoll() == nil {
+								tagWatcher = revisions.NewTagWatcher(s.kubeClient, args.Revision)
+							}
```
jewertow, line 182: JFYI: This change is new comparing to previous releases.

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -186,23 +183,32 @@ func NewDeploymentController(client kube.Client, clusterID cluster.ID, env *mode
...
+		dc.gatewayClasses.AddEventHandler(controllers.ObjectHandler(func(o controllers.Object) {
+			for _, g := range dc.gateways.List(metav1.NamespaceAll, klabels.Everything()) {
+				if string(g.Spec.GatewayClassName) == o.GetName() {
+					dc.queue.AddObject(g)
+				}
+			}
+		}))
+	}
+
+	if tw != nil {
```
jewertow: Actually, this condition might be merged with the previous one (`GetMemberRoll() == nil`), because `tw` is also created conditionally, but outside the gateway controller, so I didn't know if this is elegant. Any opinion?

pilot/pkg/credentials/kube/secrets.go
```diff
@@ -80,15 +79,12 @@ func NewCredentialsController(kc kube.Client) *CredentialsController {
...
-		fields.OneTermNotEqualSelector("type", "helm.sh/release.v1"),
-		fields.OneTermNotEqualSelector("type", string(v1.SecretTypeServiceAccountToken))).String()
-	secrets := kclient.NewFiltered[*v1.Secret](kc, kclient.Filter{
-		FieldSelector: fieldSelector,
-	})
-
+	// TODO: Shouldn't we use "github.com/maistra/xns-informer/pkg/generated/kube/core/v1".NewFilteredSecretInformer with
+	// fieldSelector := fields.AndSelectors(
+	//     fields.OneTermNotEqualSelector("type", "helm.sh/release.v1"),
+	//     fields.OneTermNotEqualSelector("type", string(v1.SecretTypeServiceAccountToken))).String()
```
jewertow, line 85: It's not filtered in previous releases, so I wasn't sure if I should apply this change in rebase, because we may need additional tests... What do you think?

pkg/kube/client.go
jewertow, line 1: JFYI: I reduced the number of changes in this file by not returning xns-infomer's types from any function. Only internal fields use these types.

tests/binary/binaries_test.go
```diff
@@ -94,7 +94,7 @@ func TestBinarySizes(t *testing.T) {
 		"istioctl":    {60, 100},
 		"pilot-agent": {30, 47},
 		// TODO(https://github.com/kubernetes/kubernetes/issues/101384) bump this down a bit?
-		"pilot-discovery": {60, 99},
+		"pilot-discovery": {60, 101},
```
jewertow, line 97: JFYI: I had to change this value, because this test was failing.

tests/integration/servicemesh/managingroutes/main_test.go
```diff
@@ -0,0 +1,179 @@
...
+		Setup(maistra.RemoveDefaultRBAC).
+		Setup(maistra.ApplyRestrictedRBAC(namespace.Future(&istioNamespace))).
+		// We cannot disable webhooks in maistra.Install(), because then we would need maistra/istio-operator
+		// to properly patch CA bundles in the webhooks. To avoid that problem we restart Istio with disabled webhooks
+		// and without roles for managing webhooks once they are already created and patched.
+		Setup(maistra.DisableWebhooksAndRestart(namespace.Future(&istioNamespace))).
+		Run()
+}
+
+func TestManagingGateways(t *testing.T) {
```
jewertow, line 63: This test may seem to be redundant since we have multi-tenancy test, but it will be used to test IOR, so I didn't remove it.

go.mod
bartoszmajsak: I think the problem is that the default branch for the `maistra/api` repo is set to 2.4 and you want to import commits from 2.5.

Not quite sure why we have this branching strategy for api repo, having it linear (one `main` branch) with tags would probably make it work.

bartoszmajsak: It would be also nice to have proper versions so we don't need to use those `v0.0.0-*` :)

manifests/charts/istio-control/istio-discovery/templates/clusterrole.yaml
```diff
@@ -116,6 +116,14 @@ rules:
     resources: ["secrets"]
     verbs: ["get", "watch", "list"]
 
+  # Maistra specific
+  - apiGroups: ["maistra.io"]
+    resources: ["servicemeshmemberrolls"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["route.openshift.io"]
+    resources: ["routes", "routes/custom-host"]
+    verbs: ["get", "list", "watch", "create", "delete", "update"]
```
bartoszmajsak, line 125: I wonder - can't we have our roles instead which can be applied side by side with the upstream? This way we would just carry the files over instead of dealing with additions to existing upstream files. Perhaps not a thing for this PR, but maybe something worth considering?

pilot/pkg/bootstrap/configcontroller.go
bartoszmajsak: I obviously miss something, I wonder how `s.kubeClient.GetMemberRoll()` being not `nil` means we are in the multi-tenant mode?

pilot/pkg/bootstrap/server.go
```diff
@@ -574,6 +574,14 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
 		if err != nil {
 			return fmt.Errorf("failed creating kube client: %v", err)
 		}
+
+		memberRollName := args.RegistryOptions.KubeOptions.MemberRollName
+		if memberRollName != "" {
+			err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName)
+			if err != nil {
+				return fmt.Errorf("failed creating member roll: %v", err)
+			}
```
bartoszmajsak: Alternatively you can consider this:
```suggestion
			if err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName); err != nil {
				return fmt.Errorf("failed creating member roll: %v", err)
			}
```

pilot/pkg/config/kube/crdclient/client.go
```diff
@@ -161,20 +161,27 @@ func NewForSchemas(client kube.Client, opts Option, schemas collection.Schemas)
...
 	}
 	for _, s := range schemas.All() {
 		// From the spec: "Its name MUST be in the format <.spec.name>.<.spec.group>."
 		name := fmt.Sprintf("%s.%s", s.Plural(), s.Group())
 		if s.IsBuiltin() {
 			handleCRDAdd(out, name)
 		} else {
-			if _, f := known[name]; f {
+			// If EnableCRDScan is false, then ignore whether the CRD is in the map
+			// and just try to add informers for all types.
```
bartoszmajsak, line 183: Maybe explaining why we do it would help when I look back at this code in the next rebase round :)

pilot/pkg/config/kube/gateway/deploymentcontroller.go
bartoszmajsak: Maybe add a comment describing what you just pointed out. I personally would leave this `if` as in - a bit easier to follow for me.

pilot/pkg/config/kube/ingress/status.go
```diff
@@ -144,7 +156,7 @@ func (s *StatusSyncer) runningAddresses() []string {
 
 	for _, pod := range igPods {
 		// only Running pods are valid
-		if pod.Status.Phase != corev1.PodRunning {
+		if pod.Status.Phase != corev1.PodRunning || s.nodes == nil {
```
bartoszmajsak: `s.nodes == nil` what does it mean?

pilot/pkg/credentials/kube/secrets.go
bartoszmajsak: What do we gain from switching to `xns-informer`?

tests/binary/binaries_test.go
bartoszmajsak: Why 101? Why did the binary size grow?

pkg/kube/namespace/filter.go
```diff
@@ -240,3 +241,69 @@ func (d *discoveryNamespacesFilter) isSelected(labels labels.Set) bool {
...
+	// xns-informers make sure we only receive objects in member namespaces
+	return true
+}
+
+// FilterNamespace is not implemented, because Maistra does not filter namespaces;
+// this function should never be called if memberRollName is specified, so it's safe to call panic.
+func (d *maistraDiscoveryNamespacesFilter) FilterNamespace(nsMeta metav1.ObjectMeta) bool {
+	panic("Not implemented")
+}
+
```
bartoszmajsak, line 272: I understand that multi-tenant mode does not rely on label selectors, but we did introduce support for it in 2.4. Is it coming as the next PR or is it implemented somewhere else? Asking for a friend ;)

bartoszmajsak (commented):
LGTM, just a few questions and suggestions along the way :)

pkg/kube/client.go
```diff
@@ -339,22 +353,25 @@ type client struct {
...
 	istio         istioclient.Interface
-	istioInformer istioinformer.SharedInformerFactory
+	istioInformer xnsistioinformer.SharedInformerFactory
 
 	gatewayapi         gatewayapiclient.Interface
-	gatewayapiInformer gatewayapiinformer.SharedInformerFactory
+	gatewayapiInformer xnsgatewayapiinformer.SharedInformerFactory
 
-	started atomic.Bool
 	// If enabled, will wait for cache syncs with extremely short delay. This should be used only for tests
```
yannuil: Is this comment duplicate of [line 374](https://github.com/maistra/istio/pull/814/files#diff-137bba79988de442c1fde0fc29520d6d7e5a5239ceac82f5439a5bf513ebb2d4R374) at the updated file?

yannuil (commented):
LGTM. I can see there are a quite few `client.GetMemberRoll() == nil` check. Would it be good to extract to a method like `client.IsMultiTenant()` to improve the readability?

manifests/charts/istio-control/istio-discovery/templates/clusterrole.yaml
jewertow: Actually, we were changing roles only in the istio-operator in the past, but when I worked on https://github.com/maistra/istio/pull/458, I had to modify this file. Then we discussed how to handle changes in helm manifests and there was an idea to remove [patch-charts.sh](https://github.com/maistra/istio-operator/blob/maistra-2.4/build/patch-charts.sh), but we never moved those changes from istio-operator to istio.
I also considered removing this change, because now we apply custom roles from [tests/integration/servicemesh/maistra/testdata/roles.yaml](https://github.com/maistra/istio/blob/maistra-2.4/tests/integration/servicemesh/maistra/testdata/role.yaml), so we no longer need these roles here. If we will decided to do that, we will have to revert this [commit](https://github.com/maistra/istio-operator/pull/1042) as well.
@luksa could you also share your opinion?

pilot/pkg/bootstrap/configcontroller.go
jewertow: > I wonder how s.kubeClient.GetMemberRoll() being not nil means we are in the multi-tenant mode?

Being not nil means that `--memberRoll=default` argument was passed to istiod -> istiod watches ServiceMeshMemberRoll -> istiod runs in multi-tenant mode.

tests/binary/binaries_test.go
jewertow: I don't know. Maybe xns-informer and maistra api add these 2 bytes...?

pilot/pkg/bootstrap/server.go
jewertow: Thanks, I prefer this notation. I just modified only not working code, but I will change it as you suggest.

pilot/pkg/config/kube/ingress/status.go
jewertow: You can see below the following piece of code:
```
		// Find node external IP
		node := s.nodes.Get(pod.Spec.NodeName, "")
		if node == nil {
			continue
		}
```
First of all, `s.nodes` is not set in multi-tenant mode, because then istiod has no permissions to perform any operation on `Node` objects. So this condition ensures that we don't reach that peace of code and istiod will not try to get nodes when it has no proper permissions.
I agree that it's a little bit confusing, but I didn't care too much about readability, because I understand all these hacks... I can extract and comment this condition to make it more clear.

bartoszmajsak: I think a comment on `why` might be helpful.

go.mod
jewertow: It's a shame... but I just forgot to run `go mod tidy` and we don't need to change the default branch.

manifests/charts/istio-control/istio-discovery/templates/clusterrole.yaml
jewertow: I thought about it more and I would prefer to keep it here, even though I don't like it, because it's already in maistra/istio 2.2 - 2.4, so maybe it does not make sense to change it once again...

pilot/pkg/bootstrap/server.go
jewertow: Done

pkg/kube/namespace/filter.go
jewertow: Label selectors are supported only in cluster-wide mode.

pkg/kube/client.go
jewertow: Yes, look like a mistake. I removed the added comment.

pilot/pkg/config/kube/ingress/status.go
jewertow: I noticed that when `s.nodes == nil`, iterating over ingress pods does not make sense, because we will end up in `continue` in all iterations, so I returned earlier and skipped getting ingress pods.

pilot/pkg/credentials/kube/secrets.go
jewertow: `xns-informers` watches resources from the namespace set, not in all namespaces. If we used upstream informer, we would get permission errors like [here](https://github.com/maistra/istio/pull/803).

jewertow: Anyway, I will investigate it later.

pilot/pkg/config/kube/crdclient/client.go
jewertow: To be honest, I can't explain it better, but we will try to replace scanning CRDs with API discovery.

jewertow:
All comments addressed. Please review once again.

pkg/kube/client.go
```diff
@@ -167,6 +172,18 @@ type Client interface {
...
 	RegisterFilter(t reflect.Type, filter kubetypes.Filter) error
+
+	// SetNamespaces sets watched namespaces if no MemberRoll controller exists.
+	SetNamespaces(namespaces []string)
+
+	// AddMemberRoll creates a MemberRollController and adds it to the client.
+	AddMemberRoll(namespace, memberRollName string) error
+
+	// GetMemberRoll returns the member roll for the client, which may be nil.
+	GetMemberRoll() memberroll.MemberRollController
```
luksa: Can we rename this to `AddMemberRollController` and `GetMemberRollController` so that they aren't misleading. When I saw a call to `AddMemberRoll()`, I thought it was creating the SMMR resource.

pilot/pkg/bootstrap/server.go
```diff
@@ -574,6 +574,13 @@ func (s *Server) initKubeClient(args *PilotArgs) error {
 		if err != nil {
 			return fmt.Errorf("failed creating kube client: %v", err)
 		}
+
+		memberRollName := args.RegistryOptions.KubeOptions.MemberRollName
+		if memberRollName != "" {
+			if err := s.kubeClient.AddMemberRoll(args.Namespace, memberRollName); err != nil {
+				return fmt.Errorf("failed creating member roll: %v", err)
```
luksa: This message is also misleading.

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -186,23 +183,29 @@ func NewDeploymentController(client kube.Client, clusterID cluster.ID, env *mode
...
+		dc.namespaces.AddEventHandler(controllers.ObjectHandler(func(o controllers.Object) {
+			// TODO: make this more intelligent, checking if something we care about has changed
+			// requeue this namespace
+			for _, gw := range dc.gateways.List(o.GetName(), klabels.Everything()) {
+				dc.queue.AddObject(gw)
 			}
-		}
-	}))
+		}))
+		dc.gatewayClasses = kclient.New[*gateway.GatewayClass](client)
```
luksa, line 198: I wonder why we don't support GatewayClasses in multitenant mode. Do you know if it was because the gatewayclass controller in istiod requires create&update privileges for gatewayclasses? If that's the only reason, we could give istiod read-only permissions and disable the gatewayclass controller in multitenant mode.

luksa (approved):
LGTM but I'd fix the naming around memberroll controller

pilot/pkg/config/kube/gateway/deploymentcontroller.go
jewertow: > Do you know if it was because the gatewayclass controller in istiod requires create&update privileges for gatewayclasses?

I think it's because `GatewayClass` is cluster-scoped resource.

luksa: Yeah, I'm trying to remember why we thought having access to cluster-scoped resources was such a bad thing. For some resources (such as GatewayClass), having read-only privileges don't seem like such a bad thing. After all, the whole purpose of Gateway API is allowing regular users to create their own gateways. And for that, they need access to GatewayClasses.

pilot/pkg/bootstrap/server.go
jewertow: I changed it

pkg/kube/client.go
jewertow: Done

luksa (approved):

PR #816, opened 2023-07-06
OSSM-4390 With cert-manager, IOR can not list "Route" resource

yannuil:


yannuil:
Close in favor of #828 .

PR #828, opened 2023-07-31
OSSM-4390: sync Route informers before starting IOR controller

yannuil:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -225,7 +225,11 @@ func (s *Server) initK8SConfigStore(args *PilotArgs) error {
 			leaderelection.
 				NewLeaderElection(args.Namespace, args.PodName, leaderelection.IORController, args.Revision, s.kubeClient).
 				AddRunFunction(func(leaderStop <-chan struct{}) {
-					ior.Run(s.kubeClient, configController, leaderStop)
+					controller := ior.New(s.kubeClient, configController)
+
+					s.kubeClient.RunAndWaitRouteInformer(leaderStop)
+
+					controller.Run(leaderStop)
 				}).Run(stop)
 			return nil
 		})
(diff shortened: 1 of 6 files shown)


jewertow:
Could you explain this change? I don't understand what actually changes.

yannuil:
IOR uses `pilot.pkg.config.kube.crdclient` to manage the route informer's events. According to [this comment](https://github.com/maistra/istio/blob/maistra-2.4/pkg/kube/client.go#L606), after [getting the `Route` informer](https://github.com/maistra/istio/blob/maistra-2.4/pilot/pkg/config/kube/ior/controller.go#L79) in the IOR we need to start and wait for the informer cache sync. This changes make sure that the start and wait run after the calling the `.RouteInformer()` method.

pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -225,7 +225,11 @@ func (s *Server) initK8SConfigStore(args *PilotArgs) error {
 			leaderelection.
 				NewLeaderElection(args.Namespace, args.PodName, leaderelection.IORController, args.Revision, s.kubeClient).
 				AddRunFunction(func(leaderStop <-chan struct{}) {
-					ior.Run(s.kubeClient, configController, leaderStop)
+					controller := ior.New(s.kubeClient, configController)
+
+					s.kubeClient.RunAndWaitRouteInformer(leaderStop)
+
+					controller.Run(leaderStop)
```
jewertow, line 232: Could not it be simple like this?
```suggestion
                                        s.kubeClient.RunAndWait(stop)
					ior.Run(s.kubeClient, configController, leaderStop)
```
Then you could revert all other changes from this PR.

FilipB:
See attached reproducer in https://issues.redhat.com/browse/OSSM-5720 which will simplify the testing

FilipB:
What's the status? Do we want to test this before merge?

pilot/pkg/bootstrap/configcontroller.go
jewertow: Ok, now I see that it couldn't work, because `RunAndWait()` must be called after `.Informer()`.

jewertow (approved):

PR #849, opened 2023-09-04
Fix IOR wrongly copying argocd internal annotation from Gateway to route

JulienDelacroix:

When IOR creates a route from a Gateway, it copies all annotations, including annotations managed by argoCD (https://argo-cd.readthedocs.io/en/stable/).
As a consequence, when deploying the Gateway with argocd, the argocd.argoproj.io/instance (used by argoCD to track the objects it manages) is copied in the route. From this point ArgoCD constantly deletes the route (since it is not part of the desired manifest). This makes it impossible to deploy Gateway with argoCD, unless IOR automatic route creation is disabled.

This PR filter the argocd annotations when generating routes from Gateways.

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/ior/controller.go
+++ b/pilot/pkg/config/kube/ior/controller.go
@@ -126,6 +126,21 @@ func (r *routeController) deleteRoute(route *v1.Route) error {
 	return nil
 }
 
+func filteredRouteAnnotation(routeAnnotation string) bool {
+	filteredPrefixes := [...]string{
+		ShouldManageRouteAnnotation,
+		"kubectl.kubernetes.io",
+		"argocd.argoproj.io",
+	}
+
+	for _, prefix := range filteredPrefixes {
+		if strings.HasPrefix(routeAnnotation, prefix) {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


jewertow (approved):

reiniervl (commented):
This removes all the annotations, including:
`argocd.argoproj.io/compare-options: IgnoreExtraneous`
`argocd.argoproj.io/sync-options: Prune=false`

This means when tracking is done on a [label](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/#additional-tracking-methods-via-an-annotation) (the default), the Routes are removed. This change does not fix the issue, but does introduces a bug.

JulienDelacroix:
> argocd.argoproj.io/

I do not think that it introduces a bug. It makes no sense to copy  _argocd.argoproj.io/compare-options_, _argocd.argoproj.io/sync-options_ or any other argoCD annotations. Since the route is not part of the helmchart: there is nothing to compare it with.

You are right though that the original issue is not fixed, as labels should be filtered as well (especially the _agocd.argoproj.io/instance_ one). And as for annotations, I think all labels should be filtered.

I opened https://github.com/maistra/istio/pull/947 to filter labels as well.

PR #860, opened 2023-09-30
OSSM-4880: Rebase building OpenSSL-based proxy image

jewertow:


jwendell:
@jewertow maybe a legit error? `Need executable 'rpmbuild' to convert dir to rpm`

jewertow:
> @jewertow maybe a legit error? Need executable 'rpmbuild' to convert dir to rpm

It seems that `rpmbuild` is present in the builder image:
```
docker run quay.io/maistra-dev/maistra-builder:2.5 which rpmbuild
useradd: user 'user' already exists
/usr/bin/rpmbuild
```

jewertow:
Beside that, it works on my computer, but I don't have `rpmbuild`. I only have `rpm`.

yxun:
> un quay.io/maistra-dev/maistra-builder:2.5 which rpmbuild

Here is my finding:

> $ podman run quay.io/maistra-dev/maistra-builder:2.5 which rpmbuild
useradd: user 'user' already exists
which: no rpmbuild in (/rust/bin:/usr/lib/llvm/bin:/usr/local/go/bin:/gobin:/usr/local/google-cloud-sdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
$ 

I will add `rpmbuild` to our 2.5 builder image. Not clear why it exists from your command above.

prow/lib.sh
```diff
@@ -114,10 +114,10 @@ function build_images() {
   targets="docker.pilot docker.proxyv2 "
 
   # use ubuntu:jammy to test vms by default
-  nonDistrolessTargets="docker.app docker.app_sidecar_ubuntu_jammy docker.ext-authz "
+  nonDistrolessTargets="docker.app docker.app_sidecar_centos_stream_8 docker.ext-authz "
   if [[ "${JOB_TYPE:-presubmit}" == "postsubmit" ]]; then
     # We run tests across all VM types only in postsubmit
-    nonDistrolessTargets+="docker.app_sidecar_ubuntu_xenial docker.app_sidecar_debian_11  docker.app_sidecar_centos_7 "
+    nonDistrolessTargets+="docker.app_sidecar_ubuntu_xenial docker.app_sidecar_ubuntu_jammy docker.app_sidecar_debian_11 docker.app_sidecar_centos_7 "
```
yxun, line 120: Do we need to run this postsubmit target build ? It looks like that's building ubuntu and debian images. Should we skip that ?

yxun (approved):

istio.deps
```diff
@@ -4,7 +4,7 @@
     "name": "PROXY_REPO_SHA",
     "repoName": "proxy",
     "file": "",
-    "lastStableSHA": "491c2e174bc6022e6af8beec7ee678cf48bbd130"
+    "lastStableSHA": "ae3bbc4313b45af63777a2588388796d74221cfd"
```
jwendell, line 7: if you are going to push another commit into this PR, you could also update this SHA.

optional though. We can do it in a follow up

prow/lib.sh
jewertow: Actually, we could skip these images... But let's merge it as is for now and we can remove it in a follow-up.

istio.deps
jewertow: I will do it in a follow-up, because I want to unblock our e2e tests, and I want to avoid additional commits, because the CI is very unstable recently.

PR #884, opened 2023-11-04
OSSM-5312: Remove TagWatcher from gateway deployment controller

jewertow:
TagWatcher requires cluster-scoped permissions for MutatingWebhooks, so we can't support it in multi-tenant mode. On the other hand, it can't be used in cluster-wide mode as well, because TagWatcher watches MutatingWebhooks and relies on istio.io/rev label, which is not unique across control planes in OSSM. Additionally, TagWatcher causes skipping Gateways which do not have istio.io/rev label while the control plane is a revision different than "default".

### DIFF
--- a/pilot/pkg/bootstrap/configcontroller.go
+++ b/pilot/pkg/bootstrap/configcontroller.go
@@ -38,7 +38,6 @@ import (
 	"istio.io/istio/pkg/config/analysis/incluster"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/istio/pkg/config/schema/gvk"
-	"istio.io/istio/pkg/revisions"
 	"istio.io/pkg/log"
 )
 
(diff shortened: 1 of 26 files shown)


pilot/pkg/bootstrap/configcontroller.go
```diff
@@ -184,22 +183,14 @@ func (s *Server) initK8SConfigStore(args *PilotArgs) error {
...
 						// We can only run this if the Gateway CRD is created
 						if configController.WaitForCRD(gvk.KubernetesGateway, leaderStop) {
-							var tagWatcher revisions.TagWatcher
-							// TagWatcher requires permission for MutatingWebhook, so it can't be used in multi-tenant mode
-							if !s.kubeClient.IsMultiTenant() {
-								tagWatcher = revisions.NewTagWatcher(s.kubeClient, args.Revision)
-							}
 							controller := gateway.NewDeploymentController(s.kubeClient, s.clusterID, s.environment,
-								s.webhookInfo.getWebhookConfig, s.webhookInfo.addHandler, tagWatcher, args.Revision)
+								s.webhookInfo.getWebhookConfig, s.webhookInfo.addHandler, args.Revision)
```
luksa, line 187: Will the istio integration tests work if you always remove `tagWatcher`? Shouldn't this be configurable (i.e. we only disable tagWatcher for normal OSSM deployments, but not in integration tests)?

jewertow: > Will the istio integration tests work if you always remove tagWatcher?

Yes.

> Shouldn't this be configurable (i.e. we only disable tagWatcher for normal OSSM deployments, but not in integration tests)?

I thought about it, but it's problematic, because we would have to add yet another argument to istiod, so I think it's not worth to do it.

luksa (approved):

PR #923, opened 2024-01-26
[maistra-2.5] OSSM-4390: sync Route informers before starting IOR controller (#828)

jewertow:
This is a manual cherry-pick of #828.


dgn (approved):
lgtm as a cherry-pick

PR #951, opened 2024-03-05
OSSM-5556: Set `net.ipv4.ip_unprivileged_port_start=0` in ingress and egress gateways

jewertow:


dgn (approved):
LGTM, but we should push this change upstream as well

PR #958, opened 2024-03-11
OSSM-6079: Pass --resync flag to federation registry

jewertow:
Federation registry used hard coded 5 minutes long resync period and therefore imported and exported services were applied with a long delay. I guess that it wasn't hard coded intentionally, so I passed `--resync` flag to federation registry ticker to shorten the timeout in federation tests and make them faster and less flaky.

### DIFF
--- a/tests/integration/servicemesh/federation/discovery/discovery_test.go
+++ b/tests/integration/servicemesh/federation/discovery/discovery_test.go
@@ -58,7 +58,7 @@ var (
 	// Timeout is 6 minutes long, because we have hardcoded resync period 5 minutes long in the federation discovery controller.
 	defaultRetry = echo.Retry{
 		Options: []retry.Option{
-			retry.Timeout(6 * time.Minute),
+			retry.Timeout(30 * time.Second),
 			retry.Delay(1 * time.Second),
 		},
 	}
--- a/tests/integration/servicemesh/federation/discovery/discovery_test.go
+++ b/tests/integration/servicemesh/federation/discovery/discovery_test.go
@@ -55,7 +55,6 @@ var (
 		ServicePort: 7070,
 	}
 
-	// Timeout is 6 minutes long, because we have hardcoded resync period 5 minutes long in the federation discovery controller.
 	defaultRetry = echo.Retry{
 		Options: []retry.Option{
 			retry.Timeout(30 * time.Second),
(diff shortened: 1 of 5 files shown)


tests/integration/servicemesh/federation/ha/ha_test.go
```diff
@@ -55,7 +55,7 @@ var (
 	// Timeout is 6 minutes long, because we have hardcoded resync period 5 minutes long in the federation discovery controller.
```
dgn: nit: comment should be updated

tests/integration/servicemesh/federation/discovery/discovery_test.go
```diff
@@ -58,7 +58,7 @@ var (
 	// Timeout is 6 minutes long, because we have hardcoded resync period 5 minutes long in the federation discovery controller.
```
dgn: nit: comment should be updated

tests/integration/servicemesh/federation/discovery/discovery_test.go
```diff
@@ -58,7 +58,7 @@ var (
...
 	defaultRetry = echo.Retry{
 		Options: []retry.Option{
-			retry.Timeout(6 * time.Minute),
+			retry.Timeout(30 * time.Second),
```
dgn, line 60: should the timeout be based on resync period? maybe min(resync, 30s)?

dgn (approved):
lgtm, just minor comments

tests/integration/servicemesh/federation/ha/ha_test.go
jewertow: Thanks, I removed it

tests/integration/servicemesh/federation/discovery/discovery_test.go
jewertow: Thanks, I removed it

tests/integration/servicemesh/federation/discovery/discovery_test.go
jewertow: I noticed that the first test is sometimes slow and takes ~12 seconds, so there must be also other factors to tweak. Our patches rarely affect federation, so these tests should not fail, and therefore I think 30 seconds long timeout is a good trade-off.

PR #983, opened 2024-04-15
[2.6] Rebase IOR

yannuil:


jewertow:
@yannuil please remember about including https://github.com/maistra/istio/pull/980.

jewertow (commented):
@yannuil could you please:
1. remove commit "Fix flaky IOR integration test" as it still fails;
2. squash argocd-related commits into "OSSM-6295: Copy all annotations and labels added by ArgoCD except `argocd.argoproj.io/instance`"?

Then I would approve this PR and we could merge it.

jewertow (commented):
You forgot about this: https://github.com/maistra/istio/pull/828.

jewertow (approved):

yannuil:
@jewertow , actually don't need it anymore. The `route` resource is now managed by the upstream informer factory, which has a single method to start all informers.

PR #997, opened 2024-05-18
[2.6] Rebase federation

yannuil:


jewertow (commented):
Looks good, but you could squash these commits as the first one does not work without the subsequent commits, and rebase this PR once https://github.com/maistra/istio/pull/1001 is merged.

jewertow (approved):

PR #1004, opened 2024-05-21
OSSM-5698 Rebase injection proxy UID/GID

mayleighnmyers:
OSSM-5698 Rebase injection proxy UID/GID

**To help us figure out who should review this PR, please put an X in all the areas that this PR affects.**

- [ ] Ambient
- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

**Please check any characteristics that apply to this pull request.**

- [ ] Does not have any [user-facing](https://github.com/istio/istio/tree/master/releasenotes#when-to-add-release-notes) changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

PR #1009, opened 2024-05-28
OSSM-5698 Rebase injection proxy UID/GID

mayleighnmyers:
[proxy-uid] OSSM-2292: Run sidecars with dynamically-determined user IDs (#689)

* [proxy-uid] MAISTRA-551: Run sidecars with dynamically-determined user IDs (#230) (#541)

  * MAISTRA-1751: Set securityContext.fsGroup to comply with restricted SCC (#256)

  This is a rewrite of 46f0f4a819 for the Maistra 2.1 / Istio 1.8 rebase.

  * MAISTRA-2452 Apply dynamically-determined user ID to the istio-validation init container

* [maistra-2.3] OSSM-1847 ensure ProxyUID and ProxyGID are initialized when using gat… (#618)

* OSSM-1847 ensure ProxyUID and ProxyGID are initialized when using gateway injection template

* do not default ProxyUID and ProxyGID for gateways

* initialize ProxyGID appropriately for gateway injection

* fix(gen): runs make gen

* chore: combines cni plugins golden files gen into one invocation

[maistra-2.4] OSSM-2324 Ensure ProxyUID and ProxyGID are never nil (#703)

* OSSM-2324 Ensure ProxyUID and ProxyGID are never nil

If they are nil, Helm renders them as "nil" (with quotes), which the YAML parser can't parse as a float.

* Remove redundant loop

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/pilot/pkg/config/kube/gateway/deploymentcontroller.go
+++ b/pilot/pkg/config/kube/gateway/deploymentcontroller.go
@@ -50,6 +50,7 @@ import (
 	"istio.io/istio/pkg/test/util/tmpl"
 	"istio.io/istio/pkg/test/util/yml"
 	"istio.io/istio/pkg/util/sets"
+	iptablesconstants "istio.io/istio/tools/istio-iptables/pkg/constants"
 )
 
 // DeploymentController implements a controller that materializes a Gateway into an in cluster gateway proxy
@@ -323,12 +324,6 @@ func (d *DeploymentController) configureIstioGateway(log *istiolog.Scope, gw gat
 	}
 	log.Info("reconciling")
 
-	var ns *corev1.Namespace
-	if d.namespaces != nil {
-		ns = d.namespaces.Get(gw.Namespace, "")
-	}
-	proxyUID, proxyGID := inject.GetProxyIDs(ns)
-
 	defaultName := getDefaultName(gw.Name, &gw.Spec)
 
 	serviceType := gi.defaultServiceType
@@ -346,8 +341,8 @@ func (d *DeploymentController) configureIstioGateway(log *istiolog.Scope, gw gat
 		KubeVersion122: kube.IsAtLeastVersion(d.client, 22),
 		Revision:       d.revision,
 		ServiceType:    serviceType,
-		ProxyUID:       proxyUID,
-		ProxyGID:       proxyGID,
+		ProxyUID:       iptablesconstants.DefaultProxyUIDInt,
+		ProxyGID:       iptablesconstants.DefaultProxyUIDInt,
 	}
 
 	if overwriteControllerVersion {
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	yamlDecoder "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/yaml"
 
 	"istio.io/api/annotation"
@@ -53,6 +54,13 @@ import (
 // InjectionPolicy determines the policy for injecting the
 // sidecar proxy into the watched namespace(s).
 type InjectionPolicy string
+
+// Defaults values for injecting istio proxy into kubernetes
+// resources.
+const (
+	DefaultSidecarProxyUID = int64(1337)
+	DefaultSidecarProxyGID = int64(1337)
+)
 
 const (
 	// InjectionPolicyDisabled specifies that the sidecar injector
@@ -776,6 +784,8 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig Value
 			revision:            revision,
 			proxyEnvs:           map[string]string{},
 			injectedAnnotations: nil,
+			proxyUID:            ptr.To(DefaultSidecarProxyUID),
+			proxyGID:            ptr.To(DefaultSidecarProxyGID),
 		}
 		patchBytes, err = injectPod(params)
 	}
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -55,13 +55,6 @@ import (
 // sidecar proxy into the watched namespace(s).
 type InjectionPolicy string
 
-// Defaults values for injecting istio proxy into kubernetes
-// resources.
-const (
-	DefaultSidecarProxyUID = int64(1337)
-	DefaultSidecarProxyGID = int64(1337)
-)
-
 const (
 	// InjectionPolicyDisabled specifies that the sidecar injector
 	// will not inject the sidecar into resources by default for the
@@ -407,8 +400,6 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 		return nil, nil, err
 	}
 
-	proxyUID, proxyGID := GetProxyIDs(params.namespace)
-
 	data := SidecarTemplateData{
 		TypeMeta:         params.typeMeta,
 		DeploymentMeta:   params.deployMeta,
@@ -419,8 +410,8 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 		Values:           params.valuesConfig.asMap,
 		Revision:         params.revision,
 		ProxyImage:       ProxyImage(params.valuesConfig.asStruct, params.proxyConfig.Image, strippedPod.Annotations),
-		ProxyUID:         proxyUID,
-		ProxyGID:         proxyGID,
+		ProxyUID:         *params.proxyUID,
+		ProxyGID:         *params.proxyGID,
 		CompliancePolicy: common_features.CompliancePolicy,
 	}
 
@@ -784,8 +775,8 @@ func IntoObject(injector Injector, sidecarTemplate Templates, valuesConfig Value
 			revision:            revision,
 			proxyEnvs:           map[string]string{},
 			injectedAnnotations: nil,
-			proxyUID:            ptr.To(DefaultSidecarProxyUID),
-			proxyGID:            ptr.To(DefaultSidecarProxyGID),
+			proxyUID:            ptr.To(constants.DefaultProxyUIDInt),
+			proxyGID:            ptr.To(constants.DefaultProxyUIDInt),
 		}
 		patchBytes, err = injectPod(params)
 	}
--- a/pkg/kube/inject/testdata/inject/hello-tproxy.yaml.injected
+++ b/pkg/kube/inject/testdata/inject/hello-tproxy.yaml.injected
@@ -144,7 +144,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsGroup: 1337
           runAsNonRoot: false
-          runAsUser: 0
+          runAsUser: 1337
         startupProbe:
           failureThreshold: 600
           httpGet:
--- a/pkg/kube/inject/testdata/inject/proxy-override-runas.yaml.cni.injected
+++ b/pkg/kube/inject/testdata/inject/proxy-override-runas.yaml.cni.injected
@@ -138,7 +138,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsGroup: 4321
           runAsNonRoot: true
-          runAsUser: 1234
+          runAsUser: 1337
         startupProbe:
           failureThreshold: 600
           httpGet:
@@ -171,7 +171,7 @@ spec:
         - -z
         - "15006"
         - -u
-        - "1234"
+        - "1337"
         - -m
         - REDIRECT
         - -i
@@ -203,7 +203,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsGroup: 4321
           runAsNonRoot: true
-          runAsUser: 1234
+          runAsUser: 1337
       volumes:
       - name: workload-socket
       - name: credential-socket
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -1166,7 +1166,7 @@ func (wh *Webhook) inject(ar *kube.AdmissionReview, path string) *kube.Admission
 		}
 	} else if tproxyInterceptionMode {
 		proxyUID = ptr.To(int64(0))
-		proxyGID = ptr.To(DefaultSidecarProxyGID)
+		proxyGID = ptr.To(iptablesconstants.DefaultProxyUIDInt)
 	} else {
 		// we're injecting a normal pod (with app container and optional sidecar container)
 		// we set proxyUID to the main app container's UID incremented by 1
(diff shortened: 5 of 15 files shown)


yxun:
Would you like to close the previous #1004 PR ?

go.mod
```diff
@@ -1,6 +1,8 @@
 module istio.io/istio
 
-go 1.21
+//go:built go 1.22
+
+toolchain go1.22.3
```
yxun: The upstream 1.20 is built by go 1.21 :
https://github.com/istio/istio/blob/release-1.20/go.mod

I don't think we should upgrade it here to 1.22.3

pkg/kube/inject/webhook_test.go
```diff
@@ -946,6 +946,63 @@ func TestRunAndServe(t *testing.T) {
...
+    }
+},
+{
+    "op": "add",
+    "path": "/spec/containers/1",
+    "value": {
+        "name": "istio-proxy",
+        "resources": {},
+		"securityContext": {
+		    "runAsUser": 1337
```
yxun: ~We need this runAsUser 1337 test here. 
We may not need other `op` tests if they have been removed from upstream.~

go.mod
mayleighnmyers: I changed it back to 1.21

jwendell:
I wonder if this is necessary, since in Istio 1.20 we already have this feature on upstream...

yannuil (approved):
LGTM

yannuil:
@jwendell are you talking about [056d019](https://github.com/istio/istio/commit/056d019350d7aa0232b35b274bd8981b10110ac8)? It is only available since 1.21. Maistra-2.6 is based on 1.20.

jwendell:
@yannuil https://github.com/istio/istio/pull/45394 and follow ups

jewertow:
@jwendell I think we need this change, because your upstream implementation requires permissions for namespaces.

jewertow (changes requested):
Please include recent fixes for TPROXY: https://github.com/maistra/istio/pull/991

eoinfennessy:
> @jwendell I think we need this change, because your upstream implementation requires permissions for namespaces.

If I understand correctly, we should not be using the `GetProxyIDs` function defined in `pkg/kube/inject/inject.go`.

```
// GetProxyIDs returns the UID and GID to be used in the RunAsUser and RunAsGroup fields in the template
// Inspects the namespace metadata for hints and fallbacks to the usual value of 1337.
func GetProxyIDs(namespace *corev1.Namespace) (uid int64, gid int64) {
	uid = constants.DefaultProxyUIDInt
	gid = constants.DefaultProxyUIDInt
	if namespace == nil {
		return
	}
	// Check for OpenShift specifics and returns the max number in the range specified in the namespace annotation
	if _, uidMax, err := getPreallocatedUIDRange(namespace); err == nil {
		uid = *uidMax
	}
	if groups, err := getPreallocatedSupplementalGroups(namespace); err == nil && len(groups) > 0 {
		gid = groups[0].Max
	}
	return
}
```

If this is the case, the following lines (410-425) in the same file need to be changed to instead use the constants `DefaultSidecarProxyUID` and `DefaultSidecarProxyGID`:

```
	proxyUID, proxyGID := GetProxyIDs(params.namespace)

	data := SidecarTemplateData{
		TypeMeta:         params.typeMeta,
		DeploymentMeta:   params.deployMeta,
		ObjectMeta:       strippedPod.ObjectMeta,
		Spec:             strippedPod.Spec,
		ProxyConfig:      params.proxyConfig,
		MeshConfig:       meshConfig,
		Values:           params.valuesConfig.asMap,
		Revision:         params.revision,
		ProxyImage:       ProxyImage(params.valuesConfig.asStruct, params.proxyConfig.Image, strippedPod.Annotations),
		ProxyUID:         proxyUID,
		ProxyGID:         proxyGID,
		CompliancePolicy: common_features.CompliancePolicy,
	}
```

jwendell:
@jewertow @eoinfennessy It's still not clear to me why we can't rely on upstream solution. Can you tell me more?

Is it because pilot reads namespace annotations and thus requires namespace reading permissions? Doesn't it already have this permission for all SMMR members? If not, can't we just add this permission?

eoinfennessy:
> @jewertow @eoinfennessy It's still not clear to me why we can't rely on upstream solution. Can you tell me more?

@jwendell, sorry, I'm also unsure why we can't use the changes that you contributed upstream, or what exactly the namespace permissions issue is.

jewertow:
> Doesn't it already have this permission for all SMMR members? If not, can't we just add this permission?

@jwendell no, we can't grant permissions for particular namespaces, because it's a cluster-scoped resource. You can grant access to all of them or none.

luksa:
We actually can grant permissions to specific `Namespace` objects (using `ClusterRole.rules.resourceNames`), but I'm not sure if we should. We definitely can't grant write permissions, but even read permissions might be problematic in some cases. Users typically don't have access to the namespaces (not even read-only access). We should definitely consult with the security folks about this.

On a side note, if it turns out we can give read access to the namespaces, we can remove the locality controller (the one that copies Namespace labels to Pods). 

Actually, if we *can't* give read access to the namespaces, we could extend the locality controller so that it also copies the userid range from the Namespace to the Pod so that the injection webhook could get the information from the Pod instead of the Namespace.

jewertow:
@luksa thanks for correcting me.

I am still not sure that it's worth to do it, because we would have to implement additional logic in the operator to apply proper cluster role bindings on SMMR changes. Additionally, xns-informer knows only the namespace names, so we would have to change the namespace controller and all other components that rely on it.

luksa:
@jewertow, ah, of course, the Namespace object is retrieved from the informer/lister, which means giving GET permissions only to specific Namespaces won't suffice, since the lister performs a LIST operation.

Anyway, in Sail, we don't have this problem, since istiod has the permission to list namespaces, so whatever solution we come up with, is only needed in 2.6 (and only in MultiTenant mode). And since we already have a solution for this in this PR, we might as well merge it.

pkg/kube/inject/testdata/inject/proxy-override-runas.yaml.cni.injected
```diff
@@ -138,7 +138,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsGroup: 4321
           runAsNonRoot: true
-          runAsUser: 1234
+          runAsUser: 1337
```
jwendell, line 141: Why do we have to change these files?

mayleighnmyers: This was included in the cherry-pick from the rebase in 2.5

jwendell: I think this is a new file in 1.20, it was not present on OSSM 2.5.

mayleighnmyers: https://github.com/maistra/istio/pull/853/files

jwendell: This file in particular is new in 1.20/2.6, no?

mayleighnmyers: Hmm, yes it is new (not in 1.19,2.5)

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -346,8 +340,8 @@ func (d *DeploymentController) configureIstioGateway(log *istiolog.Scope, gw gat
 		KubeVersion122: kube.IsAtLeastVersion(d.client, 22),
 		Revision:       d.revision,
 		ServiceType:    serviceType,
-		ProxyUID:       proxyUID,
-		ProxyGID:       proxyGID,
+		ProxyUID:       inject.DefaultSidecarProxyUID,
```
jwendell: I think we can re-use `constants.DefaultProxyUIDInt` ?

pkg/kube/inject/inject.go
```diff
@@ -47,13 +48,19 @@ import (
...
 )
 
 // InjectionPolicy determines the policy for injecting the
 // sidecar proxy into the watched namespace(s).
 type InjectionPolicy string
 
+// Defaults values for injecting istio proxy into kubernetes
+// resources.
+const (
+	DefaultSidecarProxyUID = int64(1337)
```
jwendell: Remove these in favor of `constants.DefaultProxyUIDInt`?

pkg/kube/inject/inject.go
```diff
@@ -868,24 +875,3 @@ func updateClusterEnvs(container *corev1.Container, newKVs map[string]string) {
 	}
 	container.Env = envVars
 }
-
-// GetProxyIDs returns the UID and GID to be used in the RunAsUser and RunAsGroup fields in the template
-// Inspects the namespace metadata for hints and fallbacks to the usual value of 1337.
-func GetProxyIDs(namespace *corev1.Namespace) (uid int64, gid int64) {
```
jwendell: You don't need to remove this code

mayleighnmyers: keep the full function?

jwendell: yes

pkg/kube/inject/testdata/inject/proxy-override-runas.yaml.cni.injected
jwendell: Anyway, it's confusing why we need to change this golden file. The input file explicitly overrides these values and expect them to not be touched: https://github.com/maistra/istio/blob/f1c1c5d98915b343c06f10b189eac6b24e74de23/pkg/kube/inject/testdata/inject/proxy-override-runas.yaml#L19-L22

jwendell (commented):
We probably need to remove/comment these lines:

https://github.com/maistra/istio/blob/f1c1c5d98915b343c06f10b189eac6b24e74de23/pkg/kube/inject/webhook.go#L205-L209

So that the namespace controller it not created.

pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -323,12 +324,6 @@ func (d *DeploymentController) configureIstioGateway(log *istiolog.Scope, gw gat
 	}
 	log.Info("reconciling")
 
-	var ns *corev1.Namespace
-	if d.namespaces != nil {
-		ns = d.namespaces.Get(gw.Namespace, "")
-	}
-	proxyUID, proxyGID := inject.GetProxyIDs(ns)
```
jewertow, line 330: This change breaks Gateway API in cluster-wide mode, e.g. gateway controller.

dgn: +1 I think this code shouldn't be removed but only surrounded by an if-clause that checks whether we're running in multi-tenant mode. In cluster-wide, behavior must match upstream ie this code block must be executed as usual

pkg/kube/inject/inject.go
```diff
@@ -399,8 +400,6 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 		return nil, nil, err
 	}
 
-	proxyUID, proxyGID := GetProxyIDs(params.namespace)
```
dgn, line 402: this call should not be removed, we want this behavior in cluster-wide deployments. To make it work in multitenant mode, we should wrap it in an if-clause that uses the other method (derive UID/GID from pod, `GetProxyIDsFromPod()`)

pkg/kube/inject/webhook.go
```diff
@@ -1128,6 +1135,75 @@ func (wh *Webhook) inject(ar *kube.AdmissionReview, path string) *kube.Admission
 	}
 
 	proxyConfig := wh.env.GetProxyConfigOrDefault(pod.Namespace, pod.Labels, pod.Annotations, wh.meshConfig)
+
+	var err error
```
dgn, line 1139: all of this code should go into a function `GetProxyIDsFromPod()` and it shouldn't be called here but in the same place where we call `GetProxyIDs()` - ie in `RunTemplate()`

pkg/kube/inject/testdata/inject/hello-tproxy.yaml.injected
```diff
@@ -144,7 +144,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsGroup: 1337
           runAsNonRoot: false
-          runAsUser: 0
+          runAsUser: 1337
```
dgn, line 147: we should not have to change this testdata as cluster-wide should be considered the default deployment model and that should behave exactly like upstream. we only need the solution that doesn't require namespace privileges for multitenant mode

pkg/kube/inject/webhook.go
```diff
@@ -1305,3 +1383,52 @@ func handleError(message string) {
 	log.Errorf(message)
 	totalFailedInjections.Increment()
 }
+
+// modifies the pod to ensure that the proxy runs with the given proxyUID instead of 1337
+func replaceProxyRunAsUserID(pod *corev1.Pod, proxyUID *int64) {
```
dgn, line 1388: usage of this func should not be required anymore as all injection templates now have the `ProxyUID` and `ProxyGID` variables which are inserted during injection

pkg/kube/inject/webhook.go
```diff
@@ -755,6 +760,8 @@ func postProcessPod(pod *corev1.Pod, injectedPod corev1.Pod, req InjectionParame
 		return err
 	}
 
+	replaceProxyRunAsUserID(pod, req.proxyUID)
```
dgn, line 763: I don't think we need to call this anymore, the func can be removed IIUC (see below)

dgn (commented):
I think we'll fundamentally need a rewrite of this patch that aligns it with the changes we made upstream. While the upstream changes are perfectly fine for cluster-wide mode, we cannot use them in multitenant mode as we don't have namespace permissions. So we need to do two things:
- make sure that behavior is unchanged for cluster-wide mode
- use the 'old' logic to derive uid/gid from podTemplate if we're running in multitenant mode

As our old patch was written long ago when things were very different, we need to revisit it and change it in a few places. Some of the changes we made for upstream to support uid/gid injection we can reuse - e.g. all the injection templates now have the ProxyUID/ProxyGID fields, so no need to do postprocessing on the pod after injection, we can just use the normal way through the templating engine. We just have to be careful to make sure that when any pod is created, either by the user or by the GatewayController, it must not have a securityContext.runAsUser defined so that OCP will fill in the correct UID. After briefly looking at https://github.com/maistra/istio/blob/f1c1c5d98915b343c06f10b189eac6b24e74de23/manifests/charts/istio-control/istio-discovery/files/kube-gateway.yaml#L48-L51 (which is used to create gateways by the Gateway Controller) we're looking good on that front.

As part of the rewrite, we should remove the old commit message as it's become unreadable.

dgn:
Superseded by https://github.com/maistra/istio/pull/1019

jwendell:
Safe to close this one, right?

@mayleighnmyers Thank you for your effort here!

PR #1019, opened 2024-06-13
OSSM-5698 Add GetProxyIDsFromPod() func for multitenant deployments

dgn:
In multitenant environments, we cannot expect to have access to namespaces to derive valid UID/GIDs, so we'll need to take whatever OCP injects into the securityContext and work with that. The reason this works is that if no securityContext is provided by the user, OCP will create one with a valid `runAsUser` value that is in the expected range.

I could only test this together with the 2.5 operator (there's no 2.6 operator yet) which has different injection templates, but it seemed to work.

### DIFF
--- a/pkg/kube/inject/inject.go
+++ b/pkg/kube/inject/inject.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	yamlDecoder "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/yaml"
 
 	"istio.io/api/annotation"
@@ -44,6 +45,7 @@ import (
 	proxyConfig "istio.io/api/networking/v1beta1"
 	opconfig "istio.io/istio/operator/pkg/apis/istio/v1alpha1"
 	"istio.io/istio/pilot/pkg/features"
+	"istio.io/istio/pilot/pkg/model"
 	"istio.io/istio/pkg/config/mesh"
 	common_features "istio.io/istio/pkg/features"
 	"istio.io/istio/pkg/log"
@@ -73,6 +75,9 @@ const (
 const (
 	// ProxyContainerName is used by e2e integration tests for fetching logs
 	ProxyContainerName = "istio-proxy"
+
+	// ProxyUIDAnnotation is an annotation that can be used to manually specify a proxy UID
+	ProxyUIDAnnotation = "sidecar.istio.io/proxyUID"
 
 	// ValidationContainerName is the name of the init container that validates
 	// if CNI has made the necessary changes to iptables
@@ -398,8 +403,13 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 	if err != nil {
 		return nil, nil, err
 	}
+	var proxyUID, proxyGID int64
 
-	proxyUID, proxyGID := GetProxyIDs(params.namespace)
+	if params.namespace != nil {
+		proxyUID, proxyGID = GetProxyIDs(params.namespace)
+	} else {
+		proxyUID, proxyGID = GetProxyIDsFromPod(params.pod)
+	}
 
 	data := SidecarTemplateData{
 		TypeMeta:         params.typeMeta,
@@ -449,7 +459,6 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 			return nil, nil, fmt.Errorf("failed parsing generated injected YAML (check Istio sidecar injector configuration): %v", err)
 		}
 	}
-
 	return mergedPod, templatePod, nil
 }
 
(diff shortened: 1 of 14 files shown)


pkg/kube/inject/inject.go
```diff
@@ -889,3 +898,61 @@ func GetProxyIDs(namespace *corev1.Namespace) (uid int64, gid int64) {
...
+
+// GetProxyIDsFromPod returns the UID and GID to be used in the RunAsUser and RunAsGroup fields in the template
+// It relies on OpenShift injecting a valid UID into the pod template which we can then reuse.
+func GetProxyIDsFromPod(pod *corev1.Pod) (uid int64, gid int64) {
+	var proxyUID, proxyGID *int64
+	tproxyInterceptionMode := pod.Annotations != nil && pod.Annotations["sidecar.istio.io/interceptionMode"] == string(model.InterceptionTproxy)
+	if tproxyInterceptionMode {
+		proxyUID = ptr.To(int64(0))
+		proxyGID = ptr.To(constants.DefaultProxyUIDInt)
+	} else {
```
dgn, line 910: note that I massively simplified the old function. Previously, we would look at the pod after injection, so we could distinguish between gateway and application cases based on number of containers and container names. This isn't really possible anymore as this now feeds into the normal injection (so we're looking at the pod before it is injected), but I think it's also not necessary.

pkg/kube/inject/inject.go
```diff
@@ -398,8 +403,13 @@ func RunTemplate(params InjectionParameters) (mergedPod *corev1.Pod, templatePod
 	if err != nil {
 		return nil, nil, err
 	}
+	var proxyUID, proxyGID int64
 
-	proxyUID, proxyGID := GetProxyIDs(params.namespace)
+	if params.namespace != nil {
```
jwendell, line 408: Not sure if it's in the scope of this PR, but we should also adapt the lines below so the namespace client is not created:

https://github.com/maistra/istio/blob/f1c1c5d98915b343c06f10b189eac6b24e74de23/pkg/kube/inject/webhook.go#L205-L209

dgn: good catch. interestingly it worked for me without changing that, but I guess that's because the 2.5 values don't set the platform field

dgn: made the change. PTAL

jwendell: `GetProxyIDs()` already checks for `namespace != nil`. Maybe we should call `GetProxyIDsFromPod()` in it?

pkg/kube/inject/inject.go
```diff
@@ -889,3 +898,61 @@ func GetProxyIDs(namespace *corev1.Namespace) (uid int64, gid int64) {
 
 	return
 }
+
+// GetProxyIDsFromPod returns the UID and GID to be used in the RunAsUser and RunAsGroup fields in the template
+// It relies on OpenShift injecting a valid UID into the pod template which we can then reuse.
+func GetProxyIDsFromPod(pod *corev1.Pod) (uid int64, gid int64) {
+	var proxyUID, proxyGID *int64
+	tproxyInterceptionMode := pod.Annotations != nil && pod.Annotations["sidecar.istio.io/interceptionMode"] == string(model.InterceptionTproxy)
```
jwendell: nit: Use the constant `annotation.SidecarInterceptionMode.Name` instead of hardcoding it.

`import "istio.io/api/annotation"`

pkg/kube/inject/inject.go
```diff
@@ -889,3 +898,61 @@ func GetProxyIDs(namespace *corev1.Namespace) (uid int64, gid int64) {
...
+		proxyGID = ptr.To(constants.DefaultProxyUIDInt)
+	} else {
+		var err error
+		proxyUID, err = getProxyUIDFromAnnotation(*pod)
+		if err != nil {
+			log.Infof("Could not get proxyUID from annotation: %v", err)
+		}
+		if proxyUID == nil {
+			if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.RunAsUser != nil {
+				proxyUID = ptr.To(*pod.Spec.SecurityContext.RunAsUser + 1)
```
jwendell, line 918: I know that this is the current state and we are unlikely to change it in 2.6, but this `+ 1` thing is fragile if the current value is the last in the range. I guess this is fine, plus hopefully 2.6 is the last release with it.

jwendell (approved):
LGTM, just a few nits, feel free to merge this as-is.

pkg/kube/inject/inject.go
dgn: yeah I kept thinking the same thing, but I don't think we should invest too much time in this now as it really is the last time we release it (and it's been fine so far)

pkg/kube/inject/inject.go
dgn: hmm.. that would mean changing the func signature to include the `corev1.Pod` as well as the `corev1.Namespace`. so we would need to change both the func and the code block that's calling it. I think it's simpler this way

dgn:
/hold

Putting a hold so that @jewertow has time to review

luksa:
BTW: I just tested this with the 2.6 operator and (at least the happy path) works as expected.

jewertow (approved):

PR #1022, opened 2024-06-13
OSSM-6615 Make sure IOR ignores services without selectors

luksa:
IOR finds a matching service for a pod by checking each service's label selector. However, this also matches services with an empty selector, which is wrong. These services should be ignored by IOR since an empty label selector indicates that the service endpoints are managed externally and not that the service should match all pods.


jwendell (approved):
worth a unit test case for this?

PR #1023, opened 2024-06-14
Prevent non-error from showing up in CI logs

luksa:
Prevent CI from highlighting the following as an error, because it's not an actual error:

error: taint "node-role.kubernetes.io/control-plane" not found


jwendell (approved):
This file comes from common-files repo, meaning there's potential merge conflicts.

I think it's fine given 2.x will rest in peace soon :)

PR #1038, opened 2024-06-28
OSSM-6694 Fix federation's resourceManager.HasSynced() and flaky TestStatusManager

luksa:
The federation controller's HasSynced() function only checked if some of its informers are synced instead of checking all of them.

The TestStatusManager test didn't use this HasSynced(), but called the HasSynced() function of some informers directly. It didn't check all informers that it should have, most importantly the pod informer.

This caused the test to be flaky, because the removeDeadPods function, which removes PodPeerDiscoveryStatuses associated with non-existent pods, erroneously removed the status that the test was checking. This happened when the podInformer wasn't yet synced and made it look as though the pod didn't exist.

### DIFF
--- a/pkg/servicemesh/federation/status/manager_test.go
+++ b/pkg/servicemesh/federation/status/manager_test.go
@@ -694,14 +694,9 @@ func TestStatusManager(t *testing.T) {
 	log.Configure(logOpts)
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			newCrdWatcher := kube.NewCrdWatcher
-			kube.NewCrdWatcher = kube.NewFastCrdWatcher
 			kubeClient := kube.NewFakeClient(&corev1.Pod{
 				ObjectMeta: istiodName,
 			})
-			kube.NewCrdWatcher = newCrdWatcher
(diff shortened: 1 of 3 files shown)


pkg/servicemesh/federation/status/manager_test.go
```diff
@@ -734,10 +731,10 @@ func TestStatusManager(t *testing.T) {
 				close(stopChan)
 				t.Fatalf("timed out waiting for leader election")
 			}
-			for !rm.ExportsInformer().Informer().HasSynced() ||
-				!rm.ImportsInformer().Informer().HasSynced() ||
-				!rm.PeerInformer().Informer().HasSynced() {
+
+			for !rm.HasSynced() {
```
jwendell: Can this cause an infinite loop or consume too much cpu for a certain amount of time?

Maybe surround it with retry.UntilSuccess with some timeout and deley between retries?

jwendell (approved):

pkg/servicemesh/federation/status/manager_test.go
luksa: Yeah, you're right. This tight loop was here before this PR, but I should have fixed it.

PR #1037, opened 2024-06-27
Two minor test fixes

luksa:

PR #1041, opened 2024-07-02
Ensure tests/integration/servicemesh/managingroutes/main_test.go can be run multiple times

luksa:


jwendell (approved):
dejavu? I think I already saw this fix before 🤔

luksa:
@jwendell yeah, but either this or the other commit in #1037 is causing an integration test to fail, so I'm splitting that PR into two to see which one.

PR #1050, opened 2024-07-12
OSSM-6812: Do not set cipher suites by default

jewertow:
We don't have to set default cipher suites, because Envoy-OpenSSL can detect proper defaults.

### DIFF
--- a/pilot/pkg/security/authn/utils/utils.go
+++ b/pilot/pkg/security/authn/utils/utils.go
@@ -75,9 +75,9 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
 		ciphers = mc.MeshMTLS.CipherSuites
 	}
 	if ciphers == nil {
-		ciphers = tls_features.TLSCipherSuites.Get()
-		if len(ciphers) == 0 {
-			ciphers = SupportedCiphers
+		opensslCiphers := tls_features.TLSCipherSuites.Get()
+		if len(opensslCiphers) > 0 {
+			ciphers = opensslCiphers
 		}
 	}
 	// Set Minimum TLS version to match the default client version and allowed strong cipher suites for sidecars.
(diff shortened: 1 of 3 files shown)


pilot/pkg/security/authn/utils/utils.go
```diff
@@ -75,9 +75,9 @@ func BuildInboundTLS(mTLSMode model.MutualTLSMode, node *model.Proxy,
 		ciphers = mc.MeshMTLS.CipherSuites
 	}
 	if ciphers == nil {
-		ciphers = tls_features.TLSCipherSuites.Get()
-		if len(ciphers) == 0 {
-			ciphers = SupportedCiphers
+		opensslCiphers := tls_features.TLSCipherSuites.Get()
+		if len(opensslCiphers) > 0 {
```
dgn, line 79: I don't think we need the if-clause anymore

pilot/pkg/security/authn/utils/utils_test.go
```diff
@@ -69,10 +69,6 @@ func TestGetMTLSCipherSuites(t *testing.T) {
 		mesh                     meshconfig.MeshConfig
 		expectedMTLSCipherSuites []string
 	}{
-		{
-			name:                     "Default MTLS supported Ciphers",
-			expectedMTLSCipherSuites: SupportedCiphers,
```
dgn, line 74: why not test for empty list?

pilot/pkg/security/authn/utils/utils.go
jewertow: Envoy documentation does not clarify what means "unset", so `nil` is safer than an empty slice.

pilot/pkg/security/authn/utils/utils_test.go
jewertow: Right. I restored this test.

dgn (approved):

PR #1057, opened 2024-08-22
[maistra-2.4] OSSM-8001: Fix setting proxy GID

jewertow:
Our injection webhook had a bug in finding proxy GID, but it was hidden, because we were overwriting GID with UID in the injection template.
In general, we set UID=app_UID + 1 and GID=app_UID and we ignored explicitly set `runAsGroup`. This is incorrect, but users did not face any issue in restricted SCC, because there only `runAsUser` is set by OCP. In such a case, application GID is 0, so when we set proxyUID=appUID+1 and proxyGID=appUID we did not face any issues, because all IDs were unique.
This bug reveal when an application has explicitly set UID and GID, e.g. 0 and 0, because then the calculated app UID is 1 and app GID = 0. This means that GID of the app and proxy overlap, so the no-redirect rules for GID owner cause failures in traffic interception. In the last version of the Istio Operator we [fixed](https://github.com/maistra/istio-operator/pull/1787) the injection template and therefore that this long-living bug was exposed.

> [!IMPORTANT]
> Please review commits one by one, because then you will see how it worked (I added unit tests without changing logic in the first 2 commits) and how it works with the actual fixes.

### DIFF
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -970,8 +970,10 @@ func findProxyUGID(pod corev1.Pod) (proxyUID, proxyGID *int64) {
 		if proxyUID == nil {
 			if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.RunAsUser != nil {
 				proxyUID = pointer.Int64(*pod.Spec.SecurityContext.RunAsUser + 1)
-				// valid GID for fsGroup defaults to first int in UID range in OCP's restricted SCC
-				proxyGID = pod.Spec.SecurityContext.RunAsUser
+				proxyGID = pointer.Int64(*proxyUID)
+				if pod.Spec.SecurityContext.RunAsGroup != nil {
+					proxyGID = pointer.Int64(*pod.Spec.SecurityContext.RunAsGroup + 1)
+				}
 			}
 			for _, c := range pod.Spec.Containers {
 				if c.Name != ProxyContainerName && c.SecurityContext != nil && c.SecurityContext.RunAsUser != nil {
@@ -979,8 +981,12 @@ func findProxyUGID(pod corev1.Pod) (proxyUID, proxyGID *int64) {
 					if proxyUID == nil || uid > *proxyUID {
 						proxyUID = &uid
 					}
-					if proxyGID == nil {
-						proxyGID = c.SecurityContext.RunAsUser
+					gid := uid
+					if c.SecurityContext.RunAsGroup != nil {
+						gid = *c.SecurityContext.RunAsGroup + 1
+					}
+					if proxyGID == nil || gid > *proxyGID {
+						proxyGID = &gid
 					}
 				}
 			}
--- a/pkg/kube/inject/webhook.go
+++ b/pkg/kube/inject/webhook.go
@@ -891,74 +891,7 @@ func (wh *Webhook) inject(ar *kube.AdmissionReview, path string) *kube.Admission
 		}
 	}
 
-	var err error
-	var proxyUID *int64
-	var proxyGID *int64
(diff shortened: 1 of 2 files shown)


pkg/kube/inject/webhook.go
```diff
@@ -970,17 +970,23 @@ func findProxyUGID(pod corev1.Pod) (proxyUID, proxyGID *int64) {
...
 					if proxyUID == nil || uid > *proxyUID {
 						proxyUID = &uid
 					}
-					if proxyGID == nil {
-						proxyGID = c.SecurityContext.RunAsUser
+					gid := uid
+					if c.SecurityContext.RunAsGroup != nil {
+						gid = *c.SecurityContext.RunAsGroup + 1
+					}
+					if proxyGID == nil || gid > *proxyGID {
```
dgn, line 988: why do we check for `gid > *proxyGID`? Shouldn't the container's securityContext always take precedence, even if the int is lower?

dgn (approved):
lgtm, just one question

pkg/kube/inject/webhook.go
jewertow: I am not the original author of this condition, so I do not know... But I would also expect that container SC takes precedence over pod SC.

luksa: All this code needs to do is find an UID and a GID that isn't used by any of the containers (i.e. it isn't specified in the pod's nor the container's `securityContext`).

But what the code effectively does is find the max UID/GID used by the containers and then adds a `+ 1`  to that value. Although this isn't wrong, it's not the best, as it wouldn't work if one of the containers uses the last value of the uid/gid range available in the pod's namespace.

So, ideally, we should fix this and return the first unused uid/gid we find. But I guess it's also okay if we leave the current behavior. It looks correct, apart from the corner case described above.

jewertow: > So, ideally, we should fix this and return the first unused uid/gid we find. But I guess it's also okay if we leave the current behavior. It looks correct, apart from the corner case described above.

@luksa do you suggest that my fix is incorrect? Now, this function finds first unused UID/GID, while previously it was only finding the first unused UID, but not GID.

luksa: No, I was just commenting on the `gid > *proxyGID` (which came from `uid > *proxyUID`). 

Your fix looks correct, but we should probably fix both uid and gid so that take the first unused UID/GID instead of taking `max(uid across all containers)+1`.

jewertow: Ok, thanks for clarifying.

dgn: we can consider a fix like that for upstream, but let's not overcomplicate the bugfix

jtyler-knime:
@jewertow, very happy to see this! 

**Tl;Dr - This issue with group ID also surfaces when using the `nonroot-v2` SCC and setting `runAsUser` and `runAsGroup` to the same value.**

To give some additional context around this issue, we ran into this for a slightly different reason. 

In our use-case, we're distributing (via Helm) a chart that needs to be compatible with both OpenShift and other Kubernetes distributions for our end-users. 

Due to requirements in certain Kubernetes platforms, security context constraints **require** we explicitly set `runAsUser` and `runAsGroup` both to non-zero values. Our approach was to follow [Bitnami's lead](https://github.com/bitnami/charts/issues/24251) and set both user and group to the same value (i.e. `1001`; **and** subsequently requested users use the `nonroot-v2` SCC until we could update our chart(s) to be `restricted-v2` compatible). 

This worked fine for us with OSSM 2.5.2, but customer's reported issues when upgrading to OSSM 2.6 as this resulted in app containers and proxy containers having the same group id. 

Note that this wasn't immediately apparent to us, as the problem doesn't surface with upgrading OSSM **until** mesh-injected pods are actually created/recreated since that's when the injector web-hook fires. 

Any insight on when we might see this patched and available in 2.5 or 2.6 on Operator Hub?

PR #1066, opened 2024-12-16
OSSM-8608 2.6 backport: fix: clear tmp files from previous run if exist before copy cni binaries again

mkralik3:
Backport fix https://github.com/istio/istio/pull/54312 from upstream
Downstream ticket: https://issues.redhat.com/browse/OSSM-8608

### DIFF
--- a/cni/pkg/install/binaries.go
+++ b/cni/pkg/install/binaries.go
@@ -47,6 +47,17 @@ func copyBinaries(srcDir string, targetDirs []string, binariesPrefix string) (se
 			}
 			targetFilepath := filepath.Join(targetDir, targetFilename)
 
+			// remove previous tmp file if some exist before creating a new one
+			// the only possible returned error is [ErrBadPattern], when pattern is malformed. Can be ignored in this case
+			matches, _ := filepath.Glob(filepath.Join(targetDir, targetFilename) + ".tmp.*")
+			if len(matches) > 0 {
+				installLog.Infof("Target folder %s contains one or more temporary files with a %s name. The temp files will be deleted.", targetDir, targetFilename)
+				for _, file := range matches {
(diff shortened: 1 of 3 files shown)


luksa:
This code still isn't completely safe. If two goroutines or two processes both invoke `copyBinaries()` with the same srcDir, targetDirs, and prefix, one invocation may delete the file created by the other invocation. Although I don't think this can currently happen, I still don't feel confident in approving this solution. Ideally, the code should create a lock file and then perform the cleanup and copy the file, but this may be overkill. 

@mkralik3 can you ask someone else for their opinion?

mkralik3:
@luksa 
For support `goroutines` for `copyBinaries` func , we can use Mutex from golang, isn't it?
e.g.:
```
var (
    lock = &sync.Mutex{}
)

func copyBinaries(srcDir string, targetDirs []string, binariesPrefix string) (sets.Set[string], error) {
    lock.Lock()
    defer lock.Unlock()

    copiedFilenames := sets.Set[string]{}
...	
```
(tested quickly here: https://github.com/mkralik3/istio-maistra/commit/b00174b83b3eb3f1b58542baff6a63eb15e7404f )

However, it still doesn't resolve the case when e.g. two CNI containers will be starting at the same time on the same node (and the storage speed is slower) the second one can delete the temporary file of the first one. So we need to implement a real file lock in the folder as you mentioned. 
However, in that case, when something kills the istio-cni container during coping binaries [as happened at the origin of this issue], the folder will be locked forever and no next istio-cni container will be able to copy there anything).

luksa (approved):
Discussed in the meeting and we're okay with accepting this as is. If the issue does occur, we expect the process to fail when it tries to rename the tmp file. The container will then be restarted automatically and will recreate the file.

PR #1070, opened 2025-02-17
OSSM-8689 Ignore Gateway API objects with maistra.io/ignore label

luksa:


pilot/pkg/config/kube/gateway/deploymentcontroller.go
```diff
@@ -191,31 +191,35 @@ func NewDeploymentController(client kube.Client, clusterID cluster.ID, env *mode
 	// the Gateway to the queue and reconcile the state of the world.
 	parentHandler := controllers.ObjectHandler(controllers.EnqueueForParentHandler(dc.queue, gvk.KubernetesGateway))
 
-	dc.services = kclient.New[*corev1.Service](client)
+	filter := kclient.Filter{
+		LabelSelector: "maistra.io/ignore!=true",
```
nrfox: Maybe just looking for the `maistra.io/ignore` key is sufficient? This is how the `maistra.io/ignore-namespace` label works and with `true` you have to remember to put quotes around the value in yaml.

luksa: Yeah, I agree, but I wasn't sure if only looking for the presence of the label is okay. What if someone sets `maistra.io/ignore=false`, expecting to turn off the ignore.

luksa: I'll change this, so that it matches the ignore-namespace behavior.

nrfox (approved):

luksa:
@unsortedhashsets We need https://github.com/maistra/test-infra/pull/410 to fix the tests.

unsortedhashsets (approved):

